PR #3, opened 2014-07-16
Move all modules into __unported__ folder + set uninstallable

jgrandguillaume:


bwrsandman:
In another PR, we decided it is best for 7.0 unported modules to be removed entirely from the 8.0 branch.

PR #5, opened 2014-07-16
Applying little fixes and PEP8 to make travis green

jgrandguillaume:


crm_rma_advance_location/__init__.py
```diff
@@ -20,5 +20,4 @@
 #
 ##############################################################################
 from . import stock
-from . import wizard
-
+from . import wizard
```
bwrsandman, line 23: This will cause a PEP8 error, make sure your editor feeds a newline when you save.

bwrsandman, line 23: Note that in the diff (not in this view), there's the symbol indicating that there is no '\n' at the end of the file.

PR #7, opened 2014-08-22
[FIX] partner_id on return picking is logically the claim partner_id

rdeheele:
Original Launchpad proposal : https://code.launchpad.net/~camptocamp/openerp-rma/7.0-crm_claim_rma-fix-lp1317045_rde/+merge/218595

It sets the correct partner_id on picking when we generate a new product return from a claim.

### DIFF
--- a/crm_claim_rma/wizard/claim_make_picking.py
+++ b/crm_claim_rma/wizard/claim_make_picking.py
@@ -211,7 +211,7 @@ class claim_make_picking(orm.TransientModel):
              'move_type': 'one',  # direct
              'state': 'draft',
              'date': time.strftime(DEFAULT_SERVER_DATETIME_FORMAT),
-             'partner_id': partner_id,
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Hi, Romain, I think it's already correct:

```
partner_id = claim.delivery_address_id.id
```

The partner is the delivery address that we have put on the claim, so :-1: 

Regards.

guewen:
The question lies in "what is the correct partner_id" and maybe there is no correct response here. @rdeheele could you elaborate why it is more correct?

@pedrobaeza the wizard generates a return, so it is not clear to me if the partner of the incoming shipment should preferably be the delivery address or the partner who created the claim.

pedrobaeza:
Maybe it can be determined by any other field or selected by the user?

nhomar:
I close this PR I thin it is not valid anymore, if somebody is against please reopenit.

PR #8, opened 2014-09-08
crm_claim_rma: warranty information is computed with on_changes

guewen:
The warranty fields now are updated automatically when the fields
are changed

### DIFF
--- a/crm_claim_rma/crm_claim_rma.py
+++ b/crm_claim_rma/crm_claim_rma.py
@@ -230,6 +230,8 @@ class claim_line(orm.Model):
     def _warranty_limit_values(self, cr, uid, ids, invoice,
                                claim_type, product, claim_date,
                                context=None):
+        if not (invoice and claim_type and product and claim_date):
+            return {}
         date_invoice = invoice.date_invoice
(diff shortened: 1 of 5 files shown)


crm_claim_rma/crm_claim_rma.py
```diff
@@ -472,21 +560,88 @@ def onchange_invoice_id(self, cr, uid, ids, invoice_id, warehouse_id,
...
+                    'product_returned_quantity': invoice_line.quantity,
+                    'unit_sale_price': invoice_line.price_unit,
+                    'location_dest_id': location_dest_id,
+                    'state': 'draft',
+                }
+                line.update(warranty_values(invoice_line.invoice_id,
+                                            invoice_line.product_id))
+                claim_lines.append(line)
+        else:  # happens when the date, warehouse or claim type is modified
+            for command in lines:
```
yvaucher, line 606: In tests lines seems to be False

yvaucher:
Red because of flakes8 from other modules

:+1:

bguillot:
:+1: 

I have tested it, its fine.

PR #17, opened 2015-04-13
CRM Claim Number (partial port of CRM Claim RMA to v8)

mistotebe:
Was looking at porting CRM Claim RMA, but realized that I am not familiar enough with it to reliably port all of it to how the new pickings work. Seeing that the only thing we actually need at the moment was the number field that does not have anything to do with RMAs, I can at least make that available in 8 easily.

### DIFF
--- a/__unported__/crm_claim_rma/__openerp__.py
+++ b/__unported__/crm_claim_rma/__openerp__.py
@@ -74,6 +74,7 @@ Contributors:
     'depends': ['sale',
                 'stock',
                 'crm_claim',
+                'crm_claim_number',
                 'product_warranty',
                 ],
     'data': ['wizard/claim_make_picking_view.xml',
--- a/__unported__/crm_claim_rma/crm_claim_rma.py
+++ b/__unported__/crm_claim_rma/crm_claim_rma.py
@@ -421,18 +421,6 @@ class claim_line(orm.Model):
 class crm_claim(orm.Model):
     _inherit = 'crm.claim'
 
-    def init(self, cr):
-        cr.execute("""
-            UPDATE "crm_claim" SET "number"=id::varchar
-            WHERE ("number" is NULL)
-               OR ("number" = '/');
-        """)
-
-    def _get_sequence_number(self, cr, uid, context=None):
-        seq_obj = self.pool.get('ir.sequence')
-        res = seq_obj.get(cr, uid, 'crm.claim.rma', context=context) or '/'
-        return res
-
     def _get_default_warehouse(self, cr, uid, context=None):
         user_obj = self.pool.get('res.users')
         user = user_obj.browse(cr, uid, uid, context=context)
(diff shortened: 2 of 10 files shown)


pedrobaeza:
We already have a similar module with a best approach for putting the number to already present claims when installing the module (via post-init hook): https://github.com/avanzosc/odoo-addons/tree/master/crm_claim_code.

Please consider to use it instead the one you put and include it on the repo.

mistotebe:
@pedrobaeza, your module looks quite nice and is already up to date on the guidelines, which is great, but I'm concerned that, as it stands, if someone were to port crm_claim_rma to v8, renaming "number" to "code" would have to be done and your unique constraint would have to be redefined to only enforce uniqueness within the same company to cater for how crm_claim_rma is set up in 7.0.

As for this merge proposal, I have only verified that the functionality can exist in v8 and made the minimal possible changes to have that happen so that porting crm_claim_rma is not made any harder and any improvements can happen later on as separate development.

As for the post-init hook, just as the commit that introduced it in your repository notes, it is too late by then: the attempt to load the constraint had already been made. This one had the same problem, that's fixed in the last commit.

Another option could be to modify your module to be a drop-in replacement.

pedrobaeza:
About the post_init_hook, there's no problem to make it there, because unique constraints doesn't apply to empty values.

About migration, it's very easy to assure the migration path putting attribute `oldname="..."` in the field definition.

mistotebe:
The problem is that installing this on a database that already has some CRM tickets, the default values get populated (with '/') and then the attempt is made to set up the uniqueness constraint. That immediately fails and in the case of crm_claim_code, the post-init hook would have been called after everything is set up, but (because of the failure?) does not seem to be called at all, in crm_claim_rma, at least the init() code is called and changes '/' to the value of 'id'. That means crm_claim_code does not actually ever set up the constraint, crm_claim_rma will do so on the next upgrade unless someone inserts a duplicate entry in the meantime. As of the last commit, crm_claim_number does not suffer from either problem and sets up the new values at the right time.

I did not say the way of migration was not fine, just that the module is not compatible with crm_claim_rma as-is, which is what your original comment suggested.

pedrobaeza:
No, I think it's compatible: we just need to rename my module to this name, and put oldname="number" on the field definition.

mistotebe:
@pedrobaeza: I know and that's why I said "as-is"... Also, if you want it merged, why not actually raise another pull request and we can get a reviewer's comment on what to get in. I'm OK with either, even renaming/merging this work with crm_claim_code to make it easier for your customers to upgrade. Just want the functionality in, really, not interested in any politics. As you can see, the changes required were so small that they might not even qualify for copyright protection so I did not update any copyright related headers either.

To recap, we need some functionality already present in this repo, I reused what was here, not looking anywhere else. You already have a module that achieves the same goal, which is superior in keeping up with 8.0 formal guidelines and want it merged instead but some changes are needed as of today...

I did not expect something as trivial as this module to spark a discussion like this that is eating time for the both of us.

mistotebe:
As pointed out in #24, crm_claim_code had been merged into OCA/crm since (where this one should have gone in the first place, sorry).

I will take a look at what is provided and propose any changes if necessary to provide a drop-in replacement as per the comments above (I suspect the uniqueness constraint in crm_claim_code is still too fragile and incompatible with 7.0 databases for being too strict).

pedrobaeza:
I didn't remember this one certainly when reviewing the inclusion of the other! But you can see that it has been also improved. Anyways, let me know what are the problems you see and we can collaborate to solve them.

I close this PR then.

PR #23, opened 2015-04-23
product warranty port v8

mistotebe:
I am not a member of the OCA, so I don't believe that I can even propose the license change from GPLv3 to AGPLv3. Other than that, it is a trivial port for a trivial module.

Issue #20, opened 2015-04-21
Porting crm_claim_rma to v8


mistotebe:
It seems we will need this module for 8 anyway, so I'm setting up a ticket to track the progress here.

I intend to proceed in several stages with separate pull request for each if you don't mind:
- [x] porting product_warranty (#23)
- [x] splitting crm_claim_number off crm_claim_rma (proposal in #17)
- [x] porting crm_claim_rma itself (#24)

There is one thing I need to know before I start: product_warranty is still licensed under GPLv3, I am not sure if, as a non-member, I am allowed to propose a pull request relicensing the module to AGPLv3?

Also, new API+odoo code guidelines are a must now for acceptance or is a simpler port not using the new API acceptable?

mistotebe:
Porting in progress:
- have not touched the copyrights on product_warranty yet, other than that, it is mostly compatible: #23
- rebased crm_claim_number port on top of current 8.0: #17
- started porting crm_claim_rma, but hit a couple of dead ends

I am not sure how to go about picking creation, whether to set up a route+picking type per warehouse and try to use procurements to drive the entire picking process or just create a picking manually or something in between as I'm not sure what the intent was behind the new stock system and have not seen a module that actually taps into it in this way

sebastienbeau:
Hi thanks for the work, regarding the licence as Akretion is one of the original author I am totally agree for moving under AGPL-v3 ! (I have not idea why it's under GPL, pretty sure it's due to a bad copy paste...)
@jgrandguillaume are you ok for moving under AGPL-v3?

mistotebe:
Porting ~finished, see above for pull requests.

eltjor:
Hi thanks for putting the effort in to port this module to v8.0. Is there a timeline for when the it will be complete?

mistotebe:
This module is feature complete and has been successfully deployed in production environments since (hence some improvements made in the past months).

See limitations in the README and consider your procurement/stock routing set-up which has become too powerful in 8.0 for this module to properly interact with for more complex cases (without unduly expanding its scope and size). Also, bear in mind that while I do not expect there to be much in terms of migration from 7.0, it has not been attempted yet.

Also, I have no idea what the community's stance is on this port vs. the port by Vauxoo that has been proposed (in #26) after I have published mine. Usually, the reasons that a proposal does not get merged boil down to the shortage of reviewers, please consider helping with that if you can.

pedrobaeza:
[...]

PR #24, opened 2015-05-12
Porting crm_claim_rma to Odoo 8

mistotebe:
This is a proposal for crm_claim_rma as outlined in #20, it depends on #17 and #23.

The picking creation might be slightly too naive, but it works for the stock set up I tested and I would need some help improving it if there are scenarios where the pickings fail to be created appropriately.


crm_claim_rma/__openerp__.py
```diff
@@ -66,6 +66,7 @@
  * Joel Grand-Guillaume <joel.grandguillaume@camptocamp.com>
  * Guewen Baconnier <guewen.baconnier@camptocamp.com>
  * Yannick Vaucher <yannick.vaucher@camptocamp.com>
+ * Ondřej Kuzník <ondrej.kuznik@credativ.co.uk>
```
yvaucher: Please move description in `README.rst` here is the template: https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst

crm_claim_rma/crm_claim_rma_view.xml
```diff
@@ -217,32 +215,31 @@
...
                                 </sheet>
                             </form>
                         </field>
                     </group>
-                    <group col="4" colspan="4" attrs="{'invisible':[('state', '&lt;&gt;','open')]}">
+                    <group col="4" colspan="4">
                         <separator string="Action" colspan="4" />
                         <button name="%(action_claim_picking_in)d"
-                                string="New Products Return" states="open"
+                                string="New Products Return"
```
yvaucher: Why do you change visibility of those buttons?

crm_claim_rma/tests/test_lp_1282584.py
```diff
@@ -78,7 +73,7 @@ def test_00(self):
 
         res = self.WizardMakePicking.action_create_picking(
             cr, uid, [wizard_id], context=wiz_context)
-        self.assertEquals(res.get('res_model'), 'stock.picking.in',
+        self.assertEquals(res.get('res_model'), 'stock.picking',
                           "Wrong model defined")
```
yvaucher: This check seems meaningless to me. Here we tested the wizard was opening the right view based on model.
We cannot rely on model anymore.

crm_claim_rma/tests/test_lp_1282584.py
```diff
@@ -22,11 +22,6 @@
 
 
 class test_lp_1282584(common.TransactionCase):
-    """ Test wizard open the right type of view
-
-    The wizard can generate picking.in and picking.out
-    Let's ensure it open the right view for each picking type
-    """
```
yvaucher, line 29: Why removing the description and not fixing the test ?

product_warranty/__openerp__.py
```diff
@@ -0,0 +1,58 @@
...
+    'data': [
+        'security/ir.model.access.csv',
+        'res_company_view.xml',
+        'product_warranty_view.xml',
+    ],
+    'demo_xml': [],
+    'test': [],
+    'installable': True,
+    'active': False,
+    'certificate': '',
```
yvaucher: certifitate key and active key can be removed

product_warranty/__openerp__.py
```diff
@@ -0,0 +1,58 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Product warranty',
+    'version': '1.0',
+    'category': 'Generic Modules/Product',
+    'description': """
+Product Warranty
```
yvaucher: Please also move this in a `README.rst` file

crm_claim_rma/crm_claim_rma_view.xml
mistotebe: State field does not exist on crm.claim anymore.

yvaucher: Ok thanks I missed that

yvaucher:
Thanks it looks good to me, one last thing is that test to be modified:

You can remove test test_lp_1282584, AFAIK there is now only a single view for incoming and outgoing pickings. So no need to check which view will be openned anymore.

You might want to convert it to testing the wizard opens. So we keep the same coverage. However, reference to launchpad bug 1282584 https://bugs.launchpad.net/openerp-rma/+bug/1282584  can be removed as picking model changed in 8.0.

mistotebe:
Unit tests are ported now and the CLA for @craiggowing should be sent over soon (we sent the signed CLAs for all credativ people over a year ago, but it must have been lost since).

yvaucher:
Thanks for the migration of tests :+1:

pedrobaeza:
I say the same as in the other PRs: instead of creating and managing a code for crm.claim, you should use the module _crm_claim_code_ that already handle that (https://github.com/OCA/crm/tree/8.0/crm_claim_code). Put it as dependency and remove it from your code, and we will have a lighter module.

mistotebe:
@pedrobaeza You could have commented on #17 which this one depends on (as stated in #20) that crm_claim_code had been merged already. I will take a look at what is provided and propose any changes if necessary (I suspect the uniqueness constraint in crm_claim_code is still too fragile and incompatible with 7.0 databases for being too strict).

aristobulo-eezee:
@pedrobaeza @gurneyalex @mistotebe I already included crm_claim_code in our PR, what I propose to you (if you agree of course) is to rebase our branch to this one and include all @mistotebe changes and ours, what do you think?

I think that it could a good starting point, since both PR are based in original crm_claim_rma features

pedrobaeza:
Well, it seems that Vauxoo's branch contains improvements, so it's better to coordinate with them the merge.

nhomar:
@aristobulo-eezee 

I will check your migration to validate everything is running as expected, that was one of the modules we touched less.

I will be back soon with the feedback to either join your PR to ours and make all compatible or give you specific feedback.

Regards.

nhomar:
I will close this PR in favor of https://github.com/OCA/rma/pull/32 which is totally migrated to new Api, if somebody opose scream aloud,

mistotebe:
@nhomar
Have you migrated the additional features as well? (Reporting fields, etc.)

nhomar:
You can have all modules migrated here:

https://github.com/vauxoo/rma

You can follow the squashing to make the PR module per module here:

https://github.com/OCA/rma/pull/26

I think in no more than 2 weeks we will have everything merged.

Regards.

mistotebe:
What I mean is that there are usability and reporting enhancements in this module that were not present in 7.0 version, an incomplete list:
- uom is tracked on the claim line
- when creating a claim line manually, the value is taken from cost price
- value of a line+claim is tracked

nhomar:
2015-08-05 5:14 GMT-05:00 mistotebe notifications@github.com:

> What I mean is that there are usability and reporting enhancements in this
> module that were not present in 7.0 version, an incomplete list:
> - uom is tracked on the claim line
> - when creating a claim line manually, the value is taken from cost
>   price
> - value of a line+claim is tracked
> 
> I will give a check on this and come back to you tomorrow with the
> comments.....

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

PR #27, opened 2015-07-02
Add indexes on claim_id fields

guewen:
Because they are often searched for


pedrobaeza:
Travis problems are unrelated (but we should treat them), so :+1:

sbidoul:
:+1:

bguillot:
:+1:

PR #33, opened 2015-08-03
8.0 squash product warranty module between oca vauxoo

yaniaular:
Product_warranty module is migrated to API 8.0 and odoo 8.0.


README.md
```diff
@@ -1,40 +0,0 @@
-[![Build Status](https://travis-ci.org/OCA/rma.svg?branch=8.0)](https://travis-ci.org/OCA/rma)
```
pedrobaeza, line 1: Why do you remove this file?

README.rst
```diff
@@ -0,0 +1,48 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
```
pedrobaeza: This file should be inside product_warranty

product_warranty/__openerp__.py
```diff
@@ -31,8 +33,14 @@
...
+        'demo/product_warranty.xml',
+        'demo/res_company.xml',
+    ],
     'test': [],
     'installable': True,
+    'active': False,
     'images': ['images/product_warranty.png'],
 }
+
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
pedrobaeza: Remove these editor specific lines

product_warranty/demo/product_warranty.xml
```diff
@@ -0,0 +1,64 @@
...
+
+        <record id="return_instruction_1" model="return.instruction">
+            <field name="name">Default Instruction</field>
+            <field name="instructions">To return a product purchased through our platform vendor Marketplace access our online Returns Center and click Return products to begin the application process for the return.
+Select the product you want to return and the reason for the return. You will need you to provide detailed information to enable the seller to handle your request. Once the seller has reviewed your application, it will respond by e-mail within 3 working days.</field>
+            <field name="is_default">1</field>
+        </record>
+
+        <record id="return_instruction_2" model="return.instruction">
+            <field name="name">Provicional Instruction</field>
```
pedrobaeza: s/Provicional/Provisional

product_warranty/i18n/es.po
```diff
@@ -159,3 +159,136 @@ msgstr "Instrucciones de devolución"
 #: field:return.instruction,instructions:0
 msgid "Instructions"
 msgstr "Instrucciones"
+
+#. module: product_warranty
+#: field:product.supplierinfo,active_supplier:0
+msgid "Active supplier"
+msgstr ""
```
pedrobaeza: Translate this

product_warranty/i18n/es_MX.po
```diff
@@ -0,0 +1,16 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* product_warranty
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Odoo Server 8.0\n"
```
pedrobaeza, line 7: Why is this file empty?

product_warranty/i18n/es_PA.po
```diff
@@ -0,0 +1,16 @@
+# Translation of Odoo Server.
```
pedrobaeza, line 1: And this one?

product_warranty/i18n/es_VE.po
```diff
@@ -0,0 +1,16 @@
+# Translation of Odoo Server.
```
pedrobaeza, line 1: And this?

product_warranty/models/product_warranty.py
```diff
@@ -0,0 +1,131 @@
...
+    _name = "return.instruction"
+    _description = "Instructions for product return"
+
+    name = fields.Char('Title', required=True)
+
+    instructions = fields.Text(
+        'Instructions',
+        help="Instructions for product return")
+
+    is_default = fields.Boolean('Is default',
```
pedrobaeza: Align this way to avoid a lot of lines:

```
is_default = fields.Boolean(
    'Is default', 
    help="If is default, will be use to set the default value in supplier "
         "infos. Be careful to have only one default")
```

product_warranty/models/product_warranty.py
```diff
@@ -0,0 +1,131 @@
...
+from openerp import api, fields, models
+
+
+class ReturnInstruction(models.Model):
+
+    _name = "return.instruction"
+    _description = "Instructions for product return"
+
+    name = fields.Char('Title', required=True)
+
```
pedrobaeza: Letting this empty line makes more confusing the code read.

product_warranty/models/product_warranty.py
```diff
@@ -0,0 +1,131 @@
...
+        help="Instructions for product return")
+
+    is_default = fields.Boolean('Is default',
+                                help="If is default, will be use "
+                                "to set the default value in "
+                                "supplier infos. Be careful to "
+                                "have only one default")
+
+
+class ProductSupplierInfo(models.Model):
```
pedrobaeza: Split each model in one file with the name of the model. For example, for this model, put it on product_supplierinfo.py file.

product_warranty/models/product_warranty.py
```diff
@@ -0,0 +1,131 @@
...
+    def _get_warranty_return_address(self):
+        """ Method to return the partner delivery address or if none,
+        the default address
+
+        dedicated_delivery_address stand for the case a new type of
+        address more particularly dedicated to return delivery would be
+        implemented.
+
+        """
+        for record in self:
```
pedrobaeza: If you put @api.one, self is already only one record and you make the operations directly with self.

product_warranty/i18n/es_MX.po
nhomar, line 7: To avoid the warning "es_**" do not have translations, on this case there is not any different terms, but without the file the warning appears always. We use it es_MX and es_PA

product_warranty/i18n/es_PA.po
nhomar, line 1: To avoid the warning "es_**" do not have translations, on this case there is not any different terms, but without the file the warning appears always. We use it es_MX and es_PA

pedrobaeza:
Don't rename README.md on the repo to .rst, you can see the explanation here: https://github.com/OCA/maintainer-tools/pull/94#issuecomment-127217187

product_warranty/i18n/es_VE.po
nhomar, line 1: To avoid the warning "es_**" do not have translations, on this case there is not any different terms, but without the file the warning appears always. We use it es_MX and es_PA and es_VE

pedrobaeza:
Thanks for the hard work, Yanina!

product_warranty/i18n/es_MX.po
pedrobaeza, line 7: Thanks for the explanation. Yeah, it makes sense

product_warranty/models/product_warranty.py
nhomar: Is there some written decided guideline for this specific topic?

Becaus _class_stuff separated from field_names on my opinion is better (which is different from your one). I do not remember we decided such guideline on any moment.

And in case we do it should be a the pylinter checker, don't you think?

pedrobaeza: No, it's just a style preference because of the reason I have told you. I put logical blocks without line breaks to group them together. In this case, all fields declaration.

yaniaular:
@pedrobaeza @nhomar It is ready

product_warranty/models/product_supplierinfo.py
```diff
@@ -0,0 +1,108 @@
...
+    def _get_warranty_return_address(self):
+        """ Method to return the partner delivery address or if none,
+        the default address
+
+        dedicated_delivery_address stand for the case a new type of
+        address more particularly dedicated to return delivery would be
+        implemented.
+
+        """
+        for record in self:
```
pedrobaeza: You can remove also here the loop and only use self as the method is @api.one decorated.

product_warranty/models/return_intruction.py
```diff
@@ -0,0 +1,40 @@
...
+
+    name = fields.Char('Title', required=True)
+    instructions = fields.Text(
+        'Instructions',
+        help="Instructions for product return")
+    is_default = fields.Boolean('Is default',
+                                help="If is default, will be use "
+                                "to set the default value in "
+                                "supplier infos. Be careful to "
+                                "have only one default")
```
LeartS: Is there a reason you didn't enforce the "only one default" rule with a constraint?

product_warranty/models/product_supplierinfo.py
```diff
@@ -0,0 +1,108 @@
...
+    def get_warranty_return_partner(self):
+        result = [('company', 'Company'),
+                  ('supplier', 'Supplier'),
+                  ('other', 'Other'),
+                  ]
+        return result
+
+    @api.model
+    def _get_default_instructions(self):
+        """ Get selected lines to add to exchange """
```
LeartS, line 41: This docstring doesn't make sense (what selected lines, what exchange?), I see it was already there before the porting, I think this could be a good occasion to change or remove it.

product_warranty/models/product_supplierinfo.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @api.one
+    @api.depends('warranty_return_partner')
+    def _get_warranty_return_address(self):
+        """ Method to return the partner delivery address or if none,
+        the default address
+
+        dedicated_delivery_address stand for the case a new type of
+        address more particularly dedicated to return delivery would be
+        implemented.
```
LeartS: The docstring references `dedicated_delivery_address` but that doesn't exist anywhere in the file or even the module!

product_warranty/models/product_supplierinfo.py
```diff
@@ -0,0 +1,108 @@
...
+    @api.model
+    def _get_default_instructions(self):
+        """ Get selected lines to add to exchange """
+        instruction_ids = self.env['return.instruction']\
+            .search([('is_default', '=', True)], limit=1)
+        return instruction_ids
+
+    @api.one
+    @api.depends('warranty_return_partner')
+    def _get_warranty_return_address(self):
```
LeartS: According to the [guidelines](https://www.odoo.com/documentation/8.0/reference/guidelines.html#symbols), the method name should be`_compute_warranty_return_address`, not `_get` (with the new API the function doesn't actually return anything!)

product_warranty/models/product_supplierinfo.py
```diff
@@ -0,0 +1,108 @@
...
+                    partner_id = record.name.id
+                elif return_partner == 'company':
+                    if record.company_id.crm_return_address_id:
+                        partner_id = record.company_id.\
+                            crm_return_address_id.id
+                elif return_partner == 'other':
+                    if record.warranty_return_other_address:
+                        partner_id = record.\
+                            warranty_return_other_address.id
+                record.warranty_return_address = partner_id
```
LeartS: I think this should be unindented, at the moment it's inside the `if` block which I don't think is correct: we want to assign the return address to the company address if there is no return partner.
If that's not the case, then the line `partner_id = record.company_id.partner_id.id` is useless and can be removed.

product_warranty/models/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# This program is distributed in the hope that it will be useful,       #
+# but WITHOUT ANY WARRANTY; without even the implied warranty of        #
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the         #
+# GNU General Public License for more details.                          #
+#                                                                       #
+# You should have received a copy of the GNU General Public License     #
+# along with this program.  If not, see <http://www.gnu.org/licenses/>. #
+#########################################################################
+
+from . import return_intruction
```
LeartS: You forgot an `s`. (in**s**truction). The file should also be renamed.

product_warranty/models/product_supplierinfo.py
yaniaular, line 41: The get_warranty_return_partner method is used in <a href="https://github.com/OCA/rma/blob/8.0/__unported__/crm_claim_rma/crm_claim_rma.py#L92">crm_claim_rma</a> module for a selection field called <a href="https://github.com/OCA/rma/blob/8.0/__unported__/crm_claim_rma/crm_claim_rma.py#L147">warranty_type</a> in claim.line model. I guess it was a way to inherit and change the options in a selection field and that change take effect in the fields of both modules (product_warranty and crm_claim_rma).

product_warranty/models/product_supplierinfo.py
yaniaular: You're right.

product_warranty/models/product_supplierinfo.py
yaniaular: @LeartS I agree with you with regard to the method name. But, 
I do not understand your comment in brackets. With the new API, the compute method not return something, Modifies the value of the field referring to the way <a href="https://github.com/odoo/odoo/blob/a9e3d74713b6882c64617345696ad326d75a4745/addons/event/event.py#L186">self.field = value</a>

LeartS: Yes exactly, my comment was only a justification of why `_compute` is better than `_get`: `_get` usually implies the method returns something.

yaniaular: Perfect, thanks.

product_warranty/models/product_supplierinfo.py
yaniaular: I think it's a TODO that someone left in the docstring. What do you suggest?

nhomar: 2015-08-04 9:50 GMT-05:00 Yanina (Vauxoo) notifications@github.com:

> I think it's a TODO that someone left in the docstring. What do you
> suggest?

Remove it, if it does not mean anything.

## 

---

Saludos Cordiales

CEO at Vauxoo https://www.vauxoo.com Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
http://about.me/nhomar?promo=email_sig

nhomar:
Hello all.

It is with all comments done applied.

Can somebody else move the PR forward due to we need this module in order to continue the migration of all the rest squashing them.

Regards.

cc @pedrobaeza @gurneyalex

product_warranty/models/product_supplierinfo.py
pedrobaeza: This hasn't been removed

pedrobaeza:
A little remark, but seems :+1: 

Thanks for the work!

nhomar:
@pedrobaeza Just pushed that change!.

Another +1¿?

LeartS:
Code review :+1:

nhomar:
@LeartS thanks, 2 +1 then merge.

Issue #46, opened 2015-09-04
Creation of picking with the New Products Return button and  New Delivery button not work with Odoo 8.0


yaniaular:
- Creation of picking with the New Products Return button not work with Odoo 8.0.
- Creation of picking with New Delivery button not work with Odoo 8.0
- Dest. Location field in wizard of New products return button is not automatically filled

Here a video with issues:

http://youtu.be/MFzc1sl4vZ4

yaniaular:
This issue is resolving here https://github.com/OCA/rma/pull/47

yaniaular:
@nhomar Can you check the  PR https://github.com/OCA/rma/pull/47 related with this issue https://github.com/OCA/rma/issues/46 to proceed to merge?

yaniaular:
this issue is resolved in the pr related. Please to review

PR #47, opened 2015-09-05
[IMP] crm_claim_rma module adapted to new features with respect to picking in Odoo 8.0

yaniaular:
## Enhancements and fixes:
- crm_claim_rma module adapted to new characteristics with respect to picking in stock module. 
  Now, create pickings in rma is working with odoo 8.0. http://youtu.be/wG-UFXpCoCA
- crm_claim_rma module updated to crm_claim_type module of the OCA/crm
- <a href="runbot.odoo-community.org/runbot/build/3114474">Error on runbot</a> is corrected =>  <b>_auto_init Table `crm_claim`: column `claim_type` has changed type (DB=int4, def=selection), data moved to column `claim_type_moved0`</b> The definition of claim_type field is repeated.
- New tests is added with demo invoice and and subsequently make claim lines from the invoice lines
- wf_service.trg_validate changed by signal_workflow (new way to activate signals in odoo 8.0)
- Add claim lines is corrected, <a href="https://github.com/OCA/rma/pull/47/files#diff-2e2d61b52eae4e2807850aa7f95aaa61L503">the cycle is changed by list of tuples</a> because on the tests only is added the last invoice line and the correct way is like <a href="https://github.com/odoo/odoo/blob/8.0/openerp/osv/fields.py#L1591">said here</a>

```
-                claim_lines.append([0, 0, line])
+                claim_lines.append((0, 0, line))

-        for line in claim_lines:
-            value = self._convert_to_cache(
-                {'claim_line_ids': line}, validate=False)
-            self.update(value)
+        value = self._convert_to_cache(
+            {'claim_line_ids': claim_lines}, validate=False)
+        self.update(value)
```

### DIFF
--- a/crm_claim_rma/tests/test_picking_creation.py
+++ b/crm_claim_rma/tests/test_picking_creation.py
@@ -2,6 +2,8 @@
 ##############################################################################
 #
 #    Author: Yannick Vaucher
+#            Yanina Aular
+#    Copyright 2015 Vauxoo
 #    Copyright 2014 Camptocamp SA
 #
 #    This program is free software: you can redistribute it and/or modify
@@ -21,16 +23,15 @@
 from openerp.tests import common
 
 
-class test_picking_creation(common.TransactionCase):
+class TestPickingCreation(common.TransactionCase):
+
     """ Test the correct pickings are created by the wizard. """
 
     def setUp(self):
-        super(test_picking_creation, self).setUp()
+        super(TestPickingCreation, self).setUp()
 
-        self.WizardMakePicking = self.env['claim_make_picking.wizard']
-        self.StockPicking = self.env['stock.picking']
-        ClaimLine = self.env['claim.line']
-        Claim = self.env['crm.claim']
+        self.wizard_make_picking = self.env['claim_make_picking.wizard']
+        claim = self.env['crm.claim']
 
         self.product_id = self.env.ref('product.product_product_4')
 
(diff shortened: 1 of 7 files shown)


yaniaular:
@nhomar PLease review this PR... thanks

osvalr:
@nhomar Can you check this PR? It solves issues pointed out in [this issue](https://github.com/OCA/rma/issues/46)

nhomar:
:+1:  @pedrobaeza can you check the body of the PR to understand the change and give a quickily feedback thanks.

@yaniaular Can you extend a little the description in order to contextualize other oca members?

yaniaular:
@nhomar @pedrobaeza When the user clicks on the "New Product Return" button or "New Delivery" button, it is created a picking of incoming type (for product returns) or outgoing type (for deliveries), this (the picking type) was managed like char ('in' or 'out') in version 7.0 of odoo, BUT, this does't work in version 8.0 of odoo, because picking type field is many2one instead of char.

Now with this PR https://github.com/OCA/rma/pull/47, when a picking is created in crm_claim_rma module, a picking type of many2one type is assigned, and it works as expected in version 8.0 of odoo.

crm_claim_rma/tests/test_picking_creation.py
```diff
@@ -39,41 +41,44 @@ def setUp(self):
...
-                'claim_id': self.claim_id.id,
-                'location_dest_id': self.warehouse_id.lot_stock_id.id,
-            })
 
     def test_00_new_product_return(self):
         """Test wizard creates a correct picking for product return
 
         """
-        wizard = self.WizardMakePicking.with_context({
+        wizard = self.wizardmakepicking.with_context({
```
pedrobaeza: Recommended way of internal variables names is snake_case: `wizard_make_picking`, but it's not the most important here

pedrobaeza:
I think some isinstance calls are not needed, but overall :+1:

crm_claim_rma/tests/test_picking_creation.py
yaniaular: @pedrobaeza I got it, thanks

yaniaular:
Thanks @pedrobaeza .

@osvalr Can you give :+1:?

osvalr:
There you go @yaniaular 
:+1:

PR #52, opened 2015-09-28
fix mistakes made when porting to the new API

gurneyalex:


crm_claim_rma/tests/test_picking_creation.py
```diff
@@ -142,3 +142,7 @@ def test_02_new_product_return(self):
         self.assertEquals(picking.location_dest_id,
                           self.warehouse_id.lot_stock_id,
                           "Incorrect destination location")
+
+    def test_copy(self):
+        new_claim = self.claim_id.copy()
+        self.assertNotEqual(new_claim.code, self.claim_id.code)
```
pedrobaeza, line 148: Isn't this test for crm_claim_code instead of this module?

gurneyalex, line 148: yep, this is just to check that the code I removed was not required (and actually it was harmful since we were consuming twice from the crm code sequence).

pedrobaeza:
:+1:

damdam-s:
:+1: thanks

lepistone:
:+1: thanks

PR #55, opened 2015-10-15
port crm_claim_rma to 8.0

osvalr:
This PR ports crm_claim_rma to 8.0 and OCA Guidelines.

**`UPDATE`**
The following video shows with its basic functioning.
[Spanish Audio](http://www.youtube.com/watch?v=Eu5cj0HlCh4)

PR #61, opened 2015-11-03
[FIX]crm_claim_rma: Removed stock.pciking create function.

jesusVMayor:
The inheritance of stock.picking create has no sense in v8.0, always returns an incorrect name.

Example:
INT/001 (without warehouse) instead of WH01/OUT/001

### DIFF
--- a/crm_claim_rma/__openerp__.py
+++ b/crm_claim_rma/__openerp__.py
@@ -24,7 +24,7 @@
 
 {
     'name': 'RMA Claim (Product Return Management)',
-    'version': '8.0.1.1.0',
+    'version': '8.0.1.1.1',
     'category': 'Generic Modules/CRM & SRM',
     'description': """
 Management of Return Merchandise Authorization (RMA)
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think the goal of this is to have a separate nomenclature, but @yaniaular and @osvalr can say it better

jesusVMayor:
But not only affect the RMA pickings. It also set the name when creates a picking from purchase order, when manually creates internal pickings...

osvalr:
Hi @jesusVMayor 
I've discussed with @yaniaular and from out part there isn't counter opinion about the changes you are suggesting. It's okay. :+1:

pedrobaeza:
:+1: then for me too

osvalr:
@jesusVMayor we are working right now to get improvements in **`crm_claim_rma`** module in [PR#55](https://github.com/OCA/rma/pull/55), Do you agree to apply changes made in this as part of it? Please feel free to give us your opinion over there

jesusVMayor:
@osvalr feel free to use this changes

osvalr:
Hi @jesusVMayor changes here have been applied into OCA/rma#55 in bbd08af, so I think this PR should be closed if there isn't aditional comments
/cc @pedrobaeza

pedrobaeza:
Please change the author of the commit for crediting @jesusVMayor

PR #62, opened 2015-11-03
[FIX]crm_claim_rma: fix the view crm_claim_rma_form_view

jesusVMayor:
it get fixed the view crm_claim_rma_form_view to allow creating new categories with the correct model.


pedrobaeza:
:+1:

LeartS:
:+1: the related code that uses the context key is [here](https://github.com/odoo/odoo/blob/06246bac9a984297d41480738ad797d3ab7d3dbb/addons/crm/crm.py#L164) if anyone was wondering where it's used.

PR #38, opened 2015-08-05
[IMP] module to manage locations in RMA

yaniaular:
This modules work to manage the RMA locations

PR #67, opened 2015-11-26
[8.0] Improved module crm_rma_stock_location 

osvalr:
# crm_rma_stock_location

This module gets improved having the following features:
- Adapted to Odoo 8.0
- Adds Sales menu section for RMA Stocks quantities that summarizes how much does the RMA locations has On-Hand and how much is Forecasted for all products.
- Includes internal locations for Refurbish and Loss.
- Includes buttons to quickly send products to Loss stock, Refurbish stock or Warehouse Stock location from warehouse transfers
- Includes a button to create product loss from Claims.
- Added unitttest to validate its functionality


nhomar:
:+1: 

cc @pedrobaeza can you give me your :+1: here it comes from #38

yaniaular:
:+1:

crm_rma_stock_location/__init__.py
```diff
@@ -19,5 +21,14 @@
...
 ##############################################################################
 
-from . import stock_warehouse
-from . import product
+from . import models
+from . import wizards
+from openerp import SUPERUSER_ID
+
+
+def post_init_hook(cr, registry):
```
moylop260: What is that?

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+==================
+RMA Stock Location
+==================
+
+Allow the user to know how much for a product is available 'On Hand' and how much
+is virtually (expected to be) available for RMA locations. Adding for the
+different product views (Tree, Form and Kanban) information about it.
+
+Both quantities are computed and includes its children locations.
```
pedrobaeza: s/includes/include (is plural)

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+It is useful use it as a quick snapshot for RMA from product perspective.
```
pedrobaeza: s/use/to use

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+It also adds the following location on warehouses :
+
+ * Loss
```
pedrobaeza, line 19: Why is this needed if we have already Scrapped location?

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+ * Refurbish
```
pedrobaeza: Refurbished

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Various wizards on incoming deliveries that allow you to move your
```
pedrobaeza: s/Various/Several

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+Configuration
+=============
+
+No configuration is needed
+
+Usage
+=====
+
+* Go to Sales > After-sale Services and note that 'RMA Quantity On Hand' and
+  'RMA Forecasted Quantity' had been included and it'll be shown when at least
```
pedrobaeza: s/had/has

pedrobaeza: it'll be shown/they'll be shown

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+  a product has either a on hand or forecasted quantities available.
```
pedrobaeza: s/a product/when a product

pedrobaeza: s/a on hand/on hand

crm_rma_stock_location/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+* In the other hand, it provides three wizards to make stock moves (transfers)
+  allowing to do product returns (incoming), send a product to loss or, to a refurbish
```
pedrobaeza: s/refurbish/refurbished

crm_rma_stock_location/__init__.py
pedrobaeza: This is a method that is run after module installation

crm_rma_stock_location/models/product_product.py
```diff
@@ -0,0 +1,131 @@
...
+                localdict = {'virtual': element.rma_virtual_available,
+                             'qty': element.rma_qty_available,
+                             'value': value}
+                if eval('qty %s value or virtual %s value' %
+                        (operator, operator), localdict):
+                    ids.append(element.id)
+            res.append(('id', 'in', ids))
+        return res
+
+    def _rma_product_available(self):
```
pedrobaeza, line 68: Mark this explicitly as @api.multi

crm_rma_stock_location/models/product_template.py
```diff
@@ -0,0 +1,49 @@
...
+                                     string=_('RMA Quantity On Hand'))
+
+    rma_virtual_available = fields.Float(compute='_rma_product_available',
+                                         digits_compute=dp.
+                                         get_precision('Product Unit'
+                                                       ' of Measure'),
+                                         string=_('RMA Forecasted Quantity'))
+
+    def _rma_product_available(self):
+        for product in self:
```
pedrobaeza, line 46: Call it `template`, not `product`

crm_rma_stock_location/models/product_template.py
```diff
@@ -0,0 +1,49 @@
...
+            product.rma_qty_available = sum(
+                [p.rma_qty_available for p in product.product_variant_ids])
```
pedrobaeza: Use `sum(template.mapped('product_variant_ids.rma_qty_available'))`

crm_rma_stock_location/views/stock_picking.xml
```diff
@@ -0,0 +1,30 @@
...
+<openerp>
+    <data>
+        <!-- INHERITED VIEW FOR THE OBJECT : stock_picking -->
+        <record id="picking_in_form" model="ir.ui.view">
+            <field name="name">crm_claim_rma.picking_in_form</field>
+            <field name="model">stock.picking</field>
+            <field name="inherit_id" ref="stock.view_picking_form"/>
+            <field name="arch" type="xml">
+                <xpath expr="//field[@name='state']" position="before">
+                        <button name="%(action_stock_picking_from_claim_picking)d"
```
pedrobaeza, line 11: You can use the same action in the three buttons, but putting a different context on each one, and you'll get the same result.

crm_rma_stock_location/__init__.py
osvalr: @moylop260 yes, as @pedrobaeza says this is a key for a method that executes just after the module is installed.. look over [here](https://github.com/odoo/odoo/blob/8.0/openerp/modules/loading.py#L185) to see when is called.

crm_rma_stock_location/models/product_template.py
osvalr: Does [mapped(..)](https://www.odoo.com/documentation/8.0/reference/orm.html#openerp.models.Model.mapped) method returns a list with duplicates removed, Doesn't it?  This is why this line was made in this way

osvalr: IMO using mapped here, it can introduce a bug

crm_rma_stock_location/__init__.py
moylop260: Why not use a separated file?
IMHO `__init__.py` files should be just for imports sentences
FYI we have disabled some lint checks in  `__init__.py` files

yaniaular: The module adds new location fields to stock.warehouse model, like lot_rma_id in this module, or lot_stock_id in odoo stock module, the locations should be created ever for each warehouse, when this location fields is added in this module, there may be previously warehouses created, then, the location loss_loc_id, lot_refurbish_id and lot_rma_id shoulde be created for these warehouse. https://github.com/odoo/odoo/blob/8.0/addons/stock/stock.py#L3507-L3523

moylop260: FYI I'm talking about of a refactory similar to this one:
https://github.com/OCA/product-attribute/blob/9.0/product_pricelist_fixed_price/__init__.py#L19

yaniaular: I see, what you propose is placing it in a separate file.
@osvalr Osval helps us to separate. Just that the file should be named post_init_hook.py, because
init_hook.py sounds like it runs at the beginning of the installation.

yaniaular:
``` python
from .post_init_hook import post_init_hook
```

yaniaular: in **openerp**.py 

``` python
'post_init_hook': 'post_init_hook',
```

nhomar:
@pedrobaeza answering:

## 

## Why is this needed if we have already Scrapped location?

Actual Scrapped location is an inventory location, which make an account move when you have real valuation and exclude from inventory the returned product, that's ok ones you finish the process, the objective of this new location is follow the process:

RMA -> Loss (yet on my inventory but saved in an special place before decide.) -> Scrap (I decided not any more and adjusted once per per period).

Due to this you can (before send it to scrap) pass the product to MRP in order to refurbish it, if possible, but it can be a manual verification which need the product temporarly on your inventories not on Scrap location.

With Loss location (the new one) it is part of the new Workflow proposed and necesary to ensure double verification to achieve this deal.

I hope it clarifies.

Thanks for the feedback.

crm_rma_stock_location/models/product_product.py
nhomar, line 68: +1

crm_rma_stock_location/models/product_template.py
nhomar, line 46: +1

crm_rma_stock_location/models/product_template.py
nhomar: @osvalr can you try it? I think you are confusing .set with .mapped, but can you make a little PoC pliz in order to be sure?

crm_rma_stock_location/views/stock_picking.xml
nhomar, line 11: @pedrobaeza but both options give the same result, which is the difference? I personally prefer an action per view (as it is). Which is the added value of change that dude?

crm_rma_stock_location/models/product_template.py
pedrobaeza: mapped returns in this case a list like `[1, 3, 4.5, 23, 3, 1]` (with duplicates if any), so it's correct.

crm_rma_stock_location/views/stock_picking.xml
pedrobaeza, line 11: The difference is to avoid the definition of three action window. More compact and robust, because there is only one point of failure.

crm_rma_stock_location/__init__.py
osvalr: @moylop260 I think this is probably something extra to add for directory module structure, I'll propose that later, what do you think? I'm going to proceed to do the refactor you pointed out.

moylop260: +1 with "add for directory module structure" Thank you!

pedrobaeza:
OK, that makes sense. Maybe the name is not the best one (but I can't imagine any other), but the approach is correct.

crm_rma_stock_location/models/product_template.py
osvalr: @pedrobaeza okay, you were right

nhomar:
@pedrobaeza I will merge with this name, and may be for v9.0 we rebautism it I do not know any better name.

thanks for the feedback.

PR #70, opened 2016-01-18
[FIX] crm_claim_rma: Fixed company_id default fields

hugosantosred:
Change company_id field default method to fix a problem  when upgrading this module from a previous version.

This is happening because in new api for Many2one fields is expecting an object and _company_default_get function is returning an id. So I've changed the default method of company_id field and converted it to a function so it can return a company object instead of an integer id.

### DIFF
--- a/crm_claim_rma/models/claim_line.py
+++ b/crm_claim_rma/models/claim_line.py
@@ -44,6 +44,13 @@ class ClaimLine(models.Model):
     _description = "List of product to return"
     _rec_name = "display_name"
 
+    @api.model
+    def _get_company_default(self):
+        company_env = self.env['res.company']
+        default_company_id = company_env._company_default_get(
+            'claim.line')
+        return company_env.browse(default_company_id)
+
     SUBJECT_LIST = [('none', 'Not specified'),
                     ('legal', 'Legal retractation'),
                     ('cancellation', 'Order cancellation'),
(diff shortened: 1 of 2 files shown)


nhomar:
@hugosantosred 

Sorry I do not understand why you change the lambda for the method, it is doing the same, did I miss something?

hugosantosred:
@nhomar The lambda is returning the company_id as an integer value. With the lambda, when upgrading this addon from a previous version, the server is throwing this error:

```
  File "/home/vagrant/develop/odoo/odoo/openerp/models.py", line 2400, in _set_default_value_on_column
    default = default(self, cr, SUPERUSER_ID, context)
  File "/home/vagrant/develop/odoo/odoo/openerp/api.py", line 241, in wrapper
    return old_api(self, *args, **kwargs)
  File "/home/vagrant/develop/odoo/odoo/openerp/api.py", line 336, in old_api
    result = method(recs, *args, **kwargs)
  File "/home/vagrant/develop/odoo/odoo/openerp/fields.py", line 417, in <lambda>
    lambda recs: self.convert_to_write(value(recs))
  File "/home/vagrant/develop/odoo/odoo/openerp/fields.py", line 1578, in convert_to_write
    return value.id
AttributeError: 'int' object has no attribute 'id'
```

So I made the function to pick that integer value that _company_default_get is returning and with it, return the company that matches that ID.

gurneyalex:
:+1:

yvaucher:
Indeed in 8.0 _company_default_get has not be ported:

https://github.com/odoo/odoo/blob/8.0/openerp/addons/base/res/res_company.py#L225

:+1:

PR #74, opened 2016-01-21
[FIX] crm_claim_rma Exception in claim_make_picking.wizard

hugosantosred:
Fixed raise Warning. Changed to raise exceptions.Warning


pedrobaeza:
:+1: Fast-tracking as obvious

PR #76, opened 2016-03-09
[FIX] Picking should be in state 'assigned' or 'partially_available' before transfer

yaniaular:
According to change https://github.com/vauxoo-dev/odoo/commit/f89220a51313e1bf46ec82175f2449c2e1a0455c the picking state should be in 'assigned' or 'partially_available' before to transfer.


yaniaular:
@nhomar FYI

pedrobaeza:
:+1:

yaniaular:
@nhomar can you give +1?

PR #77, opened 2016-03-10
Add product type filter when adding claim lines

mdietrichc2c:
When creating a new reception/delivery from the claim, claim lines with non-stockable products are not used: https://github.com/OCA/rma/blob/7.0/crm_claim_rma/wizard/claim_make_picking.py#L284 
However, when creating those claim lines in `onchange_invoice_id`, all invoice lines are used, including lines with non-stockable products.

Since a check is done afterwards in https://github.com/OCA/rma/blob/7.0/crm_claim_rma/wizard/claim_make_picking.py#L53 to ensure one and only one IN/OUT is created per claim line, this means that you can still create an IN/OUT for the non-stockable line, which will result in an empty picking.

### DIFF
--- a/crm_claim_rma/crm_claim_rma.py
+++ b/crm_claim_rma/crm_claim_rma.py
@@ -551,7 +551,8 @@ class crm_claim(orm.Model):
         warehouse_obj = self.pool['stock.warehouse']
         invoice_line_ids = invoice_line_obj.search(
             cr, uid,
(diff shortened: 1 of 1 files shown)


yaniaular:
@mdietrichc2c I am agree with your change. :+1:

gurneyalex:
:+1:

PR #80, opened 2016-04-14
Add missing model decorator

kkarolis:
Breaks when module crm_claim_rma is installed and a "Modify/Cancel" refund is created.

PR fixes it.


pedrobaeza:
:+1:

PR #81, opened 2016-04-21
product_warranty: migration V9

cyrilgdn:


product_warranty/__openerp__.py
```diff
@@ -22,23 +22,23 @@
...
     'category': 'Generic Modules/Product',
     'author': "Akretion,Odoo Community Association (OCA),Vauxoo",
     'website': 'http://akretion.com',
     'license': 'AGPL-3',
     'depends': ['product'],
     'data': [
         'security/ir.model.access.csv',
-        'views/res_company_view.xml',
-        'views/product_warranty_view.xml',
+        'views/res_company.xml',
```
pedrobaeza: Don't rename these files. They are good with the `_view` suffix. You can have also a `res_company_data.xml` file for example for insert data in the companies, and thus the name will be ambiguous without the suffix.

cyrilgdn: OK sorry, rename reverted.

guewen:
:+1:

mdietrichc2c:
:+1:

product_warranty/models/product_supplierinfo.py
```diff
@@ -81,22 +55,27 @@ def _compute_warranty_return_address(self):
                 "brand manufacturer. Doesn't necessarly mean that the "
                 "warranty to be applied is the one of the return partner "
                 "(ie: can be returned to the company and be under the "
-                "brand warranty")
+                "brand warranty"
```
yvaucher: Missing a closing parenthese inside the string.

yvaucher:
Only a tiny remark otherwise LGTM

PR #83, opened 2016-05-03
crm_claim_rma: use procurement for delivery

cyrilgdn:
Create a procurement.order instead of manually creating stock.picking for delivery in claim.

So delivery in claim can use the same flow that classic delivery (e.g. if warehouse uses advanced routes).


yvaucher:
@cyrilgdn Wasn't the target `9.0` branch ?

cyrilgdn:
@yvaucher No, this new feature will be ported in V9 but is also needed in V8 because of modifications on stock between V7 and V8.

gurneyalex:
superseeded

PR #84, opened 2016-05-03
crm_rma_lot_mass_return: Fix build

cyrilgdn:
RMA 8.0 build is broken since https://github.com/odoo/odoo/commit/f89220a51313e1bf46ec82175f2449c2e1a0455c

Demo data try to call do_detailed_transfer on a non-assigned picking.

I removed this call and fixed tests. (force assigign the picking and call do_detailed_transfer in setup)


cyrilgdn:
Seems to be fixed by @yaniaular in https://github.com/OCA/rma/commit/7850037cfef6d95690e956df8ddfd790212999a5

PR #85, opened 2016-05-03
crm_claim_rma_code: migration V9

cyrilgdn:
Depends on:
- [x] https://github.com/OCA/crm/pull/90 
- [x] https://github.com/OCA/crm/pull/92

PR #82, opened 2016-04-27
crm_rma_location: migration V9

cyrilgdn:

PR #86, opened 2016-05-09
crm_claim_rma: migration V9

cyrilgdn:
(Depends on already merged  #81 #82 https://github.com/OCA/rma/pull/85)

Based on V8 #83


crm_claim_rma/models/crm_claim.py
```diff
@@ -0,0 +1,206 @@
...
+
+        return results
+
+    @api.multi
+    def message_get_suggested_recipients(self):
+        recipients = super(CrmClaim, self).message_get_suggested_recipients()
+        try:
+            for claim in self:
+                if claim.partner_id:
+                    self._message_add_suggested_recipient(
```
mdietrichc2c: Should be `claim._message_add_suggested_recipient(recipients, partner=claim.partner_id, reason=_('...')` for this method and the one below.

crm_claim_rma/models/claim_line.py
```diff
@@ -0,0 +1,413 @@
...
+
+            if not line_id.invoice_line_id:
+                raise exceptions.UserError(_('Please set invoice first'))
+
+            line_id.set_warranty_limit()
+            line_id.set_warranty_return_address()
+
+    @api.model
+    def _get_sequence_number(self):
+        """
```
yaniaular: First line empty in method docstring, according to https://www.python.org/dev/peps/pep-0257/#specification

crm_claim_rma/models/claim_line.py
```diff
@@ -0,0 +1,413 @@
...
+        @return the value of the sequence for the number field in the
+        claim.line model.
+        """
+        return self.env['ir.sequence'].next_by_code('claim.line')
+
+    @api.model
+    def create(self, vals):
+        """
```
yaniaular: First line empty in method docstring, according to https://www.python.org/dev/peps/pep-0257/#specification

cyrilgdn:
_message_add_suggested_recipient and Vauxoo's docstring fixed.

I fixed some pylint-odoo warning too.

guewen:
:+1:

crm_claim_rma/models/crm_claim.py
```diff
@@ -0,0 +1,210 @@
...
+                                          "products.")
+    sequence = fields.Integer(default=lambda *args: 1)
+    warehouse_id = fields.Many2one('stock.warehouse', string='Warehouse',
+                                   required=True,
+                                   default=_get_default_warehouse)
+    rma_number = fields.Char(size=128, help='RMA Number provided by supplier')
+
+    @api.model
+    def _get_claim_type_default(self):
+        claim_type = self.env['crm.claim.type'].search([])
```
mdietrichc2c: Should retrieve the ID for `crm_claim_type.crm_claim_type_customer` instead of returning the model if None is found.

crm_claim_rma/models/crm_claim.py
```diff
@@ -0,0 +1,210 @@
...
+        return claim_type_code and sequence.next_by_code(
+            claim_type_code
+        ) or '/'
+
+    @api.model
+    def create(self, values):
+        values = values or {}
+        if 'code' not in values or not values.get('code') \
+                or values.get('code') == '/':
+            values['code'] = self._get_sequence_number(values['claim_type'])
```
mdietrichc2c: You can have a `KeyError` if 'claim_type' is not in values (it is not filled with the default value) ; it should be tested, and retrieved with the default method if not available.

mdietrichc2c:
:-1: for the moment ; the `claim_type` field has no default value during creation, and is mandatory.

crm_claim_rma/models/crm_claim.py
cyrilgdn: @mdietrichc2c If None is found, it means that crm_claim_type.crm_claim_type_customer does not exists too... So it should never append as crm_claim_type module installs types.

Maybe it should return crm_claim_type_customer directly ?

mdietrichc2c: @cyrilgdn sorry, I wan't clear : the ID should be returned directly of course, no need to still have the search.

mdietrichc2c:
LGTM, :+1: (code review and small tests)

crm_claim_rma/views/crm_claim.xml
```diff
@@ -0,0 +1,291 @@
...
+            </xpath>
+            <xpath expr="//field[@name='team_id']/parent::group" position="replace">
+                <group colspan="4" col="4" groups="base.group_user">
+                    <field name="warehouse_id"/>
+                    <field name="date"/>
+                    <field name="user_id"
+                           context="{'default_groups_ref': ['base.group_user', 'base.group_partner_manager', 'base.group_sale_salesman_all_leads']}"/>
+                    <field name="priority" widget="priority"/>
+                    <field name="date_deadline"/>
+                    <field name="team_id" groups="base.group_multi_salesteams"/>
```
guewen: `base.group_multi_salesteams` does no longer exist, to remove

crm_claim_rma/models/claim_line.py
```diff
@@ -0,0 +1,408 @@
...
+                       select=True,
+                       default=fields.date.today())
+    name = fields.Char('Description', default='none', required=True,
+                       help="More precise description of the problem")
+    priority = fields.Selection([('0_not_define', 'Not Define'),
+                                 ('1_normal', 'Normal'),
+                                 ('2_high', 'High'),
+                                 ('3_very_high', 'Very High')],
+                                'Priority', default='0_not_define',
+                                compute='_set_priority',
```
moylop260: Should we take advantage of this PR to use `_compute_priority` [standard](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#field)?

mdietrichc2c:
@moylop260 that's right ; I renamed the method accordingly.

moylop260:
👍

moylop260:
@yaniaular 
Could you give us your feedback, please?

crm_claim_rma/views/crm_claim.xml
```diff
@@ -0,0 +1,291 @@
...
+            </xpath>
+            <xpath expr="//field[@name='partner_phone']" position="attributes">
+                <attribute name="required">1</attribute>
+            </xpath>
+            <xpath expr="//field[@name='email_from']" position="attributes">
+                <attribute name="required">1</attribute>
+            </xpath>
+
+            <!-- Smart buttons -->
+            <xpath expr="//field[@name='name']/parent::group" position="after">
```
yvaucher: Buttons should be on top thus replace after by before

As reference, see : https://github.com/OCA/OCB/blob/9.0/openerp/addons/base/res/res_partner_view.xml#L146-L152

crm_claim_rma/views/crm_claim.xml
```diff
@@ -0,0 +1,291 @@
...
+                <div class="oe_right oe_button_box" name="buttons">
```
yvaucher: You can remove oe_right and s/buttons/button_box in name (it will use the responsive design putting buttons in a More drop down menu)

crm_claim_rma/views/crm_claim.xml
```diff
@@ -0,0 +1,291 @@
...
+                    <button class="oe_inline oe_stat_button"
```
yvaucher: oe_inline can be removed on each buttons

See
https://github.com/OCA/OCB/blob/9.0/openerp/addons/base/res/res_partner_view.xml#L148

yvaucher:
@cyrilgdn I made the fixes here: https://github.com/cyrilgdn/rma/pull/1

yvaucher:
:+1:

PR #92, opened 2016-08-26
Modify onchange in order to work for invoice_id

mdietrichc2c:
Found an issue in the module `crm_claim_rma` : since the context from the view is not transferred to the onchange (it's for related fields only), the claim lines were not created when changing the invoice.

It's now fixed, and I added a comment on how this works.


guewen:
LGTM

pedrobaeza:
:+1:

yvaucher:
:+1:

PR #93, opened 2016-08-31
Change picking type to "Out"

mdietrichc2c:
Small issue (or maybe an explanation is needed?), but when creating an "OUT" picking with the method `_create_picking`, the picking type is set as an internal transfer, instead of a delivery.


guewen:
:+1:

moylop260:
👍

osvalr:
:+1:

pedrobaeza:
:+1:

PR #96, opened 2016-09-10
[Fix] Make warranty_return_partner field translatable

levkar:
self explanetory

### DIFF
--- a/product_warranty/models/product_supplierinfo.py
+++ b/product_warranty/models/product_supplierinfo.py
@@ -21,7 +21,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.  #
 ##########################################################################
 
-from openerp import api, fields, models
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I prefer to put the selection in the field itself, as new API allows an easy inheritability through `selection_add`. That way, the translation will be automatic.

levkar:
I agree but this function is called here:

https://github.com/OCA/rma/blob/8.0/crm_claim_rma/models/claim_line.py#L145

I suppose it happened while migrating from v9. If I were migrating, I would put them on both places since it is very unlikely to have another selection there.

pedrobaeza:
You can change it this way:

```
def get_warranty_return_partner(self):
    return self.env['product.supplierinfo']._fields['warranty_return_partner'].selection
```

and the translations will also work.

gurneyalex (approved):

PR #102, opened 2016-10-19
[FIX] crm_rma_prodlot_supplier: better inheritance

cubells:
The rest of method does some stuff related to transfer details items.

If model is called from another model, e.g, a button or a function, etc. then, an nonexistent record error is raised, because transfer_item_id at line 51 is not related with the model from which is called.


RoelAdriaans (approved):
Had the same issue, and same fix

pedrobaeza (approved):

cubells:
Thanks a lot.

PR #103, opened 2016-11-30
[8.0 FIX] Fixes returned quantity

levkar:
Adding this field makes quantity work as it is expected

### DIFF
--- a/crm_claim_rma/models/claim_line.py
+++ b/crm_claim_rma/models/claim_line.py
@@ -89,7 +89,7 @@ class ClaimLine(models.Model):
                                  ('2_high', 'High'),
                                  ('3_very_high', 'Very High')],
                                 'Priority', default='0_not_define',
-                                compute='_set_priority',
+                                compute='_compute_priority',
                                 store=True,
                                 readonly=False,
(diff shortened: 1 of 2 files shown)


rubencabrera:
Edited my comment. Yes, this works fine!
The quantity is correctly added to the claim line (but there's still something wrong with the new returns created from the claim).

levkar:
If you can explain your findings (if you think it is related) we can take a look.

rubencabrera:
Please excuse my stupidity, it gets worse on Mondays. It works like a charm!

pedrobaeza:
Can you check Travis?

levkar:
I am having hard time resolving checks. I thought it is because of compute but I was wrong. Is it this line?

self.assertEquals(eval(res['context'])['type'], 'out_refund')

If it is, then I'm not sure how to fix it. This is not related to my commit anyway.

I get discouraged when i fix something and get stuck for other reasons when i try to contribute back. Maybe we should practice full-oca-compliant CI cycle in our company to improve our contributions.

Don't get me wrong, this is not a complaint, just how I feel and think.

gurneyalex:
superseesed by #121

PR #105, opened 2016-12-27
[9.0][FIX] crm_claim_rma migration

jjscarafia:
Fix migration from v8. xml id for sales team was renamed as you can see here: [V8](https://github.com/OCA/rma/blob/8.0/crm_claim_rma/data/crm_case_section.xml) vs [V9](https://github.com/OCA/rma/blob/9.0/crm_claim_rma/data/crm_team.xml)


crm_claim_rma/migrations/9.0.1.0/pre-migration.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Shorten headers, like in https://github.com/OCA/maintainer-tools/blob/master/template/module/__init__.py

pedrobaeza (approved):
Little comment

pedrobaeza:
One thing, is version number correct? Shouldn't be `9.0.1.0.0`?

jjscarafia:
Thanks @pedrobaeza . Version works ok with 9.0.1.0 but I've renamed it to 9.0.1.0.0 because is the standard, thanks.
I've also use the shorten headers

moylop260 (approved):

PR #106, opened 2017-01-10
Update in the older modules for rma

techspawn:
Resolved the compatibility issues for crm claim and rma module for odoo10 now we are able to install apps in odoo10


pedrobaeza:
Please, only one module per PR, so it's easier to review. You should start with `crm_claim` and continue with each dependency.

techspawn:
ya alright i will create different PR for each module and give pull request for the same

PR #110, opened 2017-03-16
[ADD][9.0]RMA (WIP)

AaronHForgeFlow:
RMA
===

You use this module to manage RMA (Return Merchandise Authorization) in Odoo

The RMA can be send from the company to the customer or from the supplier to
the company

The RMA user creates the RMA and assign a responsible by selecting the
affected invoice. The available operations are refund, repair and deliver or
replace.
The assigned person checks if the conditions for the RMA are correct and
approve the RMA

The assigned person have to go to the RMA lines from the RMA and operate. THe
available options are creating receipts, delivery orders and refunds.

Once the assigned person considers that the RMA is settled he/she can
set the RMA to Done.

Configuration
=============

To configure this module, you need to:

1. Go to Settings > Users and add the user to the group RMA manager
2. Go to Inventory > Settings > Return Merchandising Authorization and select
 the option "Display 3 fields on rma: partner, invoice address, delivery
 address" if needed
3. Go to Inventory > Settings > Configuration > Warehouse management >
Warehouses and add a default RMA location and RMA picking type for customers
and suppliers RMA picking type. It's very important to select the type of
operation supplier if we are moving in the company and customer if we are
moving out of the company.

Usage
=====

RMA are accessible though Inventory menu. There's four menus, divided by type
. Users can access to the list of RMA or RMA lines.

Create an RMA:
1. Select a partner. Fill the rma lines by selecting an invoice.
2. Request approval and approve.
3. Click on RMA Lines button.
4. Click on more and select an option: "Receive products", "Create Delivery
Order, Create Refund"
5. Go back to the RMA. Set the RMA to done if not further action is required

Issues
=====
You can also comment about this module in this issue: https://github.com/OCA/rma/issues/111

OBS: Additional modules will allow to create RMA from SO or PO


rubencabrera:
@drl9319
@carlosap92
@jesuspablo

nhomar:
Dude I am really sorry.

But this is a copy of our work in Vauxoo/rma (may be with little changes).

I have several questions:

1. Why remove our authorship?
2. Why propose a PR to OCA/RMA first instead explain the changes into vauxoo/rma?.

We had invested on this work a HUGE amount of work and we are opensourcing everything, but IMHO this is not the way to propose a PR or a Refactor... 

AFAICR we are all in the same team did not we?

cc @jbeficent

JordiBForgeFlow:
@nhomar Obviously we have to respect the authorship. We'll add that ASAP!

We have been struggling with the current proposed approach in v8, and wanted to brainstorm with new ideas. We researched what has been done by your team, also we looked into Oracle, SAP, and Odoo v9 and v10, to try to identify new ideas to improve the process.

We did not see https://github.com/Vauxoo/rma/pull/167. And we are going to review it now.
 
We are happy to redirect the PR to Vauxoo/RMA and to work together on this!

@yaniaular I understand that you are under active development. We would be happy to discuss together.

AaronHForgeFlow:
@nhomar Of course we will respect the authorship, that was not intended, apologizes for the misunderstanding.
I agree to move the PR to Vauxoo/RMA and keep up with the work together.
We will review your work at https://github.com/Vauxoo/rma/pull/167

pedrobaeza:
You should move it to OCA, not to Vauxoo fork I think.

jgrandguillaume:
Hi,


First of all thank you for refactoring this module, it is an very old one that we took back from a version 5.0 and never had the budget to clean it up ! I support this move, and like very much the new way of handling the supplier and customer RMA.

I've tested it and it works well overall. I have a few remarks:

 * In the previews versions, I was able to create a refund from a line, that I didn't succeeded. It was very useful and once agreed with management, the person in charge was able to create a refund for a customer automatically, avoiding mistakes (e.g. wrong price or qty). That was very much appreciated.

 * In my opinion, one of the most important feature in the previous module was to be able to let the system create the appropriate picking according to the infos set on the inventory tab of the product (supplier.info). There you were able to tel the system wether a product should go back to you, the supplier or another adresse (entity, e.g. for Smartphone, you don't return to the seller, not the manufacturer, but the entity in charge of the repair).
Previous behavior: 

1. If set to company, the picking created was returning in the RMA location with address=company address
2. If set to supplier, the picking created was returning in the supplier location with address=supplier address
3. If set to other, the picking created was returning in the supplier location with address=specified address address

That was helpful, in our case, we used this to record the picking in the system, so our customer can then print the shipping label properly, send it back to the customer in order for him to ship back to the proper place (without costs in our case). Now I didn't figured out how to do this anymore.

 * In the previous version (again ;) ), the system computed also for each line if the warranty was still going on or not (well, it was based on the current value of the supplier infos, not historical, but well..better than nothing). That is IMO important to keep. We want to know *before* deciding what to do, if the warranty is still valid or not.


Apart from those points, I'm happy to support this refactoring !

Thanks again for the work,

:+1: 

Joël

AaronHForgeFlow:
Hello @jgrandguillaume,

Thank you for your review. I'm glad you like it. This is the purpose of the RFC. Regarding your remarks:

- It's possible to create to create refunds from the rma lines. Previously I assigned the wrong permission to the action but now it should be visible from the rma lines:

image

- Currently, the delivery address is taken from the delivery address in the rma but I agree with you it's more valid to take the address from the supplier.info model. Also, when the destination location is the company the system should take the company address not the company itself. Totally right.

- Regarding the utility of the RMA type "Other", I think it would be fine if we add the delivery address in the wizard, so the user can change the delivery address of the picking created. In my opinion this would solve that use case.

- Checking the warranty of the line is pending to be implemented but I agree that this is a must.

We'll keep working on these improvements :+1: 

Thanks  for your support,

Aaron

jgrandguillaume:
Hi @aheficent 


Thanks for the answers ! So after checking this again:

 * Refund are alright, I confirm it works well
 * Ok for your suggestion about the type "other"
 * Perfect about warranty check !

One thing more I'm sure I was not sure to explain properly. When returning goods, there are two things to work on:

 * Address
 * Location

For the address,  you get it right. For the location, I think we should also change the destination location according to the selected option in supplier.info:

 * Company -> Set it to RMA location
 * Supplier -> Set it to supplier
 * Other -> Set it to supplier as well

My goal with this was (in the first version), to always represent what's physically done by the customer in the returning picking of Odoo. So if the customer ships the goods back to supplier, the returning picking should go there. Then, when making the delivery, it should take it FROM the supplier location to customer. Doing so looks good to me because you can followup in Odoo what is happening in the physical world.

What do you think ?

In any case, thank a lot for taking care of this module.

Best regards,

Joël

AaronHForgeFlow:
Hello @jgrandguillaume 

Thanks for your comments,

We continue to analyze. Regarding the dropship case, we are currently thinking about what it's the best approach but it will be covered.

Best regards,

Aaron

rubencabrera:
I'm trying the backport to 8.0 and it's been easier to do so than trying to fix the issues on the current 8.0 branch. Still have to tinker around a lot, but seems to work quite fine. If anyone is interested, it is in[ my fork of OCA/rma branch](https://github.com/rubencabrera/rma/tree/refactor) named refactor, I would really appreciate some help testing.

Thanks a lot for this great work!

AaronHForgeFlow:
That's interesting @rubencabrera I'll review. 

BTW, I'm modifying the wizards so they create procurements instead of pickings. The reason is to make the process more "standard", more similar to the PO and SO process. From the user perspective it's almost the same.

Besides, I'm adding another RMA type, "dropship", this type will allow the user to return products directly from the customer to the supplier. 

I'll let you know when it's ready so everyone will be able to give their feedback.

angelmoya:
I'm trying to install without demo data, with no other module preinstalled and get this error:

ParseError: "null value in column "code" violates not-null constraint
DETAIL:  Failing row contains (2, null, 2017-04-06 08:31:54.285736, 21, 25, 1, null, null, null, 1, null, null, 22, ship_only, null, 20, 23, one_step, f, null, 24, 2017-04-06 08:31:54.285736, null, null, 1, null, null, null, 9, 6, 7, 8, 19).
" while parsing /mnt/extra-addons/rma/rma/demo/stock.xml:53, near
<record id="stock.stock_warehouse_shop0" model="stock.warehouse">
        <field name="lot_rma_id" ref="location_rma_shop0"/>
        <field name="rma_cust_out_type_id" ref="picking_type_rma_cust_out"/>
        <field name="rma_cust_in_type_id" ref="picking_type_rma_cust_in"/>
        <field name="rma_sup_out_type_id" ref="picking_type_rma_sup_out"/>
        <field name="rma_sup_in_type_id" ref="picking_type_rma_sup_in"/>
    </record>

rma/__openerp__.py
```diff
@@ -0,0 +1,33 @@
...
+             'security/ir.model.access.csv',
+             'views/rma_view.xml',
+             'views/stock_view.xml',
+             'views/stock_warehouse.xml',
+             'views/invoice_view.xml',
+             'views/product_view.xml',
+             'wizards/rma_make_picking.xml',
+             'wizards/rma_refund.xml',
+             'wizards/stock_config_settings.xml',
+             'demo/stock.xml',
```
angelmoya: demo data must be on demo element:
'demo': ['demo/stock.xml',
             'demo/rma.xml',]
and split demo from required data

rma/README.rst
```diff
@@ -0,0 +1,101 @@
...
+
+#. Go to Settings > Users and add the user to the group RMA manager.
+#. Go to Inventory > Settings > Return Merchandising Authorization and select
+   the option "Display 3 fields on rma: partner, invoice address, delivery
+   address" if needed.
+#. Go to Inventory > Settings > Configuration > Warehouse management >
+   Warehouses and add a default RMA location and RMA picking type for customers
+   and suppliers RMA picking type. It's very important to select the type of
+   operation supplier if we are moving in the company and customer if we are
+   moving out of the company.
```
angelmoya: This option is only configurable if you config your warehouse to use routes, if not, then default configuration is ok (is ok after fix demo and data xml)

angelmoya:
Invoice selector shows draft invoices, and also invoices that had ben returned.

What do you think about filter at least draft invoices?

I'm thinking that returned invoice could be a parcial return.

And also, it would be nice to control how, who and when return invoice is done.

angelmoya:
And one more think about refund, I think that refund would generate shipping in.

angelmoya:
Ok, I understand now that operations are not generated automatically, and it could be OK...

But I think that RMA Lines view could be improved, with an statusbar, and action buttons on header

rma/__openerp__.py
AaronHForgeFlow: You're right. Thank you @angelmoya :1st_place_medal:

rma/README.rst
AaronHForgeFlow: I'll attend to this also

AaronHForgeFlow:
Yes, draft invoices or returned invoices should not be considered in the current status. Currently I think that draft invoices are not considered. Returned invoices will imply that there's already a refund created for the RMA. We will consider to add this feature.

Regarding the RMA lines view maybe it's better to add buttons for the actions but regarding the status, there's no status yet in the RMA line. The status is a related field to the RMA order. Maybe would be good to put a status in the rma order line that depends on the quantities pending to receive/deliver/refund. What do you think?

rma/__openerp__.py
rubencabrera: Tried to change it to demo list and had the same error, also installing without demo data. It might be because a broken previous installation, but just in case...
Anyway, the code is required (at least in 8.0, where I am testing), so might have to add it on the demo records.

AaronHForgeFlow: I changed the demo files to the demo list and then I installed in a blank database without demo data. It worked. I haven't pushed the changes yet.

rubencabrera:
> Regarding the RMA lines view maybe it's better to add buttons for the actions but regarding the status, there's no status yet in the RMA line. The status is a related field to the RMA order. Maybe would be good to put a status in the rma order line that depends on the quantities pending to receive/deliver/refund. What do you think?

For multiple lines RMAs this could be very useful, because each product might have a different process depending on the state of warranty or damage. The RMA order is sometimes more useful to the post-sales team, while RMA lines can be helpful for technical team and their communication of the status of the claim.

rubencabrera:
> Regarding the RMA lines view maybe it's better to add buttons for the actions but regarding the status, there's no status yet in the RMA line. The status is a related field to the RMA order. Maybe would be good to put a status in the rma order line that depends on the quantities pending to receive/deliver/refund. What do you think?

For multiple lines RMAs this could be very useful, because each product might have a different process depending on the state of warranty or damage. The RMA order is sometimes more useful to the post-sales team, while RMA lines can be helpful for technical team and their communication of the status of the claim.

AaronHForgeFlow:
In the last commit, we modified the  wizard in order to create procurements instead of pickings directly. The manager have to create first the proper routes. In the rma line the user selects an operation, this operation contains the applicable routes depending on the rma type (customer, supplier or dropshipping). For example: We create a customer RMA with one RMA line with the operation of replace. In the operation we have to specify a route for Customer RMA:
image
In that route we can specify different pull rules. This allows to receive/send the items from/to different locations.

The next step should be introduce the warranty to the rma_lines :smiling_imp:

rubencabrera:
@aheficent For the warranty check of rma_lines, please take into account an issue of the previous rma implementation:
Invoices are the right reference for warranty calculations, but it is quite common for businesses changing to odoo that their invoices are not registered, so no warranties can be computed. A manual introduction of the field would be sufficient, IMO, but it wasn't possible in the former modules.

AaronHForgeFlow:
I'll take that into account @rubencabrera Thank you!! :+1:

rubencabrera:
I have a few points to review:
1) When creating an invoice from add_invoice_id on a Customer RMA, the invoice form loaded has the supplier label on partner, tried to use a context with {'default_type': 'out_invoice'} to fix it but can't make it work this way.

2) On the RMA lines, operation_id is readonly and required, son no manual lines can be added, just the ones generated adding an invoice. What if we don't have the invoice in odoo because we are migrating from another ERP?

3) When loading demo data, my warehouse and locations get overriden with the RMA ones in demo data. Is this intended? EDIT: this only happend on the backport I'm doing to 8.0, sorry for the inconvenience, I guess it's not that important anyway, the README explains quite well how to set RMA locations.

AaronHForgeFlow:
Hello @rubencabrera,
Thanks for your feedback. Regarding your comments:

1. You're correct, I'll figure it out how to fix this
2. I'll remove those restrictions in order to allow to create manual lines.
3. The demo data is intended to no be updated (noupdate=1), I'll review this.

rma/models/rma.py
```diff
@@ -413,17 +421,35 @@ def _compute_move_count(self):
...
+        if self.rma_id.warehouse_id.lot_rma_id:
+            return self.rma_id.warehouse_id.lot_rma_id.id
+        else:
+            return False
+
+    @api.model
+    def _default_src_location_id(self):
+        if self.type == 'customer':
+            if self.rma_id.partner_id.property_stock_customer:
+                return lines.rma_id.partner_id.property_stock_customer.id
```
rubencabrera, line 539: Isn't lines an undefined variable here and in line 440?

rubencabrera (commented):
Undefined variables in _default_dest/src_location_id methods

rma/models/rma.py
AaronHForgeFlow: Yes, this method is not used indeed. I'll remove it. Thank you! @rubencabrera

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+        compute=_compute_qty_to_refund, store=True)
+    qty_refunded = fields.Float(
+        string='Qty Refunded', copy=False,
+        digits=dp.get_precision('Product Unit of Measure'),
+        readonly=True, compute=_compute_qty_refunded, store=True)
+
+    @api.onchange('product_id')
+    def _onchange_product_id(self):
+        if not self.invoice_id:
+            return
```
andhit-r, line 683: Hi @aheficent , is this by design? If RMA unrelated to invoice then onchange will not fill **name** field. Meanwhile field **name** is readonly thus data can not be saved.

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+# © 2015 Eezee-It, MONK Software, Vauxoo
+# © 2013 Camptocamp
+# © 2009-2013 Akretion,
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+from openerp import _, api, fields, models
+from openerp.addons import decimal_precision as dp
+from openerp.exceptions import UserError
+from dateutil.relativedelta import relativedelta
+from openerp.tools import DEFAULT_SERVER_DATE_FORMAT
+import math
```
andhit-r, line 12: unused import

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+from datetime import datetime
+import calendar
```
andhit-r, line 14: unused import

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+                                 required=True, readonly=True,
+                                 states={'draft': [('readonly', False)]})
+    assigned_to = fields.Many2one('res.users', 'Assigned to',
+                                  track_visibility='onchange')
+    requested_by = fields.Many2one('res.users', 'Requested by',
+                                   track_visibility='onchange',
+                                   default=lambda self: self.env.user)
+    company_id = fields.Many2one('res.company', string='Company',
+                                 default=lambda self: self.env.user.company_id)
+    rma_line_ids = fields.One2many('rma.order.line', 'rma_id',
```
andhit-r, line 79: You should make it readonly unless it's draft

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+                                   required=True, readonly=True,
+                                   states={'draft': [('readonly', False)]},
+                                   default=_default_warehouse_id)
+
+    @api.model
+    def create(self, vals):
+        if self.env.context.get('supplier'):
+            vals['name'] = self.env['ir.sequence'].get('rma.order.supplier')
+        else:
+            vals['name'] = self.env['ir.sequence'].get('rma.order')
```
andhit-r, line 92: IMO, the best way to assign sequence are according these flow:

1. Give a default value for name (e.g. /)
2. Remove readonly from field **name**, so user can override the sequence when they see fit
3. If no change needed (field **name** value == /) then assign sequence when data created

Just an opinion, no need to change if you don't see fit.

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+
+    @api.one
+    def _compute_move_count(self):
+        move_list = []
+        for move in self.move_ids:
+            move_list.append(move.id)
+        self.move_count = len(list(set(move_list)))
+
+    @api.model
+    def _default_dest_location_id(self):
```
andhit-r, line 529: This method also unused

rma/models/rma.py
```diff
@@ -0,0 +1,772 @@
...
+    assigned_to = fields.Many2one('res.users', related='rma_id.assigned_to')
+    requested_by = fields.Many2one('res.users', related='rma_id.requested_by')
+    partner_id = fields.Many2one('res.partner', related='rma_id.partner_id')
+    sequence = fields.Integer(default=10,
+                              help="Gives the sequence of this line "
+                              "when displaying the rma.")
+    rma_id = fields.Many2one('rma.order', string='RMA',
+                             ondelete='cascade')
+    uom_id = fields.Many2one('product.uom', string='Unit of Measure')
+    product_id = fields.Many2one('product.product', string='Product',
```
andhit-r, line 607: Readonly unless its draft

andhit-r (changes requested):

rma/models/rma.py
AaronHForgeFlow: Hi @andhit-r Creating manual RMA lines was not possible at that time, but this design is going to be  modified

rma/models/rma.py
AaronHForgeFlow: Thanks for your advice. It could be a good option.

AaronHForgeFlow:
Thank you for you review @andhit-r I hope to include some improvements soon

berpweb:
Guys any idea of pushing this PR forward? 
@aheficent you still working on the improvements?

drl9319:
Hi guys. Now I working in migrate this code to V 8.0, fix some errors and add new funcionality.

I add module: **rma_serial_number**, this module allow select product and serial number, with this data the code can search the related invoice, (for example a big company have more mobile phone and a few moths since the company purchase this mobile have and problem with a unique serial number, but how this company have more invoices and pruchase order can't search the related invoice, for thah is this new funcionality).

The second funcionality is when the company have and rma from a invoice haven't in the system becouse it is from the old system. For thah we have create a new object rma_external_invoice_line with the invoice number, date, client/supplier and serial number.

can be this changes in this url: https://github.com/praxya/rma/tree/8.0-rma-serial-number

regards

AaronHForgeFlow:
Hi @berpweb We are still working on this. Regards.

gurneyalex:
Can you rebase the PR to solve the conflict?

JordiBForgeFlow:
@gurneyalex this is our current working branch https://github.com/Eficent/rma/tree/9.0-rma-v2

As we have deviated significantly from 8.0 we may not finally propose to OCA/rma to respect the current desig, used already by others.

JordiBForgeFlow:
We have decided to continue our diverging approach in https://github.com/Eficent/stock-rma

PR #124, opened 2017-10-05
[IMP] add hook to override field and picking type

bguillot:
Hi

This PR adds a hook method to be able to override the written field in the wizard that creates a picking from the claim.
The PR that used this hook will come next (in the module crm_rma_stock_location)

Thanks for your reviews

### DIFF
--- a/crm_claim_rma/tests/test_picking_creation.py
+++ b/crm_claim_rma/tests/test_picking_creation.py
@@ -21,6 +21,7 @@
 #
 ##############################################################################
 from openerp.tests import common
+from openerp.tools import safe_eval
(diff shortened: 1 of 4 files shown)


sebastienbeau (approved):
Ok for me, travis is red to due pep 8 issue that was already there so not blocking for that PR

pedrobaeza:
For fixing Travis you just need to change the use of 2 evals. Please do it and we leave a green one.

rousseldenis:
@bguillot

rousseldenis:
@bguillot Could you rebase?

PR #131, opened 2018-03-28
[FIX] Fix error that occured when the invoice_id of a crm_claim is empty

daramousk:


crm_claim_rma/wizards/account_invoice_refund.py
```diff
@@ -39,6 +39,6 @@ def _default_description(self):
     def compute_refund(self, mode='refund'):
         self.ensure_one()
         invoice_ids = self.env.context.get('invoice_ids', [])
-        if invoice_ids:
+        if invoice_ids and invoice_ids[0]:
```
hbrunn: better wrap the line above this one in `filter(None, ...)`

But this looks a lot to my like this is fixing symptoms. Could you determine where the falsy id comes from?

hbrunn (changes requested):

crm_claim_rma/wizards/account_invoice_refund.py
daramousk: The problem here is that if the invoice_id on a claim is None then the whole refunding process will not work. You need to provide the invoice of which you want to make a refund if you want to use the standard mechanism, otherwise just create an invoice with the appropriate type `out_refund`. So would it be better if I made the invoice_id field required?

https://github.com/OCA/rma/blob/8.0/crm_claim_rma/views/crm_claim.xml#L87

hbrunn: how about hiding the button if the field is False?

daramousk: Sure, hidden

daramousk:
@hbrunn A lot of errors reported by travis but none of them seems to be related to this change.

hbrunn (approved):

hbrunn:
some of them are really bad, but indeed not caused by this PR, so red CI is unrelated

daramousk:
@hbrunn Is there something else for me to look here or is this ready to be merged?

hbrunn:
nothing to do, but you need another positive review

hbrunn:
as many people (including me) ignore red PRs, you could in another PR fix the problems travis talks about, that one will probably get merged fast, and then rebuild this one to be green

daramousk:
@NL66278 Can I get a review here if possible? Thanks

NL66278 (approved):
:+1:  LGTM

NL66278:
@daramousk Approved, but please also make a PR to make travis green again, as @hbrunn requested. If that is merged you can push again to this branch to get it green as well.

PR #132, opened 2018-05-23
[FIX] Do not put lines that have products of type 'service' in the st…

daramousk:
…ock_moves


pedrobaeza:
Why a claim should contain services products?

Please change module version numbers

daramousk:
Well, can't a user make a claim on a service provided by my company?

pedrobaeza:
This is not claim management, but RMA, where returned stock is involved. Any way, I let others to decide.

daramousk:
@pedrobaeza I understand, will close this since it is not appropriate in rma.

Thanks for your time!

PR #134, opened 2018-09-14
[IMP][9.0] crm_claim_rma: fix return picking choosing wrong package

mpanarin:
Previously creating a return picking with the wizard resulted in picking choosing the wrong package

### DIFF
--- a/crm_claim_rma/models/crm_claim.py
+++ b/crm_claim_rma/models/crm_claim.py
@@ -6,9 +6,6 @@
 
 from openerp import _, api, exceptions, fields, models
 
-from .invoice_no_date import InvoiceNoDate
-from .product_no_supplier import ProductNoSupplier
-
 
 class CrmClaim(models.Model):
     _inherit = 'crm.claim'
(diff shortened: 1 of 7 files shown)


crm_claim_rma/models/crm_claim.py
```diff
@@ -102,60 +99,106 @@ def _onchange_invoice(self):
...
-            for invoice_line in invoices_lines:
-                location_dest = claim_line.get_destination_location(
-                    invoice_line.product_id, warehouse)
+            self._create_claim_lines()
+
+        if self.invoice_id:
+            self.delivery_address_id = self.invoice_id.partner_id.id
+
+    @api.model
+    def __check_packages_enabled(self):
```
gurneyalex: api.model decorator makes the method public and callable from xmlrpc which contradicts the double underscore flagging the method as private.

crm_claim_rma/models/crm_claim.py
```diff
@@ -102,60 +99,106 @@ def _onchange_invoice(self):
...
+            "stock.group_tracking_lot",
+            raise_if_not_found=False,
+        )
+        employee_group = self.env.ref("base.group_user")
+        # if package managing is not activated or group deleted
+        if (not group or
+                group.id not in employee_group.implied_ids.ids or
+                not group.users):
+            return False
+        return True
```
gurneyalex, line 123: why not use "self.user_has_groups('stock.group_tracking_lot')" ?

mpanarin: Because user can be explicitly deleted from the implied group. And the correct packaging should be resolved either way.

yvaucher: Make sense to check the option on the employee group.

yvaucher (approved):
LGTM

crm_claim_rma/models/crm_claim.py
moylop260: This module depends of `stock` if the group is deleted then a `-u stock` will fix the bad practice deleting groups.

moylop260 (approved):
Just small comment (non blocking)

crm_claim_rma/models/crm_claim.py
rousseldenis, line 112: @gurneyalex That's not true. In the call stack, the check on the underscore is done before.

https://github.com/odoo/odoo/blob/c63ac10d9cd83c9fbc7748bdf4f6a87220faa2b1/addons/web/controllers/main.py#L886

crm_claim_rma/wizards/claim_make_picking.py
```diff
@@ -147,6 +146,7 @@ def _get_picking_line_data(self, claim, picking, line):
             'location_id': self.claim_line_source_location_id.id,
             'location_dest_id': self.claim_line_dest_location_id.id,
             'note': self._get_picking_note(),
+            'origin_returned_move_id': line.origin_move.id,
```
grindtildeath, line 149: :warning:  This should only be done in case we're generating a return !

If we want to create a new delivery with the wizard, this original move will restrict the search of the `stock.quant` to be used for the delivery (in `stock.move` `action_assign`), as it will search for a quant having this move in its `history_ids`. However the only quant having this move in its history is on the external location whereas for an outgoing picking, we'll search for a quant in internal location.

grindtildeath (changes requested):
@mpanarin We probably must discuss this together.

grindtildeath (approved):
@mpanarin My concern was totally due to specific overrides of this module, so it's good for me. :+1:

crm_claim_rma/models/crm_claim.py
```diff
@@ -102,61 +99,130 @@ def _onchange_invoice(self):
...
+            location_dest = claim_line.get_destination_location(
+                invoice_line.product_id, warehouse)
+            if not invoice_line.sale_line_ids:
+                continue
+            sale_line = invoice_line.sale_line_ids[0]
+            # here should be only one move
+            if not sale_line.procurement_ids:
+                continue
+            moves = sale_line.procurement_ids[0].move_ids.filtered(
+                lambda x: x.product_id == sale_line.product_id,
```
rousseldenis, line 194: it's a better practice to do:
```suggestion
                lambda x, p=sale_line.product_id: x.product_id == p,
```

rousseldenis (changes requested):
LGTM. Minor comment

rousseldenis:
@mpanarin Do you plan to attend my last comment ?

PR #141, opened 2019-01-28
[FIX] 11.0 build

gurneyalex:
the build of the 11.0 is red because the transfex build crashes on duplicate
translations in ./product_warranty/i18n/es.po

### DIFF
--- a/product_warranty/i18n/es.po
+++ b/product_warranty/i18n/es.po
@@ -7,7 +7,7 @@ msgstr ""
 "Project-Id-Version: product_warranty.es\n"
 "Report-Msgid-Bugs-To: support@openerp.com\n"
 "POT-Creation-Date: 2013-11-21 15:06+0000\n"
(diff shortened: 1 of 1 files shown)


gurneyalex:
see https://travis-ci.org/OCA/rma/jobs/479948863#L1288

gurneyalex:
@sbidoul your opinion on this PR is welcome (removing duplicate entries from a PO file which breaks the build)

sbidoul:
@gurneyalex fixing it manually is the thing to do. This .po has apparently been created or updated manually, this should not happen on .po coming from weblate.

sbidoul (approved):

sbidoul:
@gurneyalex perhaps run `msgmerge --update es.po product_warranty.pot`  to ensure the file is correct and comments are up to date.

rafaelbn (approved):

PR #150, opened 2019-12-27
[rma] adding po file spanish translation draft

Galoperd:


pedrobaeza:
Please use Weblate instead: https://odoo-community.org/blog/the-oca-blog-1/post/translations-platform-81

Galoperd:
Hi, cannot use Weblate because the po files are not there for v.12.

rousseldenis:
Hi @pedrobaeza, it seems, even if some modules are merged in v12, that weblate entry is missing :

image

pedrobaeza:
That's because some PO files are not correct. Travis is red due to this:

https://travis-ci.org/OCA/rma/builds/564794290

Specifically, the errors is in https://travis-ci.org/OCA/rma/jobs/564794292#L1159

Someone must fix it for being in Weblate.

Galoperd:
I would be glad to do this but I do not have edit access rights to the OCA\rma repository.

pedrobaeza:
To solve it you have to do the same as here: create a pull request with the .po fixed

PR #153, opened 2020-05-13
[12.0][ADD] rma: new module

ernestotejeda:
Cc @Tecnativa TT20988


rma/models/res_users.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
chienandalu: ??

rma/models/rma.py
```diff
@@ -0,0 +1,269 @@
...
+    ("replace", "Replace"),
+    ("repair", "Repair")
+]
+
+
+class Rma(models.Model):
+    _name = "rma"
+    _description = "RMA"
+    _order = "priority,date desc"
+    _inherit = ["mail.thread"]
```
chienandalu: `mail.activity.mixin` and `portal.mixin` should be enabled as well, isn't it?

rvalyi:
Hello, not sure if you are aware about the ForgeFlow RMA module https://github.com/ForgeFlow/stock-rma
We were quite pioneers on the RMA topic at Akretion, 1st I tried to reuse the old crm_claim Odoo SA module and that was the origin of the OCA discontinued RMA modules. Later I was told by @sebastienbeau that the ForgeFlow RMA modules were better, so it's definitely worth considering before merging this one... Disclaimer I didn't analyse your code, but I prefer ensure previous work is known.

rma/wizard/rma_re_delivering_views.xml
```diff
@@ -0,0 +1,34 @@
...
+                </group>
+                <footer>
+                    <button name="action_re_deliver" string="Re-deliver" type="object" class="btn-primary"/>
+                    <button string="Cancel" class="btn-secondary" special="cancel" />
+                </footer>
+            </form>
+        </field>
+    </record>
+    <act_window id="rma_re_delivering_wizard_action"
+                name="Re-deliver"
```
chienandalu: Is re-deliver the right term?

rma/models/res_users.py
```diff
@@ -0,0 +1,14 @@
...
+
+from odoo import fields, models
+
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    rma_team_id = fields.Many2one(
+        'crm.team', "User's RMA Team",
```
chienandalu: Should be this, isn't it?
```suggestion
        'rma.team', "User's RMA Team",
```

chienandalu (commented):
Thanks @ernestotejeda 

Some remarks

pedrobaeza:
@rvalyi, we have already studied ForgeFlow module, but it's too rigid as it requires to decide at the beginning which flow you are going to follow via their "strategies", so yes, we are aware of them, but it's not the way we want to go. The final goal is to have the same features (we almost do), but in a flexible workflow that you decide via buttons (but also without a complicated UI). Try yourself and judge (although it's still only draft).

rma/models/res_users.py
ernestotejeda: changed

rma/models/rma.py
ernestotejeda: RMAs are still not visible on the portal but I already added this inheritance

rma/wizard/rma_re_delivering_views.xml
ernestotejeda: I looked in some dictionaries and it seemed a correct term for this (for example: https://en.wiktionary.org/wiki/redeliver), but I have already changed it to avoid confusion.

rma/models/res_users.py
ernestotejeda: yes, changed

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+    ("replace", "Replace"),
+    ("repair", "Repair")
+]
+
+
+class Rma(models.Model):
+    _name = "rma"
+    _description = "RMA"
+    _order = "priority,date desc"
+    _inherit = ["mail.thread", "portal.mixin"]
```
chienandalu: Yo forgot the activity mixin

rma/views/rma_views.xml
```diff
@@ -0,0 +1,232 @@
...
+    </record>
+
+    <record id="rma_view_form" model="ir.ui.view">
+        <field name="name">rma.view.form</field>
+        <field name="model">rma</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
+                    <button type="object"
+                            name="action_rma_send"
```
chienandalu: The method is still missing

rma/views/rma_views.xml
```diff
@@ -0,0 +1,232 @@
...
+                            string="Send by Mail"
+                            attrs="{'invisible':[('sent','=',False)]}" />
+                    <button type="object"
+                            name="action_receive"
+                            states="draft"
+                            string="Receive product"
+                            class="btn-primary" />
+                    <button type="object"
+                            name="action_deliver"
+                            string="Deliver product"
```
chienandalu: What about "RMA Action" or something like that?

rma/views/rma_views.xml
```diff
@@ -0,0 +1,232 @@
...
+                                    <field name="delivery_product_id"/>
+                                </group>
+                                <group>
+                                    <label for="qty_to_deliver"/>
+                                    <div class="o_row">
+                                        <field name="qty_to_deliver"/>
+                                        <field name="delivery_uom"
+                                               groups="uom.group_uom"/>
+                                    </div>
+                                    <label for="delivered_qty"/>
```
chienandalu:
```suggestion
                                    <field name="delivered_qty"/>
```

rma/wizard/rma_delivery.py
```diff
@@ -0,0 +1,110 @@
...
+            delivery_uom = False
+            if self.qty_type == 'fixed_qty':
+                qty_to_deliver = self.product_uom_qty
+                delivery_uom = self.product_uom.id
+            rma.write({
+                "operation": self.operation,
+                "delivery_product_id": self.replace_product_id.id,
+                "qty_to_deliver": qty_to_deliver,
+                "delivery_uom": delivery_uom,
+            })
```
chienandalu: You need to set the RMA state.

Do you think an special state for this case is needed?

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+    def _onchange_move_id(self):
+        if self.move_id:
+            self.product_id = self.move_id.product_id
+        else:
+            self.product_id = False
+
+    @api.onchange("picking_id")
+    def _onchange_picking_id(self):
+        self.move_id = False
+        self.product_id = False
```
chienandalu: Should the RMA's partner change to fit the picking's one?

rma/views/rma_views.xml
```diff
@@ -0,0 +1,232 @@
...
+                            <field name="company_id" groups="base.group_multi_company"/>
+                        </group>
+                    </group>
+                    <notebook>
+                        <page name="reception_page" string="Reception">
+                            <group name="reception_group">
+                                <group>
+                                    <field name="picking_id"
+                                           attrs="{'invisible': [('picking_id', '=', False), ('state', '!=', 'draft')]}"/>
+                                    <field name="move_id"
```
chienandalu: Maybe not very intuitive for the user:
image

chienandalu: Also, the field should have a `options="{'no_create': True}"`

rma/views/rma_views.xml
```diff
@@ -0,0 +1,232 @@
...
+                                           domain="[('picking_id', '=', picking_id)]"
+                                           attrs="{'readonly': [('picking_id', '=', False)], 'required': [('picking_id', '!=', False)], 'invisible': [('move_id', '=', False), ('state', '!=', 'draft')]}"/>
```
chienandalu: Why not making it invisible if not picking_id is set?

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+
+    @api.onchange("product_id")
+    def _onchange_product_id(self):
+        product = self.product_id
+        domain = {
+            'reception_uom': [
```
chienandalu: Maybe a good usability improvement could be to change the `picking_id` and `move_id` domains as well so only those containing the product are available.

And maybe it could be a little bit trickier to combine, but the same combining with a partner's onchange would be amazing.

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+    )
+    state = fields.Selection(
+        [
+            ("draft", "Draft"),
+            ("received", "Received"),
+            ("refunded", "Refunded"),
+            ("delivered", "Delivered"),
+            ("rejected", "Rejected"),
```
chienandalu: I miss a 'cancelled' state and it's actions.

rma/views/rma_views.xml
```diff
@@ -0,0 +1,232 @@
...
+        <field name="binding_model_id" ref="model_rma"/>
+        <field name="state">code</field>
+        <field name="code">records.action_receive()</field>
+    </record>
+
+    <record id="rma_action" model="ir.actions.act_window">
+        <field name="name">RMA</field>
+        <field name="res_model">rma</field>
+        <field name="view_type">form</field>
+        <field name="view_mode">tree,form</field>
```
chienandalu: To have a pivot and a calendar can be quite handy

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+        action["context"]["active_id"] = self.id
+        action["context"]["active_ids"] = self.ids
+        return action
+
+    @api.multi
+    def action_reject(self):
+        self.ensure_one()
+        # - check: if it is possible
+        # - after that: self.state = "rejected"
+        pass
```
chienandalu: Still to do

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+            )
+        fpos = self.partner_id.property_account_position_id
+        if fpos and account:
+            account = fpos.map_account(account)
+        return {
+            'invoice_id': invoice_id,
+            'name': self.product_id.name,
+            'origin': self.name,
+            'account_id': account.id,
+            'price_unit': self.product_id.lst_price,
```
chienandalu: This should hook in a future module with the original sale order line price, if any. Maybe comment the method with something like:
"To be overriden in a third module with the proper origin values in case a sale order is linked to the original move"

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+                self.delivery_product_id,
+                self.qty_to_deliver,
+                self.delivery_uom,
+                self.partner_id.property_stock_customer,
+                self.product_id.name,
+                self.name,
+                values,
+            )
+        except UserError as error:
+            raise UserError(error.name)
```
chienandalu: ?

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+        states={'draft': [('readonly', False)]},
+    )
+    team_id = fields.Many2one(
+        comodel_name="rma.team",
+        string="RMA team",
+        index=True,
+        help="Responsible RMA team."
+             " Define Responsible user and Email account for"
+             " mail gateway.",
+    )
```
chienandalu, line 72: Subscribing users to the team doesn't subscribe them to the team's RMA's. I think it should work the same way other models do (crm.teams, projects...)

rma/models/rma.py
```diff
@@ -0,0 +1,715 @@
...
+            'quantity': qty,
+            'uom_id': self.reception_uom.id,
+            'product_id': self.product_id.id,
+            'rma_id': self.id,
+        }
+
+    # Delivery methods
+    @api.multi
+    def create_repair(self, scheduled_date, warehouse):
+        for record in self:
```
chienandalu: I don't understand very well this flow. For me, it would be:

- operation: repair -> rma state = 'repairing'
- then, you can trigger some possible operations:
  
  1. Repair is done -> re-deliver
  2. Repair isn't possible -> change for another product
  3. Repair isn't possible -> refund

What do you think @pedrobaeza ?

chienandalu (commented):
Thanks @ernestotejeda Some comment with what I could test

- I'd add an smart button in the res.partner form to acces to his RMA's
- A weird behavior (I had to save to be able to set the rma origin move_id:
saving-rma-error

sebastienbeau:
@bguillot, can you take a look to this and test it?

Maybe you can also be interested @max3903 @gurneyalex @AaronHForgeFlow @LoisRForgeFlow

chienandalu (commented):
Some ids can't be loaded.

rma/views/rma_views.xml
```diff
@@ -0,0 +1,269 @@
...
+                                   options="{'no_create': True}"
+                                   groups="base.group_multi_company"/>
+                        </group>
+                    </group>
+                    <notebook>
+                        <page name="page_other"
+                              string="Other Information">
+                            <group>
+                                <group>
+                                    <field name="procurement_group_id"/>
```
rousseldenis, line 196: I think you should explain a little bit as it is a little technical for user. Or maybe rename the label ?

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        domain=[('customer', '=', True)],
+        help="Refund address for current RMA."
+    )
+    picking_id = fields.Many2one(
+        comodel_name="stock.picking",
+        string="Origin Delivery",
+        domain="["
+               "    ('state', '=', 'done'),"
+               "    ('picking_type_id.code', '=', 'outgoing'),"
+               "    ('partner_id', '=', partner_id),"
```
rousseldenis: Maybe a little too restrictive. As partner_id is not required here, no deliveries at all are shown by default.

rma/views/rma_views.xml
```diff
@@ -0,0 +1,269 @@
...
+                            <field name="partner_id" widget="res_partner_many2one" context="{'search_default_customer':1, 'show_address': 1, 'show_vat': True}" options="{'always_reload': True}"/>
+                            <field name="partner_invoice_id"/>
+                            <field name="picking_id"
+                                   options="{'no_create': True}"/>
+                            <field name="move_id"
+                                   attrs="{'required': [('picking_id', '!=', False)]}"
+                                   options="{'no_create': True}"/>
+                            <field name="product_id"
+                                   force_save="1"
+                                   attrs="{'readonly': ['|', ('picking_id', '!=', False), ('state', '!=', 'draft')]}"/>
```
rousseldenis, line 171: As soon as picking is selected, we need to select the move, but maybe you can enhance move name because it's a little bit long and user is focusing on product I think.

image

rma/models/account_invoice.py
```diff
@@ -0,0 +1,32 @@
...
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def action_invoice_open(self):
+        """ Avoids to validate an refund with less quantity of product than
+        quantity in the linked RMA.
+        """
+        if self.mapped('invoice_line_ids').filtered(
+                lambda r: r.rma_id and r.quantity < r.rma_id.product_uom_qty):
```
rousseldenis: You should use float_compare()

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        string='Source Document',
+        states={
+            'locked': [('readonly', True)],
+            'cancelled': [('readonly', True)],
+        },
+        help="Reference of the document that generated this RMA.",
+    )
+    date = fields.Datetime(
+        default=lambda self: fields.Datetime.now(),
+        Index=True,
```
rousseldenis: s/Index/index

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+    @api.multi
+    def action_refund(self):
+        """Invoked when 'Refund' button in rma form view is clicked
+        and 'rma_refund_action_server' server action is run.
+        """
+        refunds = {}
+        for record in self.filtered("can_be_refunded"):
+            refund = refunds.get(record.partner_invoice_id.id)
+            if not refund:
+                invoice_form = Form(self.env['account.invoice'].with_context(
```
rousseldenis: You know I don't like this way, but anyway, you could use a hook function like _prepare_invoice(record) to help override. Same for lines

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        incoming_picking_type = self.env['stock.picking.type'].search(
+            [
+                ('code', '=', 'incoming'),
+                '|',
+                ('warehouse_id.company_id', '=', self.company_id.id),
+                ('warehouse_id', '=', False)
+            ],
+            limit=1,
+        )
+        picking_form = Form(
```
rousseldenis, line 817: same as above

rousseldenis (changes requested):
LGTM. Some functional remarks and technical ones

rma/models/rma.py
chienandalu: Still not working

chienandalu (commented):
Excellent @ernestotejeda 

* I still find it a little bit hard to work with product with serial tracking. There's no relation in the RMA about what serials/lots are being returned and it's only managed at the stock.picking level.
* The subscriptions should work as in other apps.
* The report is not properly aligned:
image

rma/README.rst
```diff
@@ -0,0 +1,179 @@
+===
+RMA
```
pedrobaeza: Return Merchandise Authorization Management

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+    :alt: Try me on Runbot
+
+|badge1| |badge2| |badge3| |badge4| |badge5| 
+
+This module allows you to manage `Return Merchandise Authorization (RMA)
+<https://en.wikipedia.org/wiki/Return_merchandise_authorization>`_.
+RMA documents can be created from scratch, from a delivery order or from
+an incoming email. Product receptions and returning delivery operations
+of the RMA module are fully integrated with the Receipts and Deliveries
+Operations of Odoo inventory core module. It also allows you to generates
```
pedrobaeza:
```suggestion
Operations of Odoo inventory core module. It also allows you to generate
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+refunds in the same way as Odoo generate it. Moreover, you can access to all
+RMA web pages form the portal.
```
pedrobaeza:
```suggestion
refunds in the same way as Odoo generates it.

Besides, you have full integration of the RMA documents in the customer portal.
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+
+Configuration
+=============
+
+If you want RMAs to be created from incoming emails, you need to:
+
+#. Go to *Settings > General Settings*.
+#. Check 'External Email Servers' checkbox under *Discuss* section.
+#. Set an 'alias domain' and an incoming server.
+#. Got to *RMA > Configuration > RMA Team* and select a team or create a new
```
pedrobaeza:
```suggestion
#. Go to *RMA > Configuration > RMA Team* and select a team or create a new
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+
+#. If everything is right, validate the operation and go back to the RMA to
+   see it in a 'received' state.
+
+   .. image:: https://raw.githubusercontent.com/OCA/rma/12.0/rma/static/images/validate_receipt.png
+      :alt: Validate receipt
+
+#. Now you are able to generate a refund, generate a delivery order to return
+   to the customer the same product or another product as a replacement, split
+   the RMA by extracting a part of a remaining quantity to another RMA, preview
```
pedrobaeza:
```suggestion
   the RMA by extracting a part of the remaining quantity to another RMA, preview
```
But put this as subpoints instead everything together in the same main point

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+   the RMA in the website. All of this operations can be done by clicking on
```
pedrobaeza:
```suggestion
   the RMA in the website. All of these operations can be done by clicking on
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+   the buttons in the status bar.
+
+   .. image:: https://raw.githubusercontent.com/OCA/rma/12.0/rma/static/images/operation_on_received.png
+      :alt: Operation on received
+
+   * If you click on 'Refund' button a refund will be create and it will be
```
pedrobaeza:
```suggestion
   * If you click on 'Refund' button, a refund will be created, and it will be
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+     accessible via the smart button labeled Refund. The RMA will be set
+     automatically to 'refunded' state when the refund is validated.
```
pedrobaeza:
```suggestion
     automatically to 'Refunded' state when the refund is validated.
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+
+     .. image:: https://raw.githubusercontent.com/OCA/rma/12.0/rma/static/images/refunded_rma.png
+      :alt: Refunded RMA
+
+   * But, if you click on 'Replace' or 'Return to customer' button instead,
```
pedrobaeza:
```suggestion
   * If you click on 'Replace' or 'Return to customer' button instead,
```

rma/README.rst
```diff
@@ -0,0 +1,179 @@
...
+     a popup wizard will guide you to create a Delivery order to the client
+     and this order will be accessible via the smart button labeled Delivery.
+     The RMA will be set automatically to 'delivered' state when the RMA
+     quantity is delivery in one or more 'done' delivery orders.
```
pedrobaeza:
```suggestion
     The RMA will be set automatically to 'Delivered' state when the RMA
     quantity is equal or lower than the quantity in done delivery orders linked to it.
```

rma/controllers/main.py
```diff
@@ -0,0 +1,110 @@
...
+                report_ref='rma.report_rma_action',
+                download=download,
+            )
+
+        values = self._rma_get_page_view_values(rma_sudo, access_token, **kw)
+        return request.render("rma.portal_rma_page", values)
+
+    @http.route(['/my/picking/pdf/<int:picking_id>'],
+                type='http', auth="public", website=True)
+    def portal_my_picking_report(self, picking_id, access_token=None, **kw):
```
pedrobaeza: The route is the same than https://github.com/odoo/odoo/blob/b88900187b87b2ca8c18d712292a367671a1fff5/addons/sale_stock/controllers/portal.py#L24, so this can collide. I think you should use another route.

The code is also not the same, although it should work the same. For example, there's no real `sudo` here.

rma/data/sequence_data.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <data>
```
pedrobaeza:
```suggestion
    <data noupdate="1">
```
and you can remove `<odoo>` tag as redundant

rma/models/account_invoice.py
```diff
@@ -0,0 +1,32 @@
...
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def action_invoice_open(self):
+        """ Avoids to validate an refund with less quantity of product than
```
pedrobaeza:
```suggestion
        """ Avoids to validate a refund with less quantity of product than
```

rma/models/res_partner.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    rma_count = fields.Integer(
+        string='RMA count',
+        compute='_compute_rma_count',
+    )
+
+    @api.multi
+    def _compute_rma_count(self):
+        for record in self:
+            record.rma_count = len(record.rma_ids)
```
pedrobaeza: Please use `read_group` here instead of this slow way.

rma/models/res_users.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ResUsers(models.Model):
+    _inherit = 'res.users'
+
+    rma_team_id = fields.Many2one(
+        comodel_name='rma.team',
+        string="RMA Team",
+        help='RMA Team the user is member of. '
+             'Used to compute the members of a '
+             'RMA Team through the inverse one2many')
```
pedrobaeza: Don't put this technical part.

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+    origin_split_rma_id = fields.Many2one(
+        comodel_name='rma',
+        string='Extracted from',
+        readonly=True,
+        copy=False,
+    )
+
+    def _compute_delivery_picking_count(self):
+        for record in self:
+            record.delivery_picking_count = len(
```
pedrobaeza: `read_group`

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+
+    @api.multi
+    @api.depends(
+        'refund_id',
+        'delivery_move_ids',
+        'delivery_move_ids.state',
+        'state',
+    )
+    def _compute_can_be_refunded(self):
+        """ Compute 'can_be_refunded'. This field to controls the visibility
```
```suggestion
        """ Compute 'can_be_refunded'. This field controls the visibility
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        of 'Refund' button in the rma form view and to determinate if
```
```suggestion
        of 'Refund' button in the rma form view and determines if
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+            record.can_be_refunded = bool(record.state == 'received'
+                                          and not record.refund_id
+                                          and not delivery_move)
+
+    @api.multi
+    @api.depends('refund_id', 'state')
+    def _compute_can_be_delivered(self):
+        """ Compute 'can_be_delivered'. This field controls the visibility
+        of 'Replace' and 'Return to customer' button in the rma form
+        view and to determinate if an rma can be refunded. It is used in
```
```suggestion
        view and determines if an rma can be refunded. It is used in
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        for record in self:
+            record.can_be_delivered = bool(record.state == 'received'
+                                           and not record.refund_id)
+
+    @api.multi
+    @api.depends('can_be_delivered', 'remaining_qty')
+    def _compute_can_be_split(self):
+        """ Compute 'can_be_delivered'. This field controls the
+        visibility of 'Replace' and 'Return to customer' button
+        in the rma form view and to determinate if an rma can be
```
```suggestion
        in the rma form view and determines if an rma can be
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        refunded. It is used in rma._compute_can_be_split and
```
```suggestion
        split. It is used in rma._compute_can_be_split and
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        for record in self:
+            record.access_url = '/my/rmas/{}'.format(record.id)
+
+    # Constrains methods (@api.constrains)
+    @api.multi
+    @api.constrains('state', 'partner_id', 'partner_invoice_id', 'product_id')
+    def _check_required_after_draft(self):
+        """ Check that RMAs are being created or edited with the
+        necessary fields filled out. Only applies to 'Draft' and
+        'canceled' states.
```
```suggestion
        'Cancelled' states.
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+    @api.multi
+    @api.returns('mail.message', lambda value: value.id)
+    def message_post(self, **kwargs):
+        """ Set 'sent' field to True when an email is sent from rma form
+        view. This field is used to set the appropriate style to the
+        'Send by Email' button in the rma form view.
+        """
+        if self.env.context.get('mark_rma_as_sent'):
+            self.write({'sent': True})
+        self_with_context = self.with_context(mail_post_autofollow=True)
```
pedrobaeza: For what is this needed?

rma/models/rma.py
pedrobaeza: I agree about the hook, but you also know that we prefer this method. I don't know why you are so reluctant. Have you tried it once in a real test?

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+                    default_type='out_refund',
+                    company_id=record.company_id.id,
+                ))
+                invoice_form.company_id = record.company_id
+                invoice_form.partner_id = record.partner_invoice_id
+                invoice_form.origin = record.name
+                invoice_form.user_id = record.user_id
+                refund = invoice_form.save()
+                refunds[record.partner_invoice_id.id] = refund
+            invoice_form = Form(refund)
```
pedrobaeza: Don't do this 2 times. Instead, please keep the same form and return at the end the dictionary values to create the invoice with its lines.

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+
+    @api.multi
+    def action_unlock(self):
+        """Invoked when 'Unlock' button in rma form view is clicked."""
+        locked_rma = self.filtered(lambda r: r.state == 'locked')
+        locked_rma.write({'state': 'received'})
+
+    @api.multi
+    def action_preview(self):
+        """Invoked when 'Unlock' button in rma form view is clicked."""
```
```suggestion
        """Invoked when 'Preview' button in rma form view is clicked."""
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        self.ensure_one()
+        return {
+            'type': 'ir.actions.act_url',
+            'target': 'self',
+            'url': self.get_portal_url(),
+        }
+
+    @api.multi
+    def action_view_receipt(self):
+        """Invoked when 'Rceipt' smart button in rma form view is clicked."""
```
```suggestion
        """Invoked when 'Receipt' smart button in rma form view is clicked."""
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+            action.update(
+                res_id=picking.id,
+                view_mode="form",
+                view_id=False,
+                views=False,
+            )
+        return action
+
+    # Validation business methods
+    @api.multi
```
pedrobaeza: Please remove all `@api.multi` that are not needed anymore in v12.

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+                    _("None of the selected RMAs can be returned "
+                      "to the customer."))
+
+    def _ensure_can_be_split(self):
+        """intended to be called before launch and after save the split wizard.
+        invoked by:
+        rma.action_split
+        rma.extract_quantity
+        """
+        if not self.can_be_split:
```
pedrobaeza, line 760:
```suggestion
        self.ensure_one()
        if not self.can_be_split:
```

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+            custom_values = {}
+        subject = msg_dict.get('subject', '')
+        body = html2plaintext(msg_dict.get('body', ''))
+        desc = _("E-mail subject: %s\n\nE-mail body:\n%s") % (subject, body)
+        defaults = {
+            'description': desc,
+            'name': _('New'),
+            'origin': _('Incoming e-mail'),
+            'product_uom_qty': 1.0,
+            'product_uom': self.env["uom.uom"].search([], limit=1).id,
```
pedrobaeza: Use `unit` UoM by default (`self.env.ref(...)`).

rma/models/rma.py
```diff
@@ -0,0 +1,904 @@
...
+        to be used to create an RMA.
+        """
+        if custom_values is None:
+            custom_values = {}
+        subject = msg_dict.get('subject', '')
+        body = html2plaintext(msg_dict.get('body', ''))
+        desc = _("E-mail subject: %s\n\nE-mail body:\n%s") % (subject, body)
+        defaults = {
+            'description': desc,
+            'name': _('New'),
```
pedrobaeza, line 1073: Shouldn't default acts here without the need of this?

rma/models/stock_move.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def _action_done(self):
+        """ Avoids to validate stock.move with less quantity than the
+        quantity in the linked receiver RMA. It also set the appropriated
+        linked RMA to 'received' or 'delivered'.
+        """
+        for move in self.filtered(
+                lambda r: r.state not in ('done', 'cancel')):
+            rma_receiver = move.rma_receiver_ids
+            if rma_receiver:
```
pedrobaeza: Don't put 2 nested if, but one. Python already aborts as soon as one prior condition is not met.

rma/models/stock_move.py
```diff
@@ -0,0 +1,89 @@
...
+                          "stock move is linked to an RMA (%s).")
+                        % (move.product_id.name, move.rma_receiver_ids.name)
+                    )
+        res = super()._action_done()
+        move_done = self.filtered(lambda r: r.state == 'done')
+        # set RMAs as received
+        to_be_received = move_done.mapped('rma_receiver_ids').filtered(
+            lambda r: r.state == 'confirmed')
+        to_be_received.write({'state': 'received'})
+        # set RMAs as received
```
pedrobaeza:
```suggestion
        # set RMAs as delivered
```

rma/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+    rma_count = fields.Integer(
+        string='RMA count',
+        compute='_compute_rma_count',
+    )
+
+    @api.multi
+    def _compute_rma_count(self):
+        for rec in self:
+            main = rec.move_lines.mapped('rma_ids')
```
pedrobaeza: `read_group`

rma/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+            receiver = rec.move_lines.mapped('rma_receiver_ids')
+            returning = rec.move_lines.mapped('rma_id')
+            rec.rma_count = len(main) + len(receiver) + len(returning)
```
pedrobaeza: RMA smart-button should only cover the generated RMAs from this picking, not the rest. The rest are not significant as smart-buttons. Put the m2o in the move form only.

rma/models/stock_picking.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @api.multi
+    def action_view_rma(self):
+        self.ensure_one()
+        action = self.env.ref('rma.rma_action').read()[0]
+        main = self.move_lines.mapped('rma_ids')
+        receiver = self.move_lines.mapped('rma_receiver_ids')
```
pedrobaeza: Don't add these 2

rma/security/rma_security.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="name">All RMAs</field>
+        <field ref="model_rma" name="model_id"/>
+        <field name="domain_force">[(1,'=',1)]</field>
+        <field name="groups" eval="[(4, ref('rma_group_user_all'))]"/>
+    </record>
+    <!-- New users will belong to rma_group_user_own  -->
+    <record id="base.default_user" model="res.users">
+        <field name="groups_id" eval="[(4, ref('rma_group_user_own'))]"/>
+    </record>
+</odoo>
```
pedrobaeza, line 60: Missing multi-company rules for `rma` and `rma.team`

pedrobaeza (changes requested):
An please attend the rest of the comments

rma/models/rma.py
rousseldenis: I'm not reluctant to it, I'm maybe too psycho-rigid to use something that has been written for tests in real code. Everyone has its defaults :smile:

pedrobaeza: But you have seen that it's not anymore only for tests. It started for that, but can be used for more than that. Please don't stay just in the import.

rma/README.rst
ernestotejeda: To do that I have to change the name in __manifest__.py and that causes the full name of the app not to show in the kanban view. Do i do it anyway?

ernestotejeda: Screenshot 2020-06-10 11:17:43

pedrobaeza: Yes, who look at that view? :stuck_out_tongue_closed_eyes:

rma/models/rma.py
ernestotejeda: That is to include email recipient contacts as RMA followers

rma/models/rma.py
ernestotejeda: I first thought that default values ​​didn't work here, but yes, so I removed those which may be default values

ernestotejeda:
Changes done.
ping @rousseldenis @pedrobaeza @chienandalu

rma/models/rma.py
```diff
@@ -0,0 +1,903 @@
...
+        string="Origin Delivery",
+        domain="["
+               "    ('state', '=', 'done'),"
+               "    ('picking_type_id.code', '=', 'outgoing'),"
+               "    ('partner_id', '=', partner_id),"
+               "    ('partner_id', '!=', False),"
+               "]",
+        readonly=True,
+        states={'draft': [('readonly', False)]},
+    )
```
chienandalu: Better to search any for the main entity.
```suggestion
    commercial_partner_id = fields.Many2one(
        comodel_name="res.partner",
        related="partner_id.commercial_partner_id",
    )
    picking_id = fields.Many2one(
        comodel_name="stock.picking",
        string="Origin Delivery",
        domain="["
               "    ('state', '=', 'done'),"
               "    ('picking_type_id.code', '=', 'outgoing'),"
               "    ('partner_id', 'child_of', commercial_partner_id),"
               "]",
        readonly=True,
        states={'draft': [('readonly', False)]},
    )
```

rma/models/rma.py
chienandalu: @ernestotejeda Maybe we could delegate this domain to the view having two field declarations:

- One with an open domain visible when no partner is set
- One with the restricted domain visible when the partner is set

rma/views/rma_views.xml
```diff
@@ -0,0 +1,269 @@
...
+            </tree>
+        </field>
+    </record>
+
+    <record id="rma_view_form" model="ir.ui.view">
+        <field name="name">rma.view.form</field>
+        <field name="model">rma</field>
+        <field name="arch" type="xml">
+            <form>
+                <header>
```
chienandalu, line 71:
```suggestion
                <header>
                    <button name="%(portal.portal_share_action)d" string="Share" type="action" class="oe_highlight oe_read_only"/>
```

rma/wizard/rma_delivery.py
```diff
@@ -0,0 +1,103 @@
...
+    )
+    product_uom = fields.Many2one(
+        comodel_name="uom.uom",
+        string="Unit of measure",
+    )
+    scheduled_date = fields.Datetime(
+        required=True,
+        default=fields.Datetime.now(),
+    )
+    warehouse_id = fields.Many2one(
```
chienandalu, line 38: Maybe we should choose an operation type instead, so we ensure warehouse and location at once

rma/views/rma_views.xml
chienandalu: This wasn't attended.

rma/models/rma.py
```diff
@@ -0,0 +1,903 @@
...
+        comodel_name="res.users",
+        string="Responsible",
+        track_visibility="always",
+        default=lambda self: self.env.user,
+        states={
+            'locked': [('readonly', True)],
+            'cancelled': [('readonly', True)],
+        },
+    )
+    team_id = fields.Many2one(
```
chienandalu, line 64: I'm missing to be able to decide where to receive the RMA. Maybe having the option to pick the operation (tha could be even a default for the team) would be the more direct way.

rma/models/stock_move.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+
+    def unlink(self):
+        if self.mapped('rma_receiver_ids'):
+            raise ValidationError(
+                _("There is at least one move you are trying to "
+                  "delete that is a reception of one or more RMAs."))
+        return super().unlink()
+
+    def _action_done(self):
```
chienandalu, line 49: When we cancel the move, we should change the state of the rma as well, allowing to make a new reception.

rma/models/rma.py
rousseldenis: Is it wanted ?

rma/models/stock_move.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        for move in self.filtered(
+                lambda r: r.state not in ('done', 'cancel')):
+            rma_receiver = move.rma_receiver_ids
+            if (rma_receiver
+                    and move.quantity_done != rma_receiver.product_uom_qty):
+                raise ValidationError(
+                    _("The quantity done for the product '%s' must "
+                      "be equal to its initial demand because the "
+                      "stock move is linked to an RMA (%s).")
```
chienandalu, line 62: Maybe at this stage we should be able to change the rma product quantity to fit the incoming quantity.

Other (maybe a little more complicated option) would be split the RMA along with the backorder.

rma/models/rma.py
```diff
@@ -0,0 +1,903 @@
...
+from odoo.tests import Form
+from odoo.tools import html2plaintext
+import odoo.addons.decimal_precision as dp
+from odoo.addons.stock.models.stock_move import PROCUREMENT_PRIORITIES
+
+
+class Rma(models.Model):
+    _name = "rma"
+    _description = "RMA"
+    _order = "priority,date desc"
```
chienandalu: Otherwise the order is a little bit weird :)
```suggestion
    _order = "date desc, priority"
```

chienandalu (changes requested):
Thanks @ernestotejeda I made a complete functional flow. Some remarks (nothing really critical):

chienandalu:
By the way, resolve attended comments so the PR (that is already very long) is easier to follow :)

rma/models/rma.py
chienandalu: Well, maybe it's an overcomplication (it would require additional onchanges, etc) as RMA's born from the customer anyway. The forget my comment here @ernestotejeda :)

rma/views/rma_views.xml
ernestotejeda: I have added this in the ROADMAP as an improvement to be made

chienandalu: Ok, then. Resolve, please ;)

rma/models/rma.py
pedrobaeza: Yeah, let's keep it this way for now.

rma/models/rma.py
pedrobaeza: Please put then a comment saying so

rma/models/stock_move.py
pedrobaeza: No, if the quantity doesn't fit, remove the RMA and create another one after the picking backorder split.

ernestotejeda:
Changes done.

rma/wizard/rma_delivery.py
chienandalu: Thanks!

rma/models/stock_move.py
chienandalu: Ok, maybe a little bit rigid, but not a big deal

rma/models/rma.py
```diff
@@ -0,0 +1,954 @@
...
+        picking_id = picking_action['res_id']
+        picking = self.env['stock.picking'].browse(picking_id)
+        picking.origin = "%s (%s)" % (self.name, picking.origin)
+        move = picking.move_lines
+        move.priority = self.priority
+        return move
+
+    def _create_receptions_from_product(self):
+        self.ensure_one()
+        incoming_picking_type = self.env['stock.picking.type'].search(
```
chienandalu: For this case it seems quite trivial to have an operation_type to choose in the rma and default to these values if it isn't set

rma/models/rma.py
```diff
@@ -0,0 +1,954 @@
...
+        return_wizard = self.env['stock.return.picking'].with_context(
+            active_id=self.picking_id.id,
+            active_ids=self.picking_id.ids,
+        ).create({})
+        return_wizard.product_return_moves.filtered(
+            lambda r: r.move_id != self.move_id
+        ).unlink()
+        return_line = return_wizard.product_return_moves
+        return_line.quantity = self.product_uom_qty
+        picking_action = return_wizard.create_returns()
```
chienandalu: For this case, maybe it's as easy as injecting a context to override this `copy()` and set the `picking_type` if any: https://github.com/odoo/odoo/blob/12.0/addons/stock/wizard/stock_picking_return.py#L112-L118

chienandalu (approved):
I tested last functional points in discussion.

Thanks Ernesto, excellent job :+1:

pedrobaeza:
@rousseldenis can you recheck your review?

rousseldenis (approved):
LGTM. Let's go for this first version!

pedrobaeza (approved):
Yes, good for a first version, that it's anyway very complete! Thanks to all the involved people.

pedrobaeza:
Fixed problem with PO in other module in 1e698904f9fb008c3097bb7deef80f5a98b31564, so

/ocabot merge nobump

PR #158, opened 2020-08-14
[FIX] rma: views permissions

chienandalu:
Regular users don't have permissions to rma models, so we should avoid
loading views that lead to permission errors.

cc @Tecnativa TT24986


chienandalu:
ping @ernestotejeda @pedrobaeza

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this bugfix:

/ocabot merge minor

PR #159, opened 2020-08-14
[FIX] rma_sale: portal permissions

chienandalu:
Some operations need to be sudoed to be reachable by the portal user

cc @Tecnativa TT24986

### DIFF
--- a/rma_sale/controllers/sale_portal.py
+++ b/rma_sale/controllers/sale_portal.py
@@ -25,12 +25,13 @@ class CustomerPortal(CustomerPortal):
             mapped_vals.setdefault(row, {}).update({field_name: value})
         line_vals = [(0, 0, vals) for vals in mapped_vals.values()]
         # Create wizard an generate rmas
-        location_id = order_obj.browse(order_id).warehouse_id.rma_loc_id.id
+        order = order_obj.browse(order_id).sudo()
+        location_id = order.warehouse_id.rma_loc_id.id
         wizard = wizard_obj.with_context(active_id=order_id).create({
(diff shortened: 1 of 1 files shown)


chienandalu:
Sorry about the double PR

rma_sale/controllers/sale_portal.py
```diff
@@ -25,12 +25,13 @@ def request_rma(self, order_id, access_token=None, **post):
...
         # Create wizard an generate rmas
-        location_id = order_obj.browse(order_id).warehouse_id.rma_loc_id.id
+        order = order_obj.browse(order_id).sudo()
+        location_id = order.warehouse_id.rma_loc_id.id
         wizard = wizard_obj.with_context(active_id=order_id).create({
             'line_ids': line_vals,
             'location_id': location_id
         })
-        rma = wizard.create_rma(from_portal=True)
+        rma = wizard.sudo().create_rma(from_portal=True)
```
pedrobaeza, line 34: The problem with this is that the create user will be system, not the proper user. Fortunately, this won't happen in v13, right?

pedrobaeza (approved):

rma_sale/controllers/sale_portal.py
chienandalu: > Fortunately, this won't happen in v13, right?

Why not?

pedrobaeza: `sudo()` only change the permissions, not the uid.

chienandalu: :+1:

PR #160, opened 2020-08-17
[FIX] rma: portal views access errors

chienandalu:
- Portal mail thread needs token config.
- Unpublished products will raise AccessError on RMAs portal views for
portal users due to record rules.
- Set split window action domain to avoid context weird behavior

cc @Tecnativa TT24986

### DIFF
--- a/rma/views/rma_portal_templates.xml
+++ b/rma/views/rma_portal_templates.xml
@@ -50,7 +50,8 @@
                             </a>
                         </td>
                         <td class="d-none d-md-table-cell"><span t-field="rma.date"/></td>
-                        <td><span t-field="rma.product_id.name"/></td>
+                        <!-- Portal users don't have access to unpublished products -->
+                        <td><span t-esc="rma.sudo().product_id.display_name"/></td>
                         <td class='text-right'><span t-field="rma.product_uom_qty"/></td>
                         <td class="d-none d-md-table-cell tx_status">
                             <span class="badge badge-pill badge-secondary"><span t-field="rma.state"/></span>
@@ -132,12 +133,14 @@
                                         <span t-field="rma.picking_id"/>
                                     </div>
                                 </div>
-                                <div t-if="rma.product_id" class="row mb-2 mb-sm-1">
+                                <!-- We need to prevent access errors if the product is
+                                     unpublished-->
+                                <div t-if="rma.sudo().product_id" class="row mb-2 mb-sm-1">
                                     <div class="col-12 col-sm-4">
                                         <strong>Product</strong>
                                     </div>
                                     <div class="col-12 col-sm-8">
-                                        <span t-field="rma.product_id"/>
+                                        <span t-esc="rma.sudo().product_id.display_name"/>
                                     </div>
                                 </div>
                                 <div t-if="rma.product_uom_qty" class="row mb-2 mb-sm-1">
@@ -262,6 +265,9 @@
                 <h2>Communication</h2>
                 <t t-call="portal.message_thread">
                     <t t-set="object" t-value="rma"/>
+                    <t t-set="token" t-value="rma.access_token"/>
+                    <t t-set="pid" t-value="pid"/>
+                    <t t-set="hash" t-value="hash"/>
                 </t>
             </div>
         </t>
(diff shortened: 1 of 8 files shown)


rma/views/rma_portal_templates.xml
```diff
@@ -262,6 +269,10 @@
                 <h2>Communication</h2>
                 <t t-call="portal.message_thread">
                     <t t-set="object" t-value="rma"/>
+                    <t t-set="object" t-value="rma"/>
+                    <t t-set="token" t-value="rma.access_token"/>
```
pedrobaeza, line 268: Do we have "Share" button in the backend?

pedrobaeza (approved):

rma/views/rma_portal_templates.xml
chienandalu: Yes

kos94ok-3D: duplicated line `<t t-set="object" t-value="rma"/>`

chienandalu: Fixed. Thanks @kos94ok-3D

rma/views/rma_portal_templates.xml
```diff
@@ -132,12 +133,20 @@
...
+                                <!-- We need to prevent access errors if the product is
+                                     unpublished. If not, we'll sudo to get the name-->
+                                <div t-if="rma.sudo().product_id" class="row mb-2 mb-sm-1">
                                     <div class="col-12 col-sm-4">
                                         <strong>Product</strong>
                                     </div>
                                     <div class="col-12 col-sm-8">
-                                        <span t-field="rma.product_id"/>
+                                        <!-- Don't depend on website -->
+                                        <t t-if="'website_published' in rma.sudo().product_id._fields and rma.sudo().product_id.website_published">
```
pedrobaeza: Why not simply put the regular one? How does `sale` works with this?

rma/views/rma_portal_templates.xml
```diff
@@ -132,12 +133,20 @@
...
+                                            <span t-field="rma.product_id"/>
+                                        </t>
+                                        <t t-else="">
+                                            <span t-esc="rma.sudo().product_id.name"/>
```
pedrobaeza: It must be `display_name`

pedrobaeza (changes requested):

rma/views/rma_portal_templates.xml
chienandalu: they put line.name, wich isn't valid here

pedrobaeza: OK, but I think you should simply avoid the website_published stuff. Why `t-field` is not valid in every case?

chienandalu: `website` adds this record rule: https://github.com/odoo/odoo/blob/12.0/addons/website/security/website_security.xml#L56-L61 So if the portal tries to access those records, it will raise an AcessError.

Another option to avoid this would be to create a compute field with the product name

pedrobaeza: Well, you have put website.page record rule, which doesn't apply here, but I suppose you mean website_sale record rule over products. Let's put directly the t-esc over display name with the sudoed record

chienandalu: Perfect. I let it like that

chienandalu:
Changes done. I detected another issue with portal pickings that I assigned to @ernestotejeda Should we wait for his changes?

pedrobaeza (approved):

chienandalu:
@pedrobaeza I've added a new commit fixing RMA Teams flows: https://github.com/OCA/rma/pull/160/commits/d4e316313a4989806fc2e6f935081fb3e9851658

pedrobaeza (approved):

pedrobaeza:
Are all the notified problems now solved?

chienandalu:
There are still two issues with Access errors in the portal views that I assigned to @ernestotejeda

pedrobaeza:
OK, let's merge this and let @ernestotejeda to open other PR.

/ocabot merge minor.

PR #161, opened 2020-08-19
[12.0][FIX+IMP] rma, rma_sale: fix bugs and add improvements

ernestotejeda:
- Fix thrown error when trying to download a picking from the portal.
- Add the hook method to prepare RMA values ​​from the return pick wizard.
- Add the access rule for portal users.
- Show the portal 'Request RMAs' button on the sales page only to users
related to the sales order.

cc @Tecnativa TT25187

### DIFF
--- a/rma/controllers/main.py
+++ b/rma/controllers/main.py
@@ -102,13 +102,13 @@ class PortalRma(CustomerPortal):
         values = self._rma_get_page_view_values(rma_sudo, access_token, **kw)
         return request.render("rma.portal_rma_page", values)
 
-    @http.route(['/my/rma/picking/pdf/<int:picking_id>'],
+    @http.route(['/my/rma/picking/pdf/<int:rma_id>/<int:picking_id>'],
                 type='http', auth="public", website=True)
-    def portal_my_rma_picking_report(self, picking_id, access_token=None,
-                                     **kw):
+    def portal_my_rma_picking_report(self, rma_id, picking_id,
+                                     access_token=None, **kw):
         try:
-            picking_sudo = self._stock_picking_check_access(
-                picking_id, access_token=access_token)
+            picking_sudo = self._picking_check_access(
+                rma_id, picking_id, access_token=access_token)
         except exceptions.AccessError:
             return request.redirect('/my')
         report_sudo = request.env.ref('stock.action_report_delivery').sudo()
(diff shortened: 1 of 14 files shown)


rma/models/rma_team.py
```diff
@@ -8,9 +8,8 @@ class RmaTeam(models.Model):
     _name = "rma.team"
     _inherit = ['mail.alias.mixin', 'mail.thread']
     _description = "RMA Team"
-    _order = "sequence, name"
+    _order = "name"
 
-    sequence = fields.Integer()
```
chienandalu: Why to remove this?

ernestotejeda: I'm sorry, I added it again

chienandalu (approved):
Reviewed. Thanks!

rma/models/rma_team.py
chienandalu: :smile: :+1:

rma_sale/views/sale_portal_template.xml
```diff
@@ -1,7 +1,11 @@
 <odoo>
     <template id="sale_order_portal_template" name="Request RMA" inherit_id="sale.sale_order_portal_template">
         <xpath expr="//div[hasclass('o_portal_sale_sidebar')]//div[hasclass('o_download_pdf')]/.." position="after">
-            <li t-if="sale_order.state in ['sale', 'done']" class="list-group-item flex-grow-1" id="li-request-rma">
+            <!-- If the sale order was not linked to the partner of the logged user, the button will not be shown -->
+            <li t-if="sale_order.state in ['sale', 'done'] and env.user.partner_id.commercial_partner_id == sale_order.partner_id.commercial_partner_id"
```
pedrobaeza: I don't think we should restrict this: if you are able to see the sales orders due to the permissions, then we can request an RMA.

pedrobaeza (changes requested):
The portal record rule for RMAs is still missing

rma_sale/views/sale_portal_template.xml
chienandalu: Good point

chienandalu:
> The portal record rule for RMAs is still missing

I see it here: https://github.com/OCA/rma/pull/161/files#diff-3f3c73a654f2fb27560b5b8a8834b6e0R44-R49

pedrobaeza:
Oh, I don't know how I missed it. The other comment still applies though.

ernestotejeda:
@pedrobaeza @chienandalu Changes done. Thanks.

pedrobaeza (approved):

PR #162, opened 2020-08-24
[FIX] rma: permissions for general users

chienandalu:
- Avoid permission blockings on users who doesn't have RMA permissions.
- Add activities view on RMA list.

cc @Tecnativa TT25229

### DIFF
--- a/rma/models/account_invoice.py
+++ b/rma/models/account_invoice.py
@@ -15,14 +15,16 @@ class AccountInvoice(models.Model):
         """
         precision = self.env['decimal.precision'].precision_get(
             'Product Unit of Measure')
-        if self.mapped('invoice_line_ids').filtered(
+        if self.sudo().mapped('invoice_line_ids').filtered(
                 lambda r: (r.rma_id and float_compare(
                     r.quantity, r.rma_id.product_uom_qty, precision) < 0)):
             raise ValidationError(
                 _("There is at least one invoice lines whose quantity is "
                   "less than the quantity specified in its linked RMA."))
         res = super().action_invoice_open()
-        self.mapped('invoice_line_ids.rma_id').write({'state': 'refunded'})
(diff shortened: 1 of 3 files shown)


rma/security/ir.model.access.csv
```diff
@@ -2,6 +2,7 @@ id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_rma_team_user_own,rma.team.user.own,model_rma_team,rma_group_user_own,1,0,0,0
 access_rma_team_manager,rma.team.manager,model_rma_team,rma_group_manager,1,1,1,1
 access_rma_portal,rma.portal,model_rma,base.group_portal,1,0,0,0
+access_rma_internal_user,rma.internal_user,model_rma,base.group_user,1,0,0,0
```
pedrobaeza: This permission shouldn't exist.

chienandalu: Well, is to avoid patching with `sudo()` every possible method reading an `rma` related field. It's harmless, as it's needed the proper record rule to access to any information.

pedrobaeza: I prefer to not give that one, as people can modify record rules, but not ACLs, so you can find that someone without RMA permissions is able to see some of them after any modification of the record rules, thinking that only affects RMA users.

ernestotejeda (commented):
Shouldn't we do the same here?
https://github.com/OCA/rma/blob/12.0/rma/models/account_invoice.py#L12

rma/security/ir.model.access.csv
chienandalu: Ok, changes done.

chienandalu:
Yes, @ernestotejeda Thanks :)

pedrobaeza (approved):

pedrobaeza:
I fast-tracked, as this is a needed bugfix:

/ocabot merge patch

PR #163, opened 2020-08-28
[IMP] rma_sale: portal improvents

chienandalu:
- Now it's possible to open several RMAs in a sale order from the portal
- A new comment button has been added to allow the portal user to enter
relevant information like serial numbers o issue description.
- If the requested operation isn't set no RMA will be opened
- The RMA product qty is now a numeric control with limits according to
the qty available to return

cc @Tecnativa TT24986

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -503,6 +503,7 @@ class Rma(models.Model):
             vals["team_id"] = self.env["rma.team"].search([], limit=1).id
         return super().create(vals)
 
+    @api.multi
     def copy(self, default=None):
         team = super().copy(default)
         for follower in self.message_follower_ids:
--- a/rma/models/rma_team.py
+++ b/rma/models/rma_team.py
@@ -1,7 +1,7 @@
 # Copyright 2020 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import _, fields, models
+from odoo import _, api, fields, models
 
 
 class RmaTeam(models.Model):
@@ -37,7 +37,9 @@ class RmaTeam(models.Model):
         string='Team Members',
     )
 
+    @api.multi
     def copy(self, default=None):
+        self.ensure_one()
         if default is None:
             default = {}
         if not default.get('name'):
--- a/rma_sale/README.rst
+++ b/rma_sale/README.rst
@@ -64,6 +64,15 @@ The customer can also create RMAs from a sales order portal page:
 #. Click on the 'Request RMAs' button and RMAs will be created linked to
    the sales order.
 
+Known issues / Roadmap
+======================
+
+* When you try to request an RMA from a Sales Order in the portal,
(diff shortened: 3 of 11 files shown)


rafaelbn:
@ernestotejeda please we need your deep review here. Technical an functional. Thanks

ernestotejeda (approved):

ernestotejeda (commented):
Code and functional review :+1:

rafaelbn (approved):
:+1:

rma/models/rma.py
```diff
@@ -503,6 +503,7 @@ def create(self, vals):
             vals["team_id"] = self.env["rma.team"].search([], limit=1).id
         return super().create(vals)
 
+    @api.multi
```
pedrobaeza, line 506: Don't add this. It's the default.

rma/models/rma_team.py
```diff
@@ -37,6 +37,7 @@ class RmaTeam(models.Model):
         string='Team Members',
     )
 
+    @api.multi
```
pedrobaeza, line 40: Don't add this. It's the default.

rma_sale/README.rst
```diff
@@ -64,6 +64,15 @@ The customer can also create RMAs from a sales order portal page:
 #. Click on the 'Request RMAs' button and RMAs will be created linked to
    the sales order.
 
+Known issues / Roadmap
+======================
+
+* When you try to request an RMA from a Sales Order in the portal,
+  now a popup appears and the inputs for the quantity doesn't allow
```
pedrobaeza: Remove the `now` word

pedrobaeza (changes requested):
An empty block with the `id` and `class` prepared for website can be added for allowing to customize this with the CMS.

ernestotejeda:
@pedrobaeza , I removed @api.multi statements but I got this error from test_flake8 '...Missing api.one or api.multi in copy function' (https://travis-ci.org/github/OCA/rma/jobs/724942370#L332)

ernestotejeda:
> An empty block with the `id` and `class` prepared for website can be added for allowing to customize this with the CMS.

@pedrobaeza , are you talking about the message in the rma request popup? If so, I don't see how that message could be edited, because it is a popup, when it appears, you cannot click on the 'Edit' button to edit the website

pedrobaeza:
OK, if it's a popup, indeed it can't be customized. Take it into account for website_rma then.

pedrobaeza:
Put `@api.one` in the copy.

pedrobaeza (approved):

pedrobaeza:
Please check Travis

ernestotejeda:
@pedrobaeza , when I use the api.one decorator for the copy method, what is returned is a list instead of a recodset with one element and that cause errors.
I have left the api.multi but with the ensure_one() method.
Check it out.

marcelsavegnago (approved):

pedrobaeza (approved):

PR #164, opened 2020-09-04
[ADD] website_rma: new module

ernestotejeda:
This module adds a website form to allows to create an RMA from scratch.

Peek 2020-09-04 13-48

cc @Tecnativa TT24986

### DIFF
--- a/website_rma/controllers/main.py
+++ b/website_rma/controllers/main.py
@@ -1,0 +1,41 @@
+# Copyright 2020 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import json
+from odoo import http
+from odoo.addons.website_form.controllers.main import WebsiteForm
+from odoo.http import request
+
+
+class WebsiteForm(WebsiteForm):
+
+    def insert_record(self, request, model, values, custom, meta=None):
+        if model.model == 'rma':
+            values['partner_id'] = request.env.user.partner_id.id
+        return super(WebsiteForm, self).insert_record(
+            request, model, values, custom, meta=meta)
+
+
+class WebsiteRMA(http.Controller):
+
+    def _get_website_rma_product_domain(self, q):
+        """Domain used for the products to be shown in selection of
+        the web form.
+        """
+        return [
+            ('name', '=ilike', "%{}%".format(q or '')),
+            ("sale_ok", "=", True),
+        ]
+
+    @http.route(['/requestrma'], type='http', auth="user", website=True)
+    def request_rma(self, **kw):
+        return http.request.render("website_rma.request_rma", {})
+
+    @http.route('/website_rma/get_products', type='http', auth="user",
+                methods=['GET'], website=True)
+    def rma_product_read(self, q='', l=25, **post):
+        data = request.env['product.product'].sudo().search_read(
+            domain=self._get_website_rma_product_domain(q),
+            fields=['id', 'display_name', 'uom_id'],
+            limit=int(l),
+        )
+        return json.dumps(data)
--- a/website_rma/views/request_rma_form.xml
+++ b/website_rma/views/request_rma_form.xml
@@ -1,0 +1,148 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+
+    <template id="request_rma" name="Request RMA">
+        <t t-call="website.layout">
+            <div id="wrap">
+                <div class="oe_structure mt-2" id="oe_structure_website_rma_form_request_0"/>
+                <section class="pt8 pb8">
+                    <div class="container">
+                        <div class="row s_nb_column_fixed">
+                            <div class="col-lg-12 s_title pt16 pb16">
+                                <h1>Request RMA</h1>
+                            </div>
+                        </div>
+                    </div>
+                </section>
+                <div class="container mt-2">
+                    <div class="row">
+                        <div class="col-lg-8">
+                            <div class="oe_structure mt-2" id="oe_structure_website_rma_form_request_1"/>
+                            <section class="s_text_block">
+                                <div class="container">
+                                    <div class="row">
+                                        <div class="col-lg-12">
+                                            <p>
+                                                Our team will process it an will reach you once it's validated.
+                                            </p>
+                                        </div>
+                                    </div>
+                                </div>
+                            </section>
+                            <div class="request_rma_container">
+                                <form action="/website_form/"
+                                      method="post"
+                                      data-model_name="rma"
+                                      data-success_page="/requestrma-thank-you"
+                                      class="s_website_form container-fluid mt32"
+                                      enctype="multipart/form-data"
+                                      data-editable-form="false"
+                                      id="rma_request_form">
+                                    <div class="form-group row">
+                                        <label class="col-lg-3 col-md-4 col-form-label"
+                                               for="product_id">
+                                            Product
+                                        </label>
+                                        <div class="col-lg-7 col-md-8">
+                                            <input type="hidden"
+                                                   name="product_id"
+                                                   class="form-control o_website_form_input"/>
+                                        </div>
+                                    </div>
+                                    <div class="form-group row form-field o_website_form_required">
+                                        <label class="col-lg-3 col-md-4 col-form-label"
+                                               for="product_uom_qty">
+                                            Quantity
+                                        </label>
+                                        <div class="col-lg-4 col-md-5">
+                                            <input class="form-control o_website_form_input"
+                                                   pattern="^\d*(\.\d{0,3})?$"
+                                                   name="product_uom_qty"
+                                                   required=""
+                                                   t-att-value="request.params.get('product_uom_qty', '1.000')"/>
+                                        </div>
+                                        <div class="col-lg-3 col-md-3">
+                                            <input type="hidden" name="product_uom"/>
+                                            <input type="text"
+                                                   class="form-control o_website_form_input"
+                                                   name="product_uom_name"
+                                                   disabled="true"/>
+                                        </div>
+                                    </div>
+                                    <div class="form-group row form-field">
+                                        <label class="col-lg-3 col-md-4 col-form-label" for="operation_id">Operation</label>
+                                        <div class="col-lg-7 col-md-8">
+                                            <select name="operation_id"
+                                                    class="form-control custom-select o_website_form_input">
+                                                <option/>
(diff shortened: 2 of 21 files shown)


oca_dependencies.txt
```diff
@@ -0,0 +1,15 @@
...
+#
+# To depend on the standard version of sale-workflow, use:
+# sale-workflow
+#
+# To explicitely give the URL of a fork, and still use the version specified in
+# .travis.yml, use:
+# sale-workflow https://github.com/OCA/sale-workflow
+#
+# To provide both the URL and a branch, use:
+# sale-workflow https://github.com/OCA/sale-workflow branchname
```
chienandalu: Why is necessary?

website_rma/views/request_rma_form.xml
```diff
@@ -0,0 +1,174 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <data>
```
chienandalu: This tag is uneeded

website_rma/views/request_rma_form.xml
```diff
@@ -0,0 +1,174 @@
...
+        <record id="request_rma_page_view" model="ir.ui.view">
```
chienandalu: Check the operations that need  `sudo` or a portal user will get a permissions error.

website_rma/views/website_rma_portal_templates.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="portal_my_home_rma" name="Portal My Home : RMA entries" inherit_id="portal.portal_my_home" priority="30">
```
chienandalu, line 5: Isn't this a little bit out of place?

website_rma/views/website_templates.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <data>
+        <template id="contact_us_request_rma" name="Request RMA" inherit_id="website.contactus">
```
chienandalu: Same here, maybe not the best place to put it

chienandalu (changes requested):

chienandalu (changes requested):
I couldn't select products with the portal user (surely due to lack of permissions)

ernestotejeda:
@chienandalu Ready to review. Now only logged in users can access the form.

website_rma/views/website_templates.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2020 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="contact_us_request_rma" name="Request RMA" inherit_id="website.contactus">
+        <xpath expr="//div[@name='mail_button']" position="after">
```
jesusVMayor: don't work with website_crm module, it removes the mail_button div.

victoralmau:
Reviewing this new functionality, i think that in other future version can also interesting to add other addon related to show stock_picking in website and integrate the RMA form in every stock_picking done.

Only is a suggestion that maybe simplify the process for the users.

chienandalu:
> I couldn't select products with the portal user (surely due to lack of permissions)

@ernestotejeda This comment wasn't attended
Peek 11-09-2020 13-40

website_rma/controllers/main.py
```diff
@@ -0,0 +1,32 @@
...
+            request, model, values, custom, meta=meta)
+
+
+class WebsiteRMA(http.Controller):
+
+    @http.route(['/requestrma'], type='http', auth="user", website=True)
+    def request_rma(self, **kw):
+        return http.request.render("website_rma.request_rma", {})
+
+    @http.route('/website_rma/get_products', type='http', auth="public",
```
chienandalu: shouldn't it be auth="user" as well?

ernestotejeda: I think so.

ernestotejeda:
@chienandalu changes done

chienandalu:
> @chienandalu changes done

Mmm...

image

(tested with portal user)

ernestotejeda:
I have limited the domain of the products to choose from on the website form. @pedrobaeza , could you take a look at that?.
@chienandalu the problem with portal user is fixed, check it out. Remember that now there must be a delivery order to the demo user's partner for selecting a product.

website_rma/controllers/main.py
```diff
@@ -0,0 +1,48 @@
...
+        if model.model == 'rma':
+            values['partner_id'] = request.env.user.partner_id.id
+        return super(WebsiteForm, self).insert_record(
+            request, model, values, custom, meta=meta)
+
+
+class WebsiteRMA(http.Controller):
+
+    def _get_website_rma_product_domain(self, q):
+        delivered_products = request.env["stock.move"].search([
```
HviorForgeFlow: You should also add a sudo in this search to avoid this ACL error:
```
{AccessError}('Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:\n- Sales/Manager\n\t- Sales/User: Own Documents Only\n\t- Accounting & Finance/Billing\n\t- Inventory/User\n\t- Inventory/Manager\n\n(Document model: stock.move)', None)
```

HviorForgeFlow (changes requested):

website_rma/controllers/main.py
pedrobaeza: In fact this domain is not shouldn't be built this way, as you can request an RMA for an operation not registered in Odoo (old things before switching ERP), so I don't think we should restrict this.

ernestotejeda:
@pedrobaeza , I have changed the domain.

website_rma/controllers/main.py
```diff
@@ -0,0 +1,41 @@
...
+    @http.route(['/requestrma'], type='http', auth="user", website=True)
+    def request_rma(self, **kw):
+        return http.request.render("website_rma.request_rma", {})
+
+    @http.route('/website_rma/get_products', type='http', auth="user",
+                methods=['GET'], website=True)
+    def rma_product_read(self, q='', l=25, **post):
+        data = request.env['product.product'].sudo().search_read(
+            domain=self._get_website_rma_product_domain(q),
+            fields=['id', 'name'],
```
chienandalu: Maybe `display_name` would be bettere as otherwise, the user can't tell which variant corresponds to which option:

 
image

website_rma/views/request_rma_form.xml
```diff
@@ -0,0 +1,160 @@
...
+                                                   pattern="^\d*(\.\d{0,3})?$"
+                                                   name="product_uom_qty"
+                                                   required=""
+                                                   t-att-value="request.params.get('product_uom_qty', '1.000')"/>
+                                        </div>
+                                        <div class="col-lg-3 col-md-3">
+                                            <t t-set="uom_unit_id" t-value="request.env.ref('uom.product_uom_unit').id"/>
+                                            <select name="product_uom" class="form-control custom-select o_website_form_input">
+                                                <option/>
+                                                <t t-foreach="request.env['uom.uom'].sudo().search([])" t-as="uom">
```
chienandalu: I find this a little bit missleading, to be honest, as it throws all the uom availble and it can lead to errors. Maybe it would be better to just put the selected product default sale UoM

image

chienandalu (changes requested):
A couple more things:

ernestotejeda:
@chienandalu , Done.

website_rma/views/website_templates.xml
chienandalu: I just tested and they seem to work fine altogether

chienandalu (approved):
Ok :+1: 

Let's move on!

pedrobaeza (changes requested):
Please add placeholder empty containers (with class="oe_structure" and an specific id if I don't remember bad) before and after each hardcoded paragraph for allowing to drag and drop website blocks without impacting this view (putting it with noupdate=1).

website_rma/views/website_templates.xml
chienandalu: Ok, @jesusVMayor I just reproduced the issue :) @ernestotejeda You should take an alternative hook

website_rma/views/request_rma_form.xml
```diff
@@ -0,0 +1,150 @@
...
+                                    </div>
+                                    <div class="form-group row">
+                                        <div class="offset-lg-3 offset-md-4 col-md-8 col-lg-7">
+                                            <a href="#" role="button" class="btn btn-primary btn-lg o_website_form_send">Request</a>
+                                            <span id="o_website_form_result"></span>
+                                        </div>
+                                    </div>
+                                </form>
+                            </div>
+                            <div class="oe_structure mt-2"/>
```
pedrobaeza: This doesn't serve AFAIK as you need an id for being saved as independent view

website_rma/views/request_rma_form.xml
```diff
@@ -0,0 +1,150 @@
...
+                        </div>
+                        <div class="col-lg-4">
+                            <t t-call="website.company_description"/>
+                        </div>
+                    </div>
+                </div>
+                <div class="oe_structure mt-2"/>
```
pedrobaeza: This doesn't serve AFAIK as you need an id for being saved as independent view

pedrobaeza (commented):
Is there more places to put placeholders?

website_rma/views/website_templates.xml
ernestotejeda: @jesusVMayor , @chienandalu it's done

rafaelbn (approved):
Functional test :+1:

pedrobaeza:
@chienandalu can you please confirm if the drag&drop blocks are correct now for not touching base view?

chienandalu:
> @chienandalu can you please confirm if the drag&drop blocks are correct now for not touching base view?

Looks like the ones with the `id` are ok, but there's none in the top so the dragged blocks go to the main view.

image

pedrobaeza:
@ernestotejeda please put `id` to all the blocks as requested multiple times, please.

chienandalu (approved):
Now it seems ok :+1: 
image

pedrobaeza (approved):

PR #167, opened 2020-09-15
[12.0][IMP] rma: incoming picking type use existing lot by default.

ernestotejeda:
cc @Tecnativa TT25510

### DIFF
--- a/rma/models/stock_warehouse.py
+++ b/rma/models/stock_warehouse.py
@@ -72,8 +72,8 @@ class StockWarehouse(models.Model):
             'rma_in_type_id': {
                 'name': _('RMA Receipts'),
                 'code': 'incoming',
-                'use_create_lots': True,
-                'use_existing_lots': False,
+                'use_create_lots': False,
+                'use_existing_lots': True,
                 'default_location_src_id': False,
                 'default_location_dest_id': self.rma_loc_id.id,
                 'sequence': max_sequence + 1,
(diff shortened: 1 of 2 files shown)


rma/models/stock_warehouse.py
```diff
@@ -73,7 +73,7 @@ def _get_picking_type_create_values(self, max_sequence):
                 'name': _('RMA Receipts'),
                 'code': 'incoming',
                 'use_create_lots': True,
```
pedrobaeza: I think this should be false, as you don't create lots for rma

joao-p-marques (approved):
Functional review ok :+1:

rma/models/stock_warehouse.py
chienandalu: I'd say it's mostly disabled by the `use_existing_lots` setting being `True`, but maybe it's better to leave it to `False` indeed.

pedrobaeza: They are not excluding in v12 AFAIK.

chienandalu: It expects `use_existing_lots` to be `False` here, for example: https://github.com/odoo/odoo/blob/12.0/addons/stock/models/stock_move.py#L490

Anyway, it's not very intuitive I must say. So I agree to set it to `False` :)

chienandalu (approved):

ernestotejeda:
Changes done and Conflicts resolved.
ping @pedrobaeza

PR #170, opened 2020-10-08
[12.0][IMP]rma: Add action to open invoices after refund.

jesusVMayor:
Just returns the action to view the invoices after create them


pedrobaeza:
Please check Travis

jesusVMayor:
travis green

ernestotejeda (commented):
Code and Functional review: OK.
But I think, to be consistent, if we bring this behavior for refunds, we should also bring it for the pickings that are generated when we do a replacement or a return. Would you please do that?.

angelmoya (approved):

PR #171, opened 2020-10-14
[FIX] rma: return permissions

chienandalu:
When a user with no stock manager permissions was doing the return, an
access error would raise.

cc @Tecnativa TT26165


pedrobaeza:
Isn't this changing the create user?

chienandalu:
But only for the moves. The picking remains with the user authorship

pedrobaeza (approved):

victoralmau (approved):

marcelsavegnago (approved):

PR #173, opened 2020-10-15
[FIX] rma: non rma users confirm

chienandalu:
The sudo had to be before calling the records

cc @Tecnativa TT26257


pedrobaeza (approved):
Yes, obvious patch!

marcelsavegnago (approved):
LGTM

victoralmau (approved):

PR #176, opened 2020-10-21
[FIX] rma: internal users portal

chienandalu:
Internal users with no RMA permissions got an access error when entering
the portal

cc @Tecnativa TT26338

### DIFF
--- a/rma/controllers/main.py
+++ b/rma/controllers/main.py
@@ -13,7 +13,11 @@ class PortalRma(CustomerPortal):
 
     def _prepare_portal_layout_values(self):
         values = super()._prepare_portal_layout_values()
-        values['rma_count'] = request.env['rma'].search_count([])
+        if request.env['rma'].check_access_rights(
+                'read', raise_exception=False):
+            values['rma_count'] = request.env['rma'].search_count([])
+        else:
+            values['rma_count'] = 0
         return values
 
     def _rma_get_page_view_values(self, rma, access_token, **kwargs):
--- a/rma/views/rma_portal_templates.xml
+++ b/rma/views/rma_portal_templates.xml
@@ -16,7 +16,7 @@
 
     <template id="portal_my_home_rma" name="Portal My Home : RMA entries" inherit_id="portal.portal_my_home" priority="30">
         <xpath expr="//div[hasclass('o_portal_docs')]" position="inside">
-            <t t-call="portal.portal_docs_entry">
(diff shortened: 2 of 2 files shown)


pedrobaeza:
Shouldn't rma_count be 0 in that case?

chienandalu:
Sure, that could be better. I was also thinking in another aproach with a try/expect so we don't depend on specific groups, what do you think?

pedrobaeza:
How is this handled in core?

victoralmau (commented):
IMO if is possible to add this groups in template https://github.com/OCA/rma/blob/12.0/rma/views/rma_portal_templates.xml#L17 is a better solution. What do you think about it?

And maybe too, can be interesting add some redirections in https://github.com/OCA/rma/blob/12.0/rma/controllers/main.py#L34 if user haven't some groups.

chienandalu:
> How is this handled in core?

They don't handle it at all, wich provokes access errors if the internal user doesn't have the proper permissions.

pedrobaeza:
@chienandalu I don't think that happen: if a user don't have sales permissions for example, is there an error on /my?

chienandalu:
An example with purchases:
```
odoo.exceptions.AccessError: ('Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:\n- Inventory/User\n\t- Accounting & Finance/Billing\n\t- Purchases/User\n\t- Purchases/Manager\n\t- User types/Portal\n\n(Document model: purchase.order) - (Operation: read, User: 6)', None)
```
image

pedrobaeza:
Ouch, that hurts... I think we can report it to Odoo. Anyway, for this, take the approach you consider more suitable.

chienandalu:
Changes done

pedrobaeza (approved):

pedrobaeza:
Oh, another doubt: what happen if you go with one of these users to /my/rma or click the link of 0 RMAs?

chienandalu:
Sure. They'll get a forbidden. The solution here should be to not show it if there aren't RMAs

pedrobaeza:
Yes, that's why I talked about putting a group in the template. I'm pretty sure that's possible even in QWeb templates.

chienandalu:
There goes the fix

pedrobaeza:
Any way, let's go with this fair solution:

/ocabot merge minor

chienandalu:
In this case, is the way Odoo does it in core

PR #179, opened 2020-10-31
[12.0][FIX] rma: RMA quantities are not quantities on hand

ernestotejeda:
cc @Tecnativa TT26542

### DIFF
--- a/rma/hooks.py
+++ b/rma/hooks.py
@@ -22,11 +22,10 @@ def post_init_hook(cr, registry):
 
     def create_rma_locations(warehouse):
         stock_location = env['stock.location']
-        location_vals = warehouse._get_locations_values({})
-        for field_name, values in location_vals.items():
-            if field_name == 'rma_loc_id' and not warehouse.rma_loc_id:
-                warehouse.rma_loc_id = stock_location.with_context(
-                    active_test=False).create(values).id
+        if not warehouse.rma_loc_id:
+            rma_location_vals = warehouse._get_rma_location_values()
+            warehouse.rma_loc_id = stock_location.with_context(
+                active_test=False).create(rma_location_vals).id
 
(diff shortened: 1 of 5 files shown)


rma/hooks.py
```diff
@@ -22,11 +22,10 @@ def _get_next_picking_type_color():
...
         stock_location = env['stock.location']
-        location_vals = warehouse._get_locations_values({})
-        for field_name, values in location_vals.items():
-            if field_name == 'rma_loc_id' and not warehouse.rma_loc_id:
-                warehouse.rma_loc_id = stock_location.with_context(
-                    active_test=False).create(values).id
+        rma_location_vals = warehouse._get_rma_location_values()
+        if not warehouse.rma_loc_id:
+            warehouse.rma_loc_id = stock_location.with_context(
+                active_test=False).create(rma_location_vals).id
```
pedrobaeza, line 28: This is not correct code. You have to extract the proper value.

rma/hooks.py
```diff
@@ -22,11 +22,10 @@ def _get_next_picking_type_color():
 
     def create_rma_locations(warehouse):
         stock_location = env['stock.location']
-        location_vals = warehouse._get_locations_values({})
-        for field_name, values in location_vals.items():
-            if field_name == 'rma_loc_id' and not warehouse.rma_loc_id:
-                warehouse.rma_loc_id = stock_location.with_context(
-                    active_test=False).create(values).id
+        rma_location_vals = warehouse._get_rma_location_values()
```
pedrobaeza: Do this only in case you need it (inside the if).

pedrobaeza (changes requested):

pedrobaeza (approved):

chienandalu (approved):

PR #183, opened 2020-11-16
[ADD] rma_sale_mrp: allow returning phantom products

chienandalu:
When a sale line has a phantom product (mrp kits) the RMA would not be
possible as the wizard couldn't pair the components moves with the
product in the line. With this approach, we can at least return the
spare components of the original kit line.

cc @Tecnativa TT26753


marcelsavegnago (approved):
LGTM

ernestotejeda (approved):

marcelsavegnago:
@chienandalu I believe you saw that you have a travis error

rma_sale/wizard/sale_order_rma_wizard.py
```diff
@@ -143,7 +146,7 @@ def _compute_move_id(self):
         for record in self:
             if record.picking_id:
                 record.move_id = record.picking_id.move_lines.filtered(
-                    lambda r: (r.sale_line_id.product_id == record.product_id
+                    lambda r: (r.sale_line_id == record.sale_line_id and r.product_id == record.product_id
```
pedrobaeza: This is not correct if there's no linked `sale_line_id` at all.

chienandalu: Is there an example in wich that could happen?

pedrobaeza: Returning a return will keep the sale_line_id?

chienandalu: But this wizard comes from de `sale.order`, how could we be returning from a return?

pedrobaeza: We do an RMA, return back to the customer, and a second RMA is needed inside the warranty period.

chienandalu: Ok, I understand now. I'll check

chienandalu: Ok, just tested it and indeed this approach fails at that point :/

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "restructuredtext.confPath": ""
+}
```
victoralmau:
```suggestion

```

rma/views/rma_views.xml
```diff
@@ -66,6 +66,7 @@
                 <field name="user_id"/>
                 <field name="partner_id"/>
                 <field name="product_id"/>
+
```
victoralmau:
```suggestion
```

rma_sale_mrp/tests/test_rma_sale.py
```diff
@@ -0,0 +1,129 @@
...
+    # def test_create_rma_with_so(self):
+    #     rma_form = Form(self.env['rma'])
+    #     rma_form.partner_id = self.partner
+    #     rma_form.order_id = self.sale_order
+    #     rma_form.product_id = self.product_kit
+    #     rma_form.product_uom_qty = 5
+    #     rma = rma_form.save()
+    #     rma.action_confirm()
+    #     self.assertTrue(rma.reception_move_id)
+    #     self.assertFalse(rma.reception_move_id.origin_returned_move_id)
```
victoralmau: You would need to uncomment this or remove it, don't you think?

victoralmau (changes requested):

chienandalu:
@ernestotejeda @pedrobaeza Can you review, please?

chienandalu:
and @victoralmau :sweat_smile:

ernestotejeda (commented):
Code review :+1:

victoralmau (approved):

victoralmau:
Travis is red in test

chienandalu:
Yes @victoralmau the test wasn't deterministic enough. I also put some order in the changes to separate the base ones from the new module ones

chienandalu:
This has been in production in customer for a while now. I think we can merge

PR #184, opened 2020-11-26
[FIX] rma: disable tracking when preparing refund

chienandalu:
We could come across access errors with `mail.tracking.value`

cc @Tecnativa TT26971

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -584,6 +584,7 @@ class Rma(models.Model):
             invoice_form = Form(self.env['account.invoice'].with_context(
                 default_type='out_refund',
                 company_id=rmas[0].company_id.id,
+                tracking_disable=True,
             ), "account.invoice_form")
             rmas[0]._prepare_refund(invoice_form, origin)
             refund = invoice_form.save()


chienandalu:
Looks like a corner case working with `Form()` as way to prepare records.

pedrobaeza (approved):

pedrobaeza:
Is this affecting the log entries we are going to find (for example, the entry for "Record created")?

chienandalu:
I tested it in the custumer demo instance in that message was correctly created.

image

pedrobaeza:
Well, that's a manual one, but not the automatic that are created by the system, with for example the affected tracking fields. You can check the difference creating a manual one. Now we have to decide if it's OK this way.

chienandalu:
Oh, yeah. That one is lost indeed.

pedrobaeza:
I'm afraid that includes possible notifications to followers, so I'm not very happy with this solution. Don't we have clear steps to reproduce the problem?

chienandalu:
Yes, you can reproduce it in a runbot (http://3452578-12-0-574333.runbot2-3.odoo-community.org/web?debug#id=1&action=345&active_id=1&model=rma&view_type=form&menu_id=182)

- Give a user RMA permission but no settings permissions
- Make an RMA.
- Click refund: error
```
Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:
- Administration/Settings

(Document model: mail.tracking.value) - (Operation: read, User: 6)
```

pedrobaeza:
OK, that's something we can work on. The weird thing is that `Form` has been used in other places without this problem (for example, in `contract` module in v13). I will check to see if we find the specific problem.

marcelsavegnago (approved):

rma/models/rma.py
```diff
@@ -584,6 +584,7 @@ def action_refund(self):
             invoice_form = Form(self.env['account.invoice'].with_context(
                 default_type='out_refund',
                 company_id=rmas[0].company_id.id,
+                tracking_disable=True,
```
victoralmau, line 587:
```suggestion
            invoice_form = Form(self.env['account.invoice'].sudo().with_context(
                default_type='out_refund',
                company_id=rmas[0].company_id.id,
```

victoralmau: IMO it's better solution.

victoralmau: Also, if I try to reproduce the error in the following way:
- Install `rma`
- Configure demo user with: RMA permissions, Account permissions and not Settings permission
- Create RMA
- Refund RMA

It doesn't show me the error.
Do I need any other addon installed and / or configured?

pedrobaeza: mail_tracking is not needed. The table is in core. sudo is not valid, as it changes other values like company.

victoralmau: Ok, `mail_tracking` is not required but i can't reproduce.

chienandalu: Are you testing with demo user?

victoralmau: Yes, demo user in local

chienandalu:
@victoralmau The key to reproduce it is to have the group.no_one set in the user (debug mode). But, probably this will be a better solution: https://github.com/odoo/odoo/pull/67621

chienandalu:
We rather close this

PR #185, opened 2020-11-30
[12.0][IMP] rma: Add customer as suggested recipient

ernestotejeda:
cc @Tecnativa TT27024


pedrobaeza (commented):
What happens if the customer is already a follower? Is it suggested again?

ernestotejeda:
No, in that case the customer is not suggested.

pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
FW-port it to v13

PR #186, opened 2020-11-30
[12.0][IMP] website_rma: Add customer as follower

ernestotejeda:
cc @Tecnativa TT27024

When creating an RMA from the website form, add the customer to the followers, the same as when creating an RMA from a sales order in the portal.


pedrobaeza (approved):

chienandalu (approved):

pedrobaeza:
/ocabot merge patch

Should this be fw-port it to v13?

PR #191, opened 2020-12-10
[12.0][FIX] rma_sale: Request RMA from SO without picking from portal

ernestotejeda:
cc @Tecnativa TT27238

### DIFF
--- a/rma_sale/views/sale_portal_template.xml
+++ b/rma_sale/views/sale_portal_template.xml
@@ -92,10 +92,12 @@
                                                     </div>
                                                 </td>
                                                 <td class="text-left">
-                                                    <span t-if="data['picking']" t-esc="data['picking'].name"/>
-                                                    <input type="hidden"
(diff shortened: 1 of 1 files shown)


rma_sale/views/sale_portal_template.xml
```diff
@@ -92,10 +92,12 @@
                                                     </div>
                                                 </td>
                                                 <td class="text-left">
-                                                    <span t-if="data['picking']" t-esc="data['picking'].name"/>
-                                                    <input type="hidden"
-                                                           t-attf-name="#{data_index}-picking_id"
-                                                           t-att-value="data['picking'].id"/>
+                                                    <t t-if="data['picking']">
+                                                        <span t-if="data['picking']" t-esc="data['picking'].name"/>
```
pedrobaeza: This t-if can be removed now

pedrobaeza:
I suppose the result is now that now RMA can be requested, isn't it?

chienandalu (approved):

PR #192, opened 2020-12-15
[12.0][IMP] *rma*: Avoid RMAs of kits until rma_mrp is installed

ernestotejeda:
cc @Tecnativa TT27238

- Avoid creating RMAs from a portal Sales Order page for lines that don't have an associated picking (like Kit lines).
- Avoid creating RMAs from the website RMA for Kit products.

In PR https://github.com/OCA/rma/pull/183/files we are working on handling RMAs for kit products, but meanwhile, it is necessary to avoid creating RMAs from the website and portal for those kit products to avoid errors.

### DIFF
--- a/website_rma/controllers/main.py
+++ b/website_rma/controllers/main.py
@@ -26,10 +26,19 @@ class WebsiteRMA(http.Controller):
         """Domain used for the products to be shown in selection of
         the web form.
         """
-        return [
-            ('name', '=ilike', "%{}%".format(q or '')),
+        domain = [
+            ("name", "=ilike", "%{}%".format(q or "")),
             ("sale_ok", "=", True),
         ]
+        # HACK: As there is no glue module for this purpose we have put
+        # this this condition to check that the mrp module is installed.
+        if "bom_ids" in request.env["product.product"]._fields:
+            domain += [
+                "|",
+                ("bom_ids.type", "!=", "phantom"),
+                ("bom_ids", "=", False),
+            ]
+        return domain
 
     @http.route(['/requestrma'], type='http', auth="user", website=True)
     def request_rma(self, **kw):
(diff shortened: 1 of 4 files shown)


website_rma/controllers/main.py
```diff
@@ -29,6 +30,9 @@ def _get_website_rma_product_domain(self, q):
         return [
             ('name', '=ilike', "%{}%".format(q or '')),
             ("sale_ok", "=", True),
+            '|',
+            ('bom_ids.type', '!=', 'phantom'),
```
pedrobaeza: This is not valid, as you don't depend on `mrp` for doing this. You need to find another way.

ernestotejeda:
@pedrobaeza changes done.
ping @chienandalu

chienandalu (commented):
Can you include this commit?: https://github.com/Tecnativa/rma/commit/244e9874ff337222467c92a34a792977703d67d1

chienandalu (approved):

chienandalu (approved):

website_rma/controllers/main.py
```diff
@@ -26,10 +27,17 @@ def _get_website_rma_product_domain(self, q):
...
-        return [
-            ('name', '=ilike', "%{}%".format(q or '')),
+        domain = [
+            ("name", "=ilike", "%{}%".format(q or "")),
             ("sale_ok", "=", True),
         ]
+        if "bom_ids" in request.env["product.product"]._fields:
+            domain += [
+                "|",
+                ("bom_ids.type", "!=", "phantom"),
```
pedrobaeza, line 39: This is still using things from `mrp`, so this can't be

pedrobaeza (changes requested):

victoralmau (approved):

website_rma/controllers/main.py
pedrobaeza: Not the ideal option, but finally there's no best one, so let's go with it.

pedrobaeza (approved):

pedrobaeza:
Please fw-port it to v13

PR #194, opened 2020-12-24
[12.0][FIX] rma_sale: choose the correct move to be compatible with mrp kits components.

ernestotejeda:
Cc @Tecnativa TT27238


pedrobaeza:
Duplicated.

PR #203, opened 2021-02-09
[IMP] rma: auto send confirmation email

chienandalu:
Now we can set if we want to send the mail notification automatically to the RMA customer

cc @Tecnativa TT26371

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -528,10 +528,23 @@ class Rma(models.Model):
                 _("You cannot delete RMAs that are not in draft state"))
         return super().unlink()
 
+    def _send_confirmation_email(self):
+        """Auto send notifications"""
+        for rma in self.filtered(lambda p: p.company_id.send_rma_confirmation):
+            rma_template_id = (
+                rma.company_id.rma_mail_confirmation_template_id.id
+            )
+            rma.with_context(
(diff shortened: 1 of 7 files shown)


rma/models/rma.py
```diff
@@ -528,10 +528,23 @@ def unlink(self):
...
 
+    def _send_confirmation_email(self):
+        """Auto send notifications"""
+        for rma in self.filtered(lambda p: p.company_id.send_rma_confirmation):
+            rma_template_id = (
+                rma.company_id.rma_mail_confirmation_template_id.id
+            )
+            rma.with_context(
+                force_send=True, 
+                mark_rma_as_sent=True
```
pedrobaeza, line 539: Why this context?

pedrobaeza (commented):
Check Travis

rma/models/rma.py
chienandalu: It changes the state of the RMA `sent` field when it passes by `message_post`: https://github.com/OCA/rma/blob/12.0/rma/models/rma.py#L1144

chienandalu:
Fixed

chienandalu:
This is ready

marcelsavegnago (approved):
LGTM

PR #205, opened 2021-03-01
[IMP] rma: automatic notification subtypes

chienandalu:
Allow users to unsubscribe from automatic notifications

cc @Tecnativa TT28344

### DIFF
--- a/rma/data/mail_data.xml
+++ b/rma/data/mail_data.xml
@@ -7,6 +7,12 @@
         <field name="default" eval="False"/>
         <field name="description">RMA in draft state</field>
     </record>
+    <record id="mt_rma_notification" model="mail.message.subtype">
+        <field name="name">RMA Notificatoin</field>
+        <field name="res_model">rma</field>
+        <field name="default" eval="False"/>
+        <field name="description">RMA automatic customer notifications</field>
+    </record>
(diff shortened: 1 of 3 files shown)


pedrobaeza:
What about migration scripts? With this, messages will be internal, which doesn't seem correct for customer messages.

chienandalu:
> What about migration scripts? 

Should we subscribe users by default or maybe take the default from the other notification type?

> With this, messages will be internal, which doesn't seem correct for customer messages.

Well, is not really internal but, in Odoo's terms, "not discussion". It's a pity that it's so hard linked to a specific subtype: https://github.com/odoo/odoo/blob/12.0/addons/mail/static/src/xml/thread.xml#L279
https://github.com/odoo/odoo/blob/12.0/addons/mail/models/mail_message.py#L525

pedrobaeza:
But the customer receives the message as "Internal message: replies will only be seen by employees..."?

chienandalu:
No:

image

pedrobaeza:
Are migration scripts needed?

chienandalu:
I don't think so

pedrobaeza (approved):

marcelsavegnago (approved):

PR #212, opened 2021-03-31
[FIX] rma_sale: fix wrong values for RMA creation

SilvioGregorini:
Lesser fix for creating RMA from website.
Current values will break any custom behaviour because Many2one fields are set using strings instead of integer (IDs).
This will result, e.g., in having stuff like:
```
>>> self
rma(1,)
>>> self.product_id
product.product('1',)
>>> self.product_id.name
*** odoo.exceptions.AccessError: Database fetch misses ids ('1') and has extra ids (1), may be caused by a type incoherence in a previous request
```

### DIFF
--- a/rma_sale/controllers/sale_portal.py
+++ b/rma_sale/controllers/sale_portal.py
@@ -25,10 +25,15 @@ class CustomerPortal(CustomerPortal):
             return request.redirect("/my")
         order_obj = request.env["sale.order"]
         wizard_obj = request.env["sale.order.rma.wizard"]
+        wizard_line_field_types = {
+            f: d["type"] for f, d in wizard_obj.line_ids.fields_get().items()
+        }
         # Set wizard line vals
         mapped_vals = {}
         for name, value in post.items():
             row, field_name = name.split("-", 1)
+            if wizard_line_field_types.get(field_name) == "many2one":
+                value = int(value) if value else False
             mapped_vals.setdefault(row, {}).update({field_name: value})
         # If no operation is filled, no RMA will be created
         line_vals = [
--- a/rma_sale/wizard/sale_order_rma_wizard.py
+++ b/rma_sale/wizard/sale_order_rma_wizard.py
@@ -130,6 +130,8 @@ class SaleOrderLineRmaWizard(models.TransientModel):
                         and r.sale_line_id.order_id == record.order_id
                     )
                 )
+            else:
+                record.move_id = False
 
     @api.depends("order_id")
     def _compute_allowed_product_ids(self):


rma_sale/controllers/sale_portal.py
```diff
@@ -29,6 +29,8 @@ def request_rma(self, order_id, access_token=None, **post):
         mapped_vals = {}
         for name, value in post.items():
             row, field_name = name.split("-", 1)
+            if field_name.endswith('_id'):
```
pedrobaeza: This is a very weak condition. It's your duty to check in your customizations if it's a string. All portal works this way. But it's as simple as always do `int(whatever)`, being already an integer or not.

pedrobaeza:
@chienandalu @ernestotejeda can you confirm my statement?

rma_sale/controllers/sale_portal.py
SilvioGregorini: I know condition is weak, and could be improved by using a better check [1]. I can do it, if necessary. However, the problem is that every customization will receive faulty data anyway, because the values herein set will be directly passed to a `create` method. The only way a customization could check if the values are correct, is to override the `create` itself, but it would be highly expensive in terms of performance, I believe.

[1] i.e.: `request.env['sale.order.line.rma.wizard']._fields[field_name].type == 'many2one'`

pedrobaeza: How the rest of the portal works with this? I don't think it's expensive to envelop the value with `int()`

SilvioGregorini: > How the rest of the portal works with this?

I'm sorry, I think I don't understand what you mean.

> I don't think it's expensive to envelop the value with int()

But overriding every model's `create` to check if every value is correct is more expensive than making this check beforehand :)

ernestotejeda:
@SilvioGregorini , I have done some tests locally and I can't reproduce the problem. Each RMA that is created from the portal is created correctly and I am able to acces to its product via shell o via web browser without any problem.
In what specific scenario does that happen? What steps can I follow to reproduce the error?

chienandalu:
> @SilvioGregorini , I have done some tests locally and I can't reproduce the problem. Each RMA that is created from the portal is created correctly and I am able to acces to its product via shell o via web browser without any problem.
> In what specific scenario does that happen? What steps can I follow to reproduce the error?

Well, I could catch it just after the rma is created from the portal. In v12, the orm would swallow an string id and parsed it in the moment, but it seems that in v13 this is done after the method is finished somehow. The final record is ok, but if you intercept the method you run into problems like the ones @SilvioGregorini is experiencing.

@SilvioGregorini  Out of curiosity, how the code of your customization, where do you intervine?

pedrobaeza:
@chienandalu how other controllers work with this?

chienandalu:
> @chienandalu how other controllers work with this?

I should take a look on how `website_form` deals with it

SilvioGregorini:
> @SilvioGregorini Out of curiosity, how the code of your customization, where do you intervine?

@chienandalu it first happened in a simple `create` override, then it happened again in a `compute` method. In both cases, I was accessing the `rma` record's `product_id` field, and then I tried to read some field from the `product.product` record I retrieved; unfortunately, it triggered the error I wrote above in the PR's description.

chienandalu:
> I should take a look on how `website_form` deals with it

After a little research, this is how Odoo deals with data coming from request parameters: https://github.com/odoo/odoo/blob/13.0/addons/website_form/controllers/main.py#L169

To what concern us now for this issue, a simple conversion to `int` is made for the many2one fields: https://github.com/odoo/odoo/blame/13.0/addons/website_form/controllers/main.py#L80

So my proposal would be having a similar method that just covers the known field types coming from this module's form but that can be extended if further cleanup is needed in other customizations.

pedrobaeza:
As we are already depending on `website_form`, I think we can do (more or less):
```python
from odoo.addons.website_form.controllers.main import _input_filters
...

    def request_rma(self, order_id, access_token=None, **post):
        ...
        for name, value in post.items():
            row, field_name = name.split("-", 1)
            Line = request.env["sale.order.rma.wizard"]
            value = _input_filters[Line._fields[field_name].ttype](Line, field_name, value)
```

Anyway, I also wonder if `website_form` provides something to be inherited for avoiding this post-processing here. Can't we use this generic controller https://github.com/odoo/odoo/blob/ce590d25f7823b1a0193ca4241c0c564bad4b8fc/addons/website_form ?/controllers/main.py#L28

chienandalu:
> As we are already depending on `website_form`, I think we can do (more or less):

The problem is that `rma_sale` doesn't depend on `website_form`. You're thinking in `website_rma` I think :slightly_smiling_face:

pedrobaeza:
Ah, right, I have mixed both. Then a basic check on type and doing `int` can be enough.

SilvioGregorini:
@pedrobaeza @chienandalu @ernestotejeda 

I have updated this PR with a better field check and value management.
Please feel free to comment and share proposal for further improvements.

However, I also discovered that there was a little bug: if the `sale.order.line.rma.wizard` record had no picking, the compute method `_compute_move_id` would have not assigned any value to field `move_id`, and once Odoo tried to access such field within the method `_prepare_rma_values`, a `CacheError` would happen.
[Here's the traceback.](http://paste.debian.net/1193666/)
This bug is easily solved by simply explicitly assigning `False` to `move_id` within the computed method (as you can see in the second commit, 1762317176b1cc4a47facd3fe6340acbc638ceb9).

chienandalu (approved):
Great! :+1:

PR #214, opened 2021-04-15
[FIX] rma_sale_mrp: infinite loop

chienandalu:
In the wizard, if a component line didn't have any quantity, we could
end in an infinite loop when creating the RMA.

cc @Tecnativa TT29223

please check it @pedrobaeza @Yajo

### DIFF
--- a/rma_sale_mrp/wizard/sale_order_rma_wizard.py
+++ b/rma_sale_mrp/wizard/sale_order_rma_wizard.py
@@ -33,33 +33,39 @@ class SaleOrderRmaWizard(models.TransientModel):
         return super().create(vals)
 
     def create_rma(self, from_portal=None):
-        """We'll decompose the RMAs and remove the phantom lines"""
+        """We kept the component lines in the shade and now we need to take
+        the ones linked to the phatom lines (kits) to create the proper RMA
+        for each component according to the proper component quantities"""
         phantom_lines = self.line_ids.filtered("phantom_kit_line")
-        # Coming from the portal, we'll compute how many per kit to receive.
-        # From backend we'll be returning them from each individual component
-        # import pdb; pdb.set_trace()
+        kit_line_vals = []
         for line in phantom_lines:
-            kit_lines = self.component_line_ids.filtered(
+            # There could be two lines for the same product, so we need to
+            # split them to get the right quantities and operations for each
+            # one and then group them by product to process them altogether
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

rma_sale_mrp/wizard/sale_order_rma_wizard.py
```diff
@@ -33,33 +33,39 @@ def create(self, vals):
...
+        for each component according to the proper component quantities"""
         phantom_lines = self.line_ids.filtered("phantom_kit_line")
-        # Coming from the portal, we'll compute how many per kit to receive.
-        # From backend we'll be returning them from each individual component
-        # import pdb; pdb.set_trace()
+        kit_line_vals = []
         for line in phantom_lines:
-            kit_lines = self.component_line_ids.filtered(
+            # There could be two lines for the same product, so we need to
+            # split the to get the right quantities and operations for each
```
yajo:
```suggestion
            # split them to get the right quantities and operations for each
```

rma_sale_mrp/wizard/sale_order_rma_wizard.py
```diff
@@ -33,33 +33,39 @@ def create(self, vals):
...
-                qty_to_return = product_kit_lines[0].per_kit_quantity * line.quantity
+            for product in kit_component_lines.mapped("product_id"):
+                product_kit_component_lines = kit_component_lines.filtered(
+                    lambda x: x.product_id == product and x.quantity
+                )
+                qty_to_return = (
+                    product_kit_component_lines[0].per_kit_quantity
+                    * line.quantity
+                )
                 while qty_to_return:
```
yajo: I'm not sure if there's any chance for `qty_to_return` to become a negative number, but if there were, this would save you from an infinite loop.

```suggestion
                while qty_to_return > 0:
```

yajo (approved):
A couple of suggestions. Notice that I don't know this module... I may be saying something wrong. 😅

rma_sale_mrp/wizard/sale_order_rma_wizard.py
chienandalu: Yes it's a good safeguard :+1:

rma_sale_mrp/wizard/sale_order_rma_wizard.py
chienandalu: Done

PR #217, opened 2021-04-26
[FIX] website_rma: controller implementation

chienandalu:
It was assumed that this method (used for any model) always had an RMA
in return, which wasn't correct at all and could lead to cross-model and
cross-customer subscripting if an RMA id matched the one of the form
model creating (eg.: `crm.lead` in 'Contact Us' form)

cc @Tecnativa TT29397


pedrobaeza (approved):

pedrobaeza:
Ouch, this is critical, so fast-tracking:

/ocabot merge patch

PR #221, opened 2021-05-06
[IMP] rma: tags

chienandalu:
Now with tags :slightly_smiling_face: 
image

TT29594

ping @pedrobaeza @victoralmau

### DIFF
--- a/rma/__manifest__.py
+++ b/rma/__manifest__.py
@@ -32,6 +32,7 @@
         "views/rma_portal_templates.xml",
         "views/rma_team_views.xml",
         "views/rma_views.xml",
+        "views/rma_tag_views.xml",
         "views/stock_picking_views.xml",
         "views/stock_warehouse_views.xml",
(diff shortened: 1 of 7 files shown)


victoralmau (changes requested):
Good job.
Please, add color field in tags view.
Add some tests about it (or add some tags and use it in actual tests).

chienandalu:
> Please, add color field in tags view.

Normally this is done via the many2many_tags widget. Having it on the form I think is a bit misleading as the user doesn't know which index corresponds to what color

chienandalu:
> Add some tests about it (or add some tags and use it in actual tests).

As for this, there's no logic added, so the test would be kinda dummy, don't you think?

victoralmau (approved):

pedrobaeza (approved):

pedrobaeza:
Please translate it when merged and fw-port it to v13:

/ocabot merge minor

PR #220, opened 2021-05-04
[MIG] rma_sale_mrp: Migration to 13.0

chienandalu:
- Added two little fixes of stuff missing in v13 base modules
- Tests and methods adapted to v13 kits changes

cc @Tecnativa TT29358

ping @pedrobaeza @victoralmau

PR #223, opened 2021-05-11
[FIX] rma_sale_mrp: kit line with no uom qty

chienandalu:
If a user corrects the original sale order and lowers a kit line quantity
to 0, we don't want to compute anything on it.

cc @Tecnativa TT29669

@pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):

chienandalu:
Just added change to https://github.com/OCA/rma/pull/220

PR #225, opened 2021-05-20
[IMP] rma: notify reception to customer

chienandalu:
Now we can configure if an automatic notification should be sent when we
receive the goods from an RMA in our warehouse

If we've got `rma_sale` or `website_rma` we can also configure draft
notifications so when the customer places an RMA from the portal the
receive an acknowledge email.

cc @Tecnativa TT29595

ping @CarlosRoca13 @victoralmau

### DIFF
--- a/rma/data/mail_data.xml
+++ b/rma/data/mail_data.xml
@@ -8,7 +8,7 @@
         <field name="description">RMA in draft state</field>
     </record>
     <record id="mt_rma_notification" model="mail.message.subtype">
-        <field name="name">RMA Notificatoin</field>
+        <field name="name">RMA Notification</field>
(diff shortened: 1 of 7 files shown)


CarlosRoca13 (approved):

victoralmau (approved):

pedrobaeza (commented):
Is this sent RMA per RMA, no matter if they come in batch?

chienandalu:
> Is this sent RMA per RMA, no matter if they come in batch?

That's a good remark. Indeed they are. What grouping criteria should we use?

pedrobaeza:
If they are from the sale sale.order/stock.picking?

PR #226, opened 2021-05-25
[12.0][IMP] rma: public tags

chienandalu:
Now we can decide which tags are visible for the customer. This way, we
can use them as pseudo-states

cc @Tecnativa TT29594

@pedrobaeza @victoralmau


victoralmau (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

To v13

PR #228, opened 2021-05-27
[FIX] rma_sale: chained returns

chienandalu:
Using move_dest_ids we can easily end in an infinite loop situation as
the return of the return of the return ends with some original moves on
in move_dest_ids. We must ensure to drop them to avoid the infinite
loop.

cc @Tecnativa TT29886

ping @pedrobaeza @Yajo


pedrobaeza (approved):

pedrobaeza:
Let's fix this:

/ocabot merge patch

pedrobaeza:
Cherry-pick to v13 at 7c9b6bc3a8da71ee7d372f160d96625f66d60dc0

PR #230, opened 2021-06-22
[IMP] rma_sale: portal rma request single page view

chienandalu:
Now it's possible to configure if the portal RMA request form is loaded
in a popup or in a single page.

In that page, we can add custom blocks (if the website is installed) a
customize the form text.

In this commit, we also add the possibility to extend the form view to allow
custom fields that will show up in the RMA description.

cc @Tecnativa TT29670

### DIFF
--- a/rma_sale/__manifest__.py
+++ b/rma_sale/__manifest__.py
@@ -20,6 +20,7 @@
         "views/rma_views.xml",
         "views/sale_views.xml",
         "views/sale_portal_template.xml",
+        "views/res_config_settings_views.xml",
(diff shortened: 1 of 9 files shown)


pedrobaeza:
Do you think it should be a configurable thing, or directly go to the full page schema?

chienandalu:
Well, I didn't want to force such  a breaking behavior in the middle of the version. Let's see how it's received and maybe in the forward port we can set it full page for everyone

pedrobaeza (approved):

victoralmau (approved):

PR #238, opened 2021-09-23
[13.0][FIX] rma: return location

chienandalu:
Added changes:

- The new implementation wasn't getting the value set in the wizard.
- Having a default responsible value as it was defined doesn't behave right when the user it's a portal one or the petition comes from a sudo().
- Also, by design it was expected to have a default team if it wasn't defined, but as it was implemented that wasn't possible.

cc @Tecnativa TT32046

ping @ernestotejeda @pedrobaeza

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -50,7 +50,6 @@ class Rma(models.Model):
         comodel_name="res.users",
         string="Responsible",
         index=True,
-        default=lambda self: self.env.user,
         tracking=True,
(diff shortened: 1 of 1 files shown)


victoralmau (approved):

chienandalu:
I just added an extra fix

victoralmau (approved):

pedrobaeza (approved):

pedrobaeza:
Please move it to 14.0

PR #241, opened 2021-10-22
[FIX] fix filter rma location based on the sale order company

sebastienbeau:
In case of multicompany (working on several company at the same time, or having the right to see warehouse but not location of other company ...), we should only propose the location based on the company of the sale order.


chienandalu (approved):

ernestotejeda (approved):

sebastienbeau:
Thanks for the quick processing !

PR #248, opened 2021-11-30
[14.0][perf] Change allowed_product_ids domain to avoid performance issues

florian-dacosta:
When testing rma_sale with a database having a big number of product (around 100k) and I had big performance issues.
Clicking on the "create" button to create the empty RMA form on the UI was a few minutes.

After some investigation, I believe it is an issue to have a computed many2many field with so many records.
I made some change in order to avoid this perf issue but keeping the feature, returning the domain in a onchange instead.
So if the rma is not link to any sale order, the product_id field will have a static domain rather than whole product database as ids.

It made me wonder about this domain on product field. I guess it may be nice to have in some case, but isn't it too restrictive ?
In case I sale product link to a BOM, phantom or manufacturable, I won't be able to make the RMA on a component ?
Any solution for this?

@chienandalu @ernestotejeda

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -34,7 +34,6 @@ class Rma(models.Model):
         comodel_name="product.product",
         compute="_compute_allowed_product_ids",
     )
-    product_id = fields.Many2one(domain="[('id', 'in', allowed_product_ids)]")
 
     @api.depends("partner_id", "order_id")
     def _compute_allowed_picking_ids(self):
@@ -67,11 +66,7 @@ class Rma(models.Model):
                     lambda r: r.type in ["consu", "product"]
                 ).ids
             else:
-                rec.allowed_product_ids = (
-                    self.env["product.product"]
-                    .search([("type", "in", ["consu", "product"])])
-                    .ids
-                )
+                rec.allowed_product_ids = []
 
     @api.onchange("partner_id")
     def _onchange_partner_id(self):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
The problem with the onchanges is that the are lost when coming again to the same form, so can you think about another solution? Maybe a fixed domain on type part + a variable one? Not sure also if the operator `?=` (`?in` exists?) can help.

florian-dacosta:
And a fixed domain like this : `[('type', 'in', ('consu', 'product'), '|', '&', ("order_id", '!=', False), ('id', 'in', allowed_product_ids), ('order_id', '=', False)]`
And allowed_product_ids would stay empty in case there are no linked sale order.
What do you think @pedrobaeza ?

pedrobaeza:
Is that avoiding your problem?

florian-dacosta:
No, sorry I went too fast, it is nonsense.
I don't see how to get around this without the onchange stuff. The `in?` operator does not seem to exist in Odoo, it would indeed do exactly what I need.

The onchange method is not perfect, but still, it would still do the job for 98% of the cases.
IMHO it is better to have this onchange feature and make the module usable for project with big product catalog than to have this small feature `allowed_products` feature (which seems to be a problem for me when using BOMs) working 100% but making the module unusable for some project because of performance.

Do you think it would be acceptable as it is ?
I'll still try to find a workaround, but I am out of ideas for now.

pedrobaeza:
I'm seeing that you can use a method for specifying the domain:

https://github.com/odoo/odoo/blob/50930f089b683f8cbd1a1dde79d2f4d1fc3f65bc/addons/hr_timesheet/models/hr_timesheet.py#L48

so that should be the solution.

florian-dacosta:
> I'm seeing that you can use a method for specifying the domain:
> 
> https://github.com/odoo/odoo/blob/50930f089b683f8cbd1a1dde79d2f4d1fc3f65bc/addons/hr_timesheet/models/hr_timesheet.py#L48
> 
> so that should be the solution.

Thanks for the tip Pedro, but I don't think it solves the issue.
The problem is that I want the domain to change depending on a field (order_id) of the RMA.
This domain method does not depend on the RMA. It may be usefull I guess to make a domain that depend on the current user or something, but not on a specific RMA. Unless I am missing something, it does not help in that case.

PR #249, opened 2021-12-01
[14.0][FIX] multi company error on rma location choice

florian-dacosta:
In a multi company environment, one the RMA form, we get an error when trying to fill the "location" field.

Explanation : 
The location_id field has a method `_domain_location_id` which search all rma locations from all warehouse (no domain).
Problem is, these domain method are called with sudo mode. So the domain will always be set with the RMA location from all database warehouse (from all database companies).
(easy to reproduce on demo database)

Then, when filling the field, we get a multi company access error as the user is actually not allowed to access all companies.
Workaround is to explicitly set company_id in the warehouse domain.

@pedrobaeza @chienandalu @ernestotejeda

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -18,7 +18,14 @@ class Rma(models.Model):
     _inherit = ["mail.thread", "portal.mixin", "mail.activity.mixin"]
 
     def _domain_location_id(self):
-        rma_loc = self.env["stock.warehouse"].search([]).mapped("rma_loc_id")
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

marcelsavegnago (approved):

chienandalu (approved):

pedrobaeza:
I suppose this should be backported to 13.0 as well, isn't it?

/ocabot merge patch

PR #251, opened 2022-01-17
[FIX] rma: return wizard to_refund

chienandalu:
To create the rma we create the core return wizard, which now sets
to_refund to True as default. This isn't right for RMAs anyway.

cc @Tecnativa TT33706

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -877,7 +877,14 @@ class Rma(models.Model):
             lambda r: r.move_id != self.move_id
         ).unlink()
         return_line = return_wizard.product_return_moves
-        return_line.quantity = self.product_uom_qty
+        return_line.update(
+            {
+                "quantity": self.product_uom_qty,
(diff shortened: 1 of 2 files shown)


rma/models/rma.py
```diff
@@ -878,6 +878,9 @@ def _create_receptions_from_picking(self):
         ).unlink()
         return_line = return_wizard.product_return_moves
         return_line.quantity = self.product_uom_qty
+        # The to_refund field is now True by default, which isn't right in the RMA
+        # creation context.
+        return_line.to_refund = False
```
pedrobaeza: Do both in the same write for performance optimization.

chienandalu: done

pedrobaeza (approved):

ernestotejeda (approved):

pedrobaeza:
/ocabot merge patch

please fw-port it to v14

PR #252, opened 2022-01-20
[IMP] rma_sale: avoid submit form in portal if no info

chienandalu:
- In the portal customer form the submit button won't be enabled until we can submit the info
- Some refactor to make it happen

Peek 21-01-2022 08-42

cc @Tecnativa TT33572

please review @Tardo @CarlosRoca13

### DIFF
--- a/rma_sale/static/src/js/rma_portal_form.js
+++ b/rma_sale/static/src/js/rma_portal_form.js
@@ -4,42 +4,121 @@
 odoo.define("rma_sale.animation", function(require) {
     "use strict";
 
-    var sAnimation = require("website.content.snippets.animation");
+    const publicWidget = require("web.public.widget");
 
-    // In the customer portal when a RMA operation is selected show the comments
-    // selector so the user doesn't miss the chance to add his comments
-    sAnimation.registry.rma_operation_portal = sAnimation.Class.extend({
-        selector: ".rma-operation",
-        start: function() {
-            this.id = this.el.name.replace("-operation_id", "");
-            this.$comment = $("#comment-" + this.id);
-            this.$comment_input = $("[name='" + this.id + "-description']");
-            var _this = this;
-            this.$el.on("change", function() {
-                _this._onChangeOperationId();
-            });
+    /**
+     * Adds some machinery to the customer portal RMA form:
+     *
+     * - Avoid submitting the form if no qty or operation is reported.
+     * - Show automatically the observations field when the operation is selected
+     *   and hide it back with no operation selected.
+     */
+    publicWidget.registry.PortalRmaSale = publicWidget.Widget.extend({
+        selector: "#form-request-rma",
+        events: {
+            "change .rma-operation": "_onChangeOperationId",
+            "change #delivery-rma-qty input": "_onChangeQty",
         },
-        _show_comment: function() {
-            if (this.$comment) {
-                this.$comment.removeClass("show");
-                this.$comment.addClass("show");
-                if (this.$comment_input) {
-                    this.$comment_input.focus();
+
+        /**
+         * @override
+         */
+        start: function() {
+            const ids = this.$("[name*='-operation_id']")
+                .map(function() {
+                    return this.name.replace("-operation_id", "");
+                })
+                .get();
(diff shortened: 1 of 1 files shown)


rma_sale/static/src/js/rma_portal_form.js
```diff
@@ -4,42 +4,122 @@
...
+        },
+
+        /**
+         * @override
+         */
         start: function() {
-            this.id = this.el.name.replace("-operation_id", "");
-            this.$comment = $("#comment-" + this.id);
-            this.$comment_input = $("[name='" + this.id + "-description']");
+            const ids = $("[name*='-operation_id']")
```
Tardo: plublicWidget extends web.Widget which defines the function "$" which calls "this.$el.find". This ensures that the widget can work together with other widget instances and, in general, not collide if other widgets identify an element in the same way.
```suggestion
            const ids = this.$("[name*='-operation_id']")
```

rma_sale/static/src/js/rma_portal_form.js
```diff
@@ -4,42 +4,122 @@
...
             var _this = this;
-            this.$el.on("change", function() {
-                _this._onChangeOperationId();
+            _.each(ids, function(id) {
+                _this.rows[id] = {};
+                _this.rows[id].$comment = $("#comment-" + id);
+                _this.rows[id].$comment_input = $("[name='" + id + "-description']");
+                _this.rows[id].$operation = $("[name='" + id + "-operation_id']");
+                _this.rows[id].$qty = $("[name='" + id + "-quantity']");
             });
```
Tardo: It might be interesting to use es6... You may notice that "_this" disappears since the function will use the same context as the one invoking it. Also, for better readability "template strings" is used.
```suggestion
            _.each(ids, (id) => {
                this.rows[id] = {
                    $comment: this.$(`#comment-${id}`),
                    $comment_input: this.$(`[name='${id}-description']`),
                    $operation: this.$(`[name='${id}-operation_id']`),
                    $qty: this.$(`[name='${id}-quantity']`)
                };
            });
```

rma_sale/static/src/js/rma_portal_form.js
```diff
@@ -4,42 +4,122 @@
...
-                if (this.$comment_input) {
-                    this.$comment_input.focus();
+        /**
+         * @private
+         * @param {Object} row: the form row structure
+         */
+        _show_comment: function(row) {
+            if (row.$comment) {
+                row.$comment.removeClass("show");
+                row.$comment.addClass("show");
```
Tardo: why?
```suggestion
                row.$comment.removeClass("show");
                row.$comment.addClass("show");
```

rma_sale/static/src/js/rma_portal_form.js
```diff
@@ -4,42 +4,122 @@
...
+            var _this = this;
+            var can_submit = false;
+            _.each(this.rows_ids, function(id) {
+                const row = _this.rows[id];
+                if (
+                    row &&
+                    // Qty greater than 0
+                    row.$qty &&
+                    row.$qty.val() &&
+                    parseInt(row.$qty.val()) &&
```
Tardo:
```suggestion
                    Number(row.$qty.val()) &&
```

rma_sale/static/src/js/rma_portal_form.js
```diff
@@ -4,42 +4,122 @@
...
+                    // An operation is defined
+                    row.$operation &&
+                    row.$operation.val()
+                ) {
+                    can_submit = true;
```
Tardo, line 87: It may be more cost effective to use a native js loop and use "break".

Tardo (changes requested):
I block it only because "this.$()" is not being used, the rest of the comments are optional.

chienandalu:
@Tardo Thanks for the juicy tips. All applied, can you review again?

Tardo (approved):
code review

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port it to v14

PR #255, opened 2022-01-28
[IMP] rma: finish RMA manually

chienandalu:
Sometimes there's no choice to return, refund or replace an RMA for
different reasons. For example, when the customer doesn't want to repair
the product because that would be too expensive. We still want finish
the RMA and document the reasons. This improvement allows it.

cc @Tecnativa TT34164


ernestotejeda (approved):
Great job!
code and functional review (tested on runboat) 👍🏼

rma/security/ir.model.access.csv
```diff
@@ -8,3 +8,6 @@ access_rma_operation_user_own,rma.operation.user.own,model_rma_operation,rma_gro
 access_rma_operation_manager,rma.operation.manager,model_rma_operation,rma_group_manager,1,1,1,1
 access_rma_tag_user_own,rma.tag.user.own,model_rma_tag,rma_group_user_own,1,0,0,0
 access_rma_tag_manager,rma.tag.manager,model_rma_tag,rma_group_manager,1,1,1,1
+access_rma_finalization_portal,rma.finalization.portal,model_rma_finalization,base.group_portal,1,0,0,0
+access_rma_finalization_user_own,rma.finalization.user.own,model_rma_finalization,rma_group_user_own,1,1,1,0
```
pedrobaeza: This should be only for read. The reasons should be only created by managers.

chienandalu: Done

pedrobaeza (approved):

pedrobaeza:
/ocabot merge minor

Please fw-port it to v14

PR #257, opened 2022-02-07
[14.0][FIX] Do not link reception and delivery move in case of RMA replacement

florian-dacosta:
Hello,

In case of returning a product to customer, I understand that the reception move is chained to the delivery move because we send the same product.
But in case of replacement, it is another product, I think it does not make sense to link the reception move to the delivery move in that case. Indeed, it will block the reservation, even if we have another product to send.
Unless I am missing something?

@chienandalu @ernestotejeda

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -1164,7 +1164,6 @@ class Rma(models.Model):
         self._action_launch_stock_rule(scheduled_date, warehouse, product, qty, uom)
         new_move = self.delivery_move_ids - moves_before
         if new_move:
-            self.reception_move_id.move_dest_ids = [(4, new_move.id)]
             self.message_post(
                 body=_(
                     "Replacement: "


pedrobaeza:
Steps to reproduce a problem?

chienandalu:
@pedrobaeza As I understand @florian-dacosta fix this is superfluous code that has to be removed for #258 to work. I can't tell the logic of linking those moves either. Maybe @ernestotejeda can recall it.

florian-dacosta:
- Install RMA module with demo data
- Create a RMA on product E-COM07
- Validate it - validate the reception
- Click on replace button and choose a substitution product : E-COM10
At this step you should have a delivery order going from WH/stock to customers
- On delivery order : click on "check availability" button.
The delivery order stays at waiting state instead of "assign"even though there are 22 E-COM10 in WH/stock

I chose a different product to be obvious (because I guess it is a normal case to be able to choose a different product than the one on the RMA if we have this option).

But testing, I feel like there is the inverse problem as well (reservation of the delivery order made when it should not)

- Install RMA module with demo data
- Create a RMA on product  E-COM06 (no product in any location)
- Validate RMA and reception step => You have 1  E-COM06 in location RMA/WH
- click on replace, and choose the same product :  E-COM06
At this step you have one delivery order for  E-COM06 from location WH/stock to Customers and you have one piece of  E-COM06 in location RMA/WH (which should not be available for the delivery order)
- Check availability on delivery order => It becomes available when there are not stock in WH/stock for  E-COM06.

chienandalu:
> At this step you have one delivery order for E-COM06 from location WH/stock to Customers and you have one piece of E-COM06 in location RMA/WH (which should not be available for the delivery order)

That's weird. Where's the stock.move.line reserving from? Is any quant affected?

florian-dacosta:
> > At this step you have one delivery order for E-COM06 from location WH/stock to Customers and you have one piece of E-COM06 in location RMA/WH (which should not be available for the delivery order)
> 
> That's weird. Where's the stock.move.line reserving from? Is any quant affected?

Yes, the quant in RMA/WH has the field reserved_quantity updated to 1.
It could also be considered has a bug in core I guess, the reason is that on stock move reservation, the location taken into account is the one of the move_orig_ids, not the one of the  current move beeing reserved.
The source issue is that we put an inconsistent move_orig_ids in this example.

pedrobaeza:
Does the original WH/OUT be linked to an MTO/dropshipping operation?

florian-dacosta:
@pedrobaeza Not sure why you ask, but no, in my example, I made a RMA with no link to anything, as the module leave this possibility.
I can do the same with a link on a delivery order with and without a mto operation and the result will be the same though.

If nobody can explain why we would like to link the reception with the delivery in case of replacement, I think we should remove it anyway. The way it is done is very problematic because in the replacement case, the delivery steps may be totally different than just a delivery order. (we could have pick/pack/ship steps or anything else) and in most of case, the reception of the dommaged product is not the step right before the delivery order.

pedrobaeza:
My reason for wanting to link them is to be shown from the original sales order. Is that lost without this link?

florian-dacosta:
The step to reproduce was not with sale nor rma_sale module, so no link with the original sale order.

With sale and rma_sale modules, this link has no impact on what is shown on the original sale order as the move_orig_ids link is purely a logistic one, usefull to manage reservation.
The RMA will still be shown on sale order, the RMA iwill still be linked to its reception and its delivery.
No change on all that

pedrobaeza:
I mean on the picking_ids smart-button, and if I don't remember bad, it's connected through that.

chienandalu:
> I mean on the picking_ids smart-button, and if I don't remember bad, it's connected through that.

I just tested that in 13.0 and 14.0 and it's not happening with replacements

pedrobaeza:
OK, so do you agree on the change?

chienandalu:
Well, indeed following the code I found that this implementation had this effect:

- Make a sale order.
- Deliver the products
- When you open the RMA wizard from the SO you'll have as many units as you delivered.
- Make the RMA and receive it.
- Now if you open the RMA wizard from the SO you'll have 0 units available for a new RMA.
- Now replace the products on the RMA with another one and deliver them.
- When you open the RMA wizard now you'll have the original units again for a new RMA, which is not correct.

chienandalu (approved):

pedrobaeza:
OK, then a gap is to be able to select the new replacement product to be RMAed from the sale, isn't it? Maybe we should handle that on `rma_sale`, but keeping this link?

chienandalu:
Well, it could be a nice to have. It could be handled from the RMA. 

Anyway this link is making the sale RMA wizard think that an RMA can be made again with the original product. This is where the wizard info is prepared:

https://github.com/OCA/rma/blob/8afb31c8836716d643c6fb43a59c050e1f644c2c/rma_sale/models/sale.py#L120-L123

pedrobaeza:
On that code, I think we should split by product, and we will have such improvement, and it's a needed one.

chienandalu:
Another side effect that @florian-dacosta was pointing (https://github.com/OCA/rma/pull/257#issuecomment-1031629172) is that picking with the product replacement can't make the stock reservation.

pedrobaeza:
Is that confirmed? I would like to have the other flow covered, being with `move_orig_ids` or with other technique.

Issue #210, opened 2021-03-12
[RFC] RMA with no product return from customer


florian-dacosta:
Hello,

I am starting to have a look at these modules to see how they can replace the old rma modules we had in older versions.
I have one main issue regarding the workflow.

From what I saw, it is mandatory to first receive the goods before going further in the process.
In my experience, it is not always the case. Depending on the firm, sometimes a replacement product may be sent without the customer sending it back first (if it is broken, with photo as proof, etc...) or if the replacement is only a missing/broken part of the product that the customer is able to change by itself. Or probably a lot if different reasons may occur.
I myself received recently a replacement piece of a good without having to send back the broken piece!

Is this something we could change? 
Maybe an option could be added on the RMA, like a flag "return merchandize neccesary" or something, that could be True by default or that would depend on an option on company level for instance.
It would be only editable on draft state. On confirmation, the rma would then directly go to received state and the workflow could go on.

I don't know these modules enough yet, to know if this would have a lot of impact or which would be the best way to achieve this.
what do you think @pedrobaeza @chienandalu @ernestotejeda ? (I am not sure who else use these modules for now)
I can investigate more and see if I can propose something else, but I'd like to have your opinion about this need first.

chienandalu:
Hi, @florian-dacosta 

I think you could play with locations in order to achieve these requirements.

> I myself received recently a replacement piece of a good without having to send back the broken piece!

In this case, you could make a sort of scrap location hanging from the RMA one. This way, you're scrapping the product from the customer location and then replacing it for another unit in your warehouse, wich is is indeed what is happening.

You can play with locations and rules in order to make more complex flows. In the end, the RMA only minds if its linked return picking is done or not.

florian-dacosta:
Hello @chienandalu 
Thanks for the hint.
Indeed, it could be kind of a workaround, but still, it adds quite a lot of steps, for something useless in that case.
If the product is not really returned, the user won't want to do the whole reception process in order to do a replacement.
He won't understand it and it would be time consuming
Mostly if it is a common case or the most common case in the firm.

I'll check some more the module to see if we could make it optional

chienandalu:
> I'll check some more the module to see if we could make it optional

In that case, I would add a check like "Scrap in customer location" that would do the thing. We should avoid showing the option for return to customer. Maybe an extra scrapped state to manage it?

florian-dacosta:
@chienandalu 

I am not sure I get your idea with the "scrap in customer location" stuff.
[...]

PR #258, opened 2022-02-07
[14.0][RMA] Allow to bypass reception step by a company

florian-dacosta:
As  talked here : https://github.com/OCA/rma/issues/210

It depends on https://github.com/OCA/rma/pull/257 for replacement to work.
Else we should just a condition here : https://github.com/OCA/rma/blob/14.0/rma/models/rma.py#L1167 To avoid link the reception move it there is no reception

@chienandalu @ernestotejeda


flotho (approved):
Thanks for the contribution
LGTM, code review and functional test

florian-dacosta:
I close since I won't be using these module finally feels free to fork it propose it again of course

PR #261, opened 2022-02-17
[FIX] rma: refund permissions

chienandalu:
Up to v12, account.invoice, was allowed for a wide variety of users. In
this version though, with account.move conversion, those permissions are
quite restricted.

We want to keep the possibility to open a refund for the RMA users.
Although only Invoicing users will be able to post it anyway.

cc @Tecnativa TT34644

please review @pedrobaeza @sergio-teruel

### DIFF
--- a/rma/security/rma_security.xml
+++ b/rma/security/rma_security.xml
@@ -91,6 +91,23 @@
             name="domain_force"
         > ['|', ('company_id', 'in', company_ids), ('company_id', '=', False)]</field>
     </record>
+    <!-- Allow to refund RMAs -->
+    <record id="rma_account_move_personal_rule" model="ir.rule">
+        <field name="name">RMA Personal Invoice</field>
+        <field ref="model_account_move" name="model_id" />
+        <field
+            name="domain_force"
+        >[('type', '=', 'out_refund'), '|', ('invoice_user_id', '=', user.id), ('invoice_user_id', '=', False)]</field>
+        <field name="groups" eval="[(4, ref('rma.rma_group_user_own'))]" />
+    </record>
+    <record id="rma_account_move_line_personal_rule" model="ir.rule">
+        <field name="name">RMA Personal Invoice Lines</field>
+        <field ref="model_account_move_line" name="model_id" />
+        <field
+            name="domain_force"
+        >[('move_id.type', '=', 'out_refund'), '|', ('move_id.invoice_user_id', '=', user.id), ('move_id.invoice_user_id', '=', False)]</field>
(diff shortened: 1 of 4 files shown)


sergio-teruel (approved):

rma/security/rma_security.xml
```diff
@@ -91,6 +91,23 @@
             name="domain_force"
         > ['|', ('company_id', 'in', company_ids), ('company_id', '=', False)]</field>
     </record>
+    <!-- Allow to refund RMAs -->
+    <record id="rma_account_move_personal_rule" model="ir.rule">
+        <field name="name">RMA Personal Invoice</field>
+        <field ref="model_account_move" name="model_id" />
+        <field
+            name="domain_force"
+        >[('type', '=', 'out_refund'), '|', ('invoice_user_id', '=', user.id), ('invoice_user_id', '=', False)]</field>
```
pedrobaeza, line 100: Can we restrict to RMA ones?

rma/security/rma_security.xml
```diff
@@ -91,6 +91,23 @@
...
+        <field name="groups" eval="[(4, ref('rma.rma_group_user_own'))]" />
+    </record>
+    <record id="rma_account_move_line_personal_rule" model="ir.rule">
+        <field name="name">RMA Personal Invoice Lines</field>
+        <field ref="model_account_move_line" name="model_id" />
+        <field
+            name="domain_force"
+        >[('move_id.type', '=', 'out_refund'), '|', ('move_id.invoice_user_id', '=', user.id), ('move_id.invoice_user_id', '=', False)]</field>
```
pedrobaeza, line 108: Aren't related stored fields at aml level for avoiding certain extra queries? And moreover, I think that combining several subqueries doesn't work properly in current ORM

chienandalu: Well, I'm copying Odoo's own rule for sale users: 

https://github.com/OCA/OCB/blob/27edc28a2c19fad80053e668a8c72e7a629e4c1a/addons/sale/security/sale_security.xml#L165

rma/security/rma_security.xml
chienandalu: Not at the document level

rma/security/rma_security.xml
pedrobaeza: OK, I see there's no cached value at aml for type. There's one for `state` though. We will need to live with this perf drill. About the combination, it seems it's working for the other level, so I recalled badly the ORM limitation.

pedrobaeza (approved):

pedrobaeza:
To v14, please

PR #264, opened 2022-03-22
[FIX] rma_sale: subscribe portal users to RMA notifications

chienandalu:
When a portal user opens an RMA subscribe them to RMA notifications by default.

cc @Tecnativa TT35269

please review @ernestotejeda @pedrobaeza

### DIFF
--- a/rma_sale/controllers/sale_portal.py
+++ b/rma_sale/controllers/sale_portal.py
@@ -67,9 +67,13 @@ class CustomerPortal(CustomerPortal):
         rma = wizard.sudo().create_rma(from_portal=True)
         for rec in rma:
             rec.origin += _(" (Portal)")
-        # Add the user as follower of the created RMAs so they can
-        # later view them.
+        # Add the user as follower of the created RMAs so they can later view them.
         rma.message_subscribe([request.env.user.partner_id.id])
+        # Subscribe the user to the notification subtype so he receives the confirmation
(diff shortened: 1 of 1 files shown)


rma_sale/controllers/sale_portal.py
```diff
@@ -68,8 +68,11 @@ def request_rma(self, order_id, access_token=None, **post):
         for rec in rma:
             rec.origin += _(" (Portal)")
         # Add the user as follower of the created RMAs so they can
-        # later view them.
-        rma.message_subscribe([request.env.user.partner_id.id])
+        # later view them. Subscribe the user to the notification subtype
+        rma.message_subscribe(
+            partner_ids=[request.env.user.partner_id.id],
+            subtype_ids=request.env.ref("rma.mt_rma_notification").ids,
```
pedrobaeza: With this you are subscribing it only to that subtype, not the rest of the standard ones, so I think this is not the desired effect.

chienandalu: True, I thought the parameter added to default ones.

pedrobaeza: I thought the same until I discovered this bad surprise, as I think there's no API for that.

pedrobaeza (approved):

ernestotejeda (approved):

PR #271, opened 2022-05-12
15.0 mig rma sale

ghost:
I followed the instructions which now make sense ;).

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0


chienandalu (changes requested):
Anyway you should migrate the modules one by one in separate PRs

PR #280, opened 2022-06-13
RMA Sale improvments

mt-software-de:
Removes field allowed_product_ids computation -> changes it to a normal domain behavior for the field product_id.
On opening the a new rma allowed_product_ids is computed, with a lot of product data this method will run a little bit
With change search happens only if someone wants to select a product

Changes onchange_partner_id -> changes the domain of the field order_id
if the partner is empty or the current selected order is not one of the partners commercial partner
then the order domain gets reset. 

Changes _onchange_order_id
This prefills the partner values and also changes the product_id domain

Adds _onchange_order_id_product_id
Sets picking and move to false if order or product is empty or if there is no moves that fits to the product
on a fitting move the the move and picking will be set

Adds _onchange_picking_id
sets the move automatically


mt-software-de:
@ernestotejeda could you apply the "no stale" label?

pedrobaeza:
This PR should solve in a better way the performance problem: #326 (will be put for 14.0 when finished).

PR #282, opened 2022-06-22
[13.0][FIX] rma: Avoid set invoice_payment_term_id from partner default value in refund.

victoralmau:
Changes done:
- [x] Avoid set `invoice_payment_term_id` from partner default value in refund.

@Tecnativa TT37690

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -1008,6 +1008,8 @@ class Rma(models.Model):
         """
         self.ensure_one()
         invoice_form.partner_id = self.partner_invoice_id
+        # Avoid set partner default value
+        invoice_form.invoice_payment_term_id = self.env["account.payment.term"]
 
     def _prepare_refund_line(self, line_form):
         """ Hook method for preparing a refund line Form.
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -39,6 +39,9 @@ class TestRma(SavepointCase):
             {
                 "name": "Partner test",
                 "property_account_receivable_id": cls.account_receiv.id,
+                "property_payment_term_id": cls.env.ref(
+                    "account.account_payment_term_30days"
+                ).id,
             }
         )
         cls.partner_invoice = cls.res_partner.create(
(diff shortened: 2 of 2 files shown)


chienandalu (approved):

rma/models/rma.py
```diff
@@ -1008,6 +1010,8 @@ def _prepare_refund(self, invoice_form, origin):
         """
         self.ensure_one()
         invoice_form.partner_id = self.partner_invoice_id
+        # Avoid set partner default value
+        invoice_form.invoice_payment_term_id = self.env["account.payment.term"]
```
pedrobaeza, line 1012: Use False instead

rma/models/rma.py
```diff
@@ -623,7 +623,9 @@ def action_refund(self):
                 self.env["account.move"]
                 .sudo()
                 .with_context(
-                    default_type="out_refund", company_id=rmas[0].company_id.id,
+                    default_type="out_refund",
+                    company_id=rmas[0].company_id.id,
+                    default_invoice_date_due=fields.Date.today(),
```
pedrobaeza: Why putting today instead of empty?

rma/models/rma.py
victoralmau: I had tried to define False, but in a Form() it is not possible.

rma/models/rma.py
victoralmau: The idea is to have the same behavior as odoo core in the wizard when creating a credit note, do you think it is not necessary?

rma/models/rma.py
pedrobaeza: ok

rma/models/rma.py
pedrobaeza: It's not the same, as in the wizard you have a date to define, and not in this case. Better to leave it without value

pedrobaeza (approved):

pedrobaeza:
With this, the date is not filled, right?

victoralmau:
Currently in 13.0 (before this PR), when creating the invoice, the due date is always defined in the invoice (if there is a payment term, the corresponding one is defined, if there is no payment term, today is defined).
With this change, the expiration date is always defined (same as before).

pedrobaeza:
What I want is to not have the date set on the refund. Is this possible?

PR #288, opened 2022-07-25
15.0 mig rma

ghost:
with working tests


rma/__manifest__.py
```diff
@@ -3,7 +3,7 @@
 {
     "name": "Return Merchandise Authorization Management",
     "summary": "Return Merchandise Authorization (RMA)",
-    "version": "14.0.3.1.0",
+    "version": "15.0.1.0.0",
```
chienandalu, line 6: Looks like you mixed prettir stuff with the proper migration commit...

jacekmichalski: @nikolausweingartmair can you change it?

victoralmau:
Superseed by: https://github.com/OCA/rma/pull/290

rma/__manifest__.py
ghost: @chienandalu:
What do you mean by prettier stuff?
If you mean the change in the templating, it's necessary to pass the tests ...

Bildschirmfoto 2022-08-09 um 13 20 24

chienandalu: Check guidelines: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0#how-to

When you migrate from one version to another, prettier rules could changes, so some code gets recomposed. This kind of changes should be isolated in a commit to ease the reading of actual migration changes.

ghost: Is it really necessary to split it? If so should I do the mandatory linting updates still in V14 oder after the migration in V15?

Bildschirmfoto 2022-08-09 um 14 14 11

pedrobaeza: It's needed to follow the procedure, that states that the automatic pre-commit stuff (black, prettier, isort) should be done in a separate commit. Then, the warnings that you are showing, can be attended in the migration commit for clearing them if you want. They are not blocking, but advised to be attended. Anyways, there's another PR with such changes. Do you want to continue with this PR? If so, you have priority being the first one doing it, but please attend such comments.

ghost: now I got your point. I made a new pull request. If it's not working please use the other one, then I try a different module to get my first contribution ;).

PR #299, opened 2022-09-01
[13.0][FIX] rma_sale: Apply the right taxes from sales order line to refund (or keep line without taxes).

victoralmau:
Apply the right taxes from sales order line to refund (or keep line without taxes).

@Tecnativa


pedrobaeza:
This is not correct. Taxes depend on fiscal position. Maybe the problem is on the delivery address?

victoralmau:
You are right, this is not correct.

PR #300, opened 2022-09-01
[13.0][IMP] rma_sale: Set sale_line_id to refund line.

victoralmau:
Set sale_line_id to refund line.

@Tecnativa

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -105,3 +105,8 @@ class Rma(models.Model):
         if line:
             line_form.discount = line.discount
             line_form.sequence = line.sequence
+
+    def action_refund(self):
+        super().action_refund()
+        if self.refund_line_id and self.sale_line_id:
+            self.refund_line_id.sale_line_ids = [(4, self.sale_line_id.id)]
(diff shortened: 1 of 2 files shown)


pedrobaeza:
What is the use case? Putting the first invoice found doesn't feel correct.

victoralmau:
> What is the use case? Putting the first invoice found doesn't feel correct.

The use case would be:
- Sales order is confirmed.
- Products are shipped.
- Invoice is created.
- RMA is created.
- Refund invoice is done from RMA.

IMO this refund should be linked to the initial invoice.

pedrobaeza:
Not really, as it's not a reversed entry. It happens the same if you return a product, and make the refund of such return: no reverse entry is linked. The important thing is that the same sales line is linked in both invoice and refund, which is already happening, isn't it?

victoralmau:
Ok, checking in runboat the field `sale_line_ids` (in the invoice line) is not being defined in the refund created from RMA.

pedrobaeza:
Then that's the thing to fix

victoralmau:
Changes done.

pedrobaeza:
Oh, wait, I'm remembering that this was discarded because if not, the SO is put again as invoiceable. Can you check the history to see if it's true?

victoralmau:
You are absolutely right, I have tested it with this change applied and it is just what happens (the SO is put again as invoiceable).

PR #305, opened 2022-09-05
[14.0][FIX] rma

Freni-OSI:

PR #306, opened 2022-09-16
[14.0][FIX] rma (Action v14 compatibility)

patrickrwilson:
Same fix as #305 but for delivery and rma on picking

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -788,8 +788,6 @@ class Rma(models.Model):
     def action_view_receipt(self):
         """Invoked when 'Receipt' smart button in rma form view is clicked."""
         self.ensure_one()
-        # Force active_id to avoid issues when coming from smart buttons
-        # in other models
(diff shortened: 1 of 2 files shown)


rma/models/rma.py
```diff
@@ -788,8 +788,6 @@ def action_preview(self):
     def action_view_receipt(self):
         """Invoked when 'Receipt' smart button in rma form view is clicked."""
         self.ensure_one()
```
pedrobaeza, line 790: Has this patch been lost?

patrickrwilson: I don't think so, seems to work just fine. Example using rma_sale, which adds a smart button on a SO, clicking through the smart button on the SO, the smart buttons on the RMA work correctly with this change.

pedrobaeza (approved):

pedrobaeza:
We can fast-track it:

/ocabot merge patch

Please fw-port it to v15

PR #308, opened 2022-09-19
[14.0][FIX] rma (Make additional actions 14.0 compatible)

patrickrwilson:
There are more places where the action needed updated with _for_xml_id to be 14.0 compatible and allow users without settings access to use the button actions. Similar fixes as #305 #306


patrickrwilson:
@pedrobaeza I didn't notice the other methods that had these same issues, so here is the fix for those too. I will also port forward to v15 when time allows.

pedrobaeza (approved):

pedrobaeza:
Let's finish it:

/ocabot merge patch

PR #313, opened 2022-09-22
[14.0][FIX] rma_sale: Allow to request a rma from portal order page only with portal user permission

victoralmau:
Allow to request a rma from portal order page only with portal user permission

@Tecnativa

### DIFF
--- a/rma_sale/README.rst
+++ b/rma_sale/README.rst
@@ -99,6 +99,7 @@ Contributors
   * Ernesto Tejeda
   * Pedro M. Baeza
   * David Vidal
+  * Víctor Martínez
 
 * Chafique Delli <chafique.delli@akretion.com>
 * Giovanni Serra - Ooops <giovanni@ooops404.com>
--- a/rma_sale/controllers/sale_portal.py
+++ b/rma_sale/controllers/sale_portal.py
@@ -25,7 +25,7 @@ class CustomerPortal(CustomerPortal):
         except (AccessError, MissingError):
             return request.redirect("/my")
         order_obj = request.env["sale.order"]
-        wizard_obj = request.env["sale.order.rma.wizard"]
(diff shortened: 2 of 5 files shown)


victoralmau:
Another related and similar issue: right now it is not possible to generate a rma from the order page without a logged in user, this is correct, right?

pedrobaeza:
>  right now it is not possible to generate a rma from the order page without a logged in user, this is correct, right?

Do you mean having the link with a token? That's correct IMO.

pedrobaeza (approved):

rma_sale/security/ir.model.access.csv
```diff
@@ -1,3 +1,5 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_sale_order_rma_wizard_user_all,sale.order.rma.wizard.user.all,model_sale_order_rma_wizard,rma.rma_group_user_all,1,1,1,1
+access_sale_order_rma_wizard_portal_all,sale.order.rma.wizard.portal.all,model_sale_order_rma_wizard,base.group_portal,1,1,1,1
```
chienandalu: Wouldn't a `sudo` fit this issue better? We're giving portal users whole rights to that model :thinking:

victoralmau: This is just about a wizard, do you think this could be a security issue?

vincenzo-riggi:
tested and working

rma_sale/security/ir.model.access.csv
pedrobaeza: Please change to sudo as commented by chat

victoralmau:
Changes done.

rma-portal

Linked RMAs does not work (rma listing) if no user is logged in.

pedrobaeza:
@chienandalu do you agree?

chienandalu (approved):
Great :)

PR #318, opened 2022-10-26
[13.0][FIX] rma: multiple substitution moves

chienandalu:
The product replacement could explode into several moves like in the case of MRP BoM kits.

cc @Tecnativa TT40194

please review @pedrobaeza @victoralmau

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -1134,11 +1134,14 @@ class Rma(models.Model):
         self._ensure_can_be_replaced()
         moves_before = self.delivery_move_ids
         self._action_launch_stock_rule(scheduled_date, warehouse, product, qty, uom)
-        new_move = self.delivery_move_ids - moves_before
-        if new_move:
-            self.reception_move_id.move_dest_ids = [(4, new_move.id)]
-            self.message_post(
-                body=_(
+        new_moves = self.delivery_move_ids - moves_before
+        self.reception_move_id.move_dest_ids += new_moves
+        body = ""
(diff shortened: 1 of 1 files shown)


rma/models/rma.py
```diff
@@ -1134,24 +1134,28 @@ def create_replace(self, scheduled_date, warehouse, product, qty, uom):
...
-                    'data-oe-model="stock.picking" data-oe-id="%d">%s</a>) '
-                    "has been created."
-                )
-                % (
-                    new_move.id,
-                    new_move.name_get()[0][1],
-                    new_move.picking_id.id,
-                    new_move.picking_id.name,
+        new_moves = self.delivery_move_ids - moves_before
+        if new_moves:
```
pedrobaeza: No need of this extra `if`. Just put the `for` and check if body has content.

pedrobaeza:
Is this a real case?

chienandalu:
> Is this a real case?

Yes. For the case that the substituted product delivery explodes in serveral moves like in mrp.bom kits

pedrobaeza:
OK, then please document that in the code saying why there can be more than one move.

rma/models/rma.py
chienandalu: Simplified

victoralmau (approved):

pedrobaeza (approved):

pedrobaeza:
/ocabot merge patch

Please fw-port it.

PR #317, opened 2022-10-19
[14.0][IMP] rma_sale: Replace allowed_xx fields by field_domain + web_domain_field dependency

victoralmau:
Related to: https://github.com/OCA/rma/pull/280 (different approach)

Replace `allowed_xx` fields by field_domain + `web_domain_field` dependency

Locked by:
- https://github.com/OCA/web/pull/2322

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa

PR #326, opened 2022-12-05
[13.0][IMP] rma_sale: Domains performance

pedrobaeza:
For avoiding a big list of IDs being transferred when no sales order is selected on the RMA, we have changed domains to make use of the possibility of pyjs (http://pyjs.org/) expressions allowed in the domains.

No ternary operators nor list sums are allowed in pyjs, but using a combination of allowed IDs with a controlled length of values + and/or operators to switch domains is enough for having the right performance and avoid to depend on other modules like `web_domain_field`.

@Tecnativa

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -20,7 +20,10 @@ class Rma(models.Model):
     allowed_picking_ids = fields.Many2many(
         comodel_name="stock.picking", compute="_compute_allowed_picking_ids",
     )
-    picking_id = fields.Many2one(domain="[('id', 'in', allowed_picking_ids)]")
+    picking_id = fields.Many2one(
+        domain="(order_id or partner_id) and [('id', 'in', allowed_picking_ids)] or "
+        "[('state', '=', 'done'), ('picking_type_id.code', '=', 'outgoing')] "
+    )
     allowed_move_ids = fields.Many2many(
         comodel_name="sale.order.line", compute="_compute_allowed_move_ids",
     )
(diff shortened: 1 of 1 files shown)


rma_sale/models/rma.py
```diff
@@ -17,10 +17,11 @@ class Rma(models.Model):
         readonly=True,
         states={"draft": [("readonly", False)]},
     )
-    allowed_picking_ids = fields.Many2many(
-        comodel_name="stock.picking", compute="_compute_allowed_picking_ids",
+    picking_id = fields.Many2one(
+        domain="[('state', '=', 'done'), ('picking_type_id.code', '=', 'outgoing')] "
+        "+ [('sale_id', '=', order_id)] if order_id else [] "
+        "+ [('partner_id', 'child_of', commercial_partner_id)] if partner_id else []"
```
rousseldenis: @pedrobaeza Are right fields really evalutated dynamically?

pedrobaeza: I'm trying. On the worst case, we will need to repeat the domain.

rousseldenis: I would say if you want something more efficient, web_domain_field is the right solution

pedrobaeza: Efficiency can be the same, and what I don't want is to depend on it. I'm seeing Pyjs (http://pyjs.org/) doesn't support ternary operators, nor list sums, but I'm working on adapting them.

rousseldenis: Sectarism will loose you :smiley: adding a dependency is not evil if the addition is good for the job. Hacking is

pedrobaeza: This is not hacking. It's using the framework possibilities. Already done and working.

pedrobaeza: Judge yourself that is very simple.

pedrobaeza: #317 is the approach with `web_domain_field`, and apart from adding the new dependency, it's not simpler.

victoralmau (approved):

ernestotejeda (approved):

chienandalu (approved):
Good to know the pyjs evaluation bounds to solve this kind of issues :+1:

rousseldenis:
As the changes here do not support full domain (still some ids in domain), I continue to believe that usage of web_domain_field is better as domain will be fully evaluated by py.js.

See https://github.com/OCA/web/pull/2356/files#diff-a27a472baf2dfd949377b5633b3a92e11eef877e8da14c2d5a769fc76cf425f1R54

pedrobaeza:
That ID list is very short, so it's good and very traceable.

/ocabot merge minor

PR #334, opened 2022-12-19
[14.0][IMP] rma configured flow

mt-software-de:
This PR would add the possibility to configure the moment,
when we are able to create a reception, return and refund

It adds for each process a own field on the rma.operation:
create_receipt_timing
create_return_timing
create_refund_timing

By configure a field as on_confirm
the process will be created on action_confirm
Also we can now disable a process by set the field to no

With the addon rma_sale we can also configure the rma to no create refunds
but change the qty_delivered of the sale.order.line to create a refund over the sale.order

Depends on:
https://github.com/OCA/rma/pull/350


chienandalu (commented):
Thanks @mt-software-de . When you're ready, a rationale for these changes will be needed

pedrobaeza:
I think the rationale should go before the code changes, to validate it.

mt-software-de:
Here is our scenario.
The Customer Support - knows at the beginning how the rma will take place
so they can select what pickings - refund should be created
So they are doing the confirm and everything is done.
For example: They don't have to wait for the return from the customer

There a different flows, as example
- only returning
- only refund
- return and resend the goods (also at the same time, not only when return happend)
- ...

Thats why the idea came up to make in configurable on the operation.
If its configured on the operation everything is created on action_confirm

@pedrobaeza @chienandalu

mt-software-de:
FYI @jbaudoux

mt-software-de:
@chienandalu and @pedrobaeza now i'm done with my changes. 
FYI: @jbaudoux

rma_sale_mrp/models/rma.py
```diff
@@ -74,3 +77,11 @@ def action_draft(self):
                 )
             )
         return super().action_draft()
+
+    def _prepare_procurement_values(
+        self, warehouse=None, scheduled_date=None, group=None
+    ):
+        values = super()._prepare_procurement_values()
```
vuwnevska:
```suggestion
        values = super()._prepare_procurement_values(warehouse, scheduled_date, group)
```

rma/models/rma.py
```diff
@@ -782,13 +846,14 @@ def action_refund(self):
                 refund = invoice_form.save()
                 line = refund.invoice_line_ids.filtered(lambda r: not r.rma_id)
                 line.rma_id = rma.id
-                rma.write(
+                _vals = deepcopy(vals)
```
jbaudoux, line 849: you could just build vals here without requiring deepcopy

rma/models/rma.py
```diff
@@ -1192,18 +1257,14 @@ def _create_delivery(self, scheduled_date=None, qty=None, uom=None):
...
-        rmas_to_return = self.filtered(
-            lambda rma: rma.can_be_returned and rma._product_is_storable()
-        )
-        rmas_to_return._create_delivery(scheduled_date, qty, uom)
+    def _post_process_create_deliveries(self):
         pickings = defaultdict(lambda: self.browse())
-        for rma in rmas_to_return:
-            picking = rma.delivery_move_ids.picking_id.sorted("id", reverse=True)[0]
+        for rma in self:
+            # TODO: warum sind hier mehr als eins vorhanden
```
jbaudoux, line 1263: comment can be dropped

rma/models/rma_operation.py
```diff
@@ -1,8 +1,16 @@
...
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import fields, models
 
+TIMING_ON_CONFIRM = "on_confirm"
+TIMING_AFTER_RECEIPT = "after_receipt"
+TIMING_NO = "no"
+
+TIMING_ON_CONFIRM_STR = "On confirm"
+TIMING_AFTER_RECEIPT_STR = "After receipt"
```
jbaudoux, line 12: I would not create globals for the labels

jbaudoux (commented):
some remarks

mt-software-de:
> some remarks

Thanks for the review but nothing will happen here in the near future. Due to the recent events in the repo rma, the contribution for this repo is stopped. Therefore I'm closing this PR.

pedrobaeza:
It's a pity that you take this decision, but I think the only one that is losing here is you. Anyway, part of the reasons for not accepting directly your PRs is changes that over-complicate the review and bring no value to the existing code. For example, in this PR, these changes are meaningless and even resting readability:

imagen

I'm also guilty here for not wanting to fight more battles - I tried in https://github.com/OCA/rma/pull/350#discussion_r1241745204 and https://github.com/OCA/rma/pull/350#discussion_r1241745204 and some other places, but each step means too much time, or the alternative to have unknown worse code to maintain, taking us to the option to rewrite from your foundations.

PR #335, opened 2022-12-20
[14.0][IMP] rma_sale: improve onchange methods

mt-software-de:
Changes:

**_onchange_partner_id:**
Before the order_id was always set to False
Now it's only set to False
if there is no partner selected or
the selected order_id.partner_id is not a child of the current selected partner_id.commercial_partner_id

**_onchange_order_id:**
if the order_id is set the corresponding partner_fields are getting set

New onchange methods:

**_onchange_order_id_product_id:**
if the order_id and the product_id is set
it tries to set the move_id with the latest shipped move with the same product_id
as well as the picking_id

**_onchange_picking_id:**
if the picking and product ist
tries to set the move by selecting the corresponding move


mt-software-de:
@rousseldenis could you please open this PR again?

mt-software-de:
ping @chienandalu do you have time for a review?

pedrobaeza:
You should check how it's done in 16.0, and backport it here, as that's a better way to go IMO.

mt-software-de:
> You should check how it's done in 16.0, and backport it here, as that's a better way to go IMO.

Thx. But i can not see any change which i can backport and matching my changes.
Could you be maybe more specific one what you mean?

pedrobaeza:
I mean the conversion to computed writable fields.

pedrobaeza:
See 54e08e856fd2f8d07057eef3ca12b2d395fcfb23

PR #338, opened 2023-02-11
[13.0][IMP] rma_sale: Link refund line with origin sale line

pedrobaeza:
Steps to reproduce:

- Create a sales order with an storable product with invoicing policy on delivered quantities.
- Confirm it and deliver the product.
- Invoice the order.
- Do an RMA, return it, and refund it.

Result: the delivered quantity is 1 instead of 0.
    
This is because the refund generated from the RMA is not linked to sales order line, nor the RMA reception move. This is done because other operations are performed:
    
- Be replaced.
- Be changed by other product.
    
And we don't also want that meanwhile the RMA is being performed, the sales order is pending to invoice.

But when the refund has been done, we have it clear, so let's link both and have the sales statistics correct.

@Tecnativa TT41645

### DIFF
--- a/rma_sale/__manifest__.py
+++ b/rma_sale/__manifest__.py
@@ -12,6 +12,7 @@
     "license": "AGPL-3",
     "depends": ["rma", "sale_stock"],
     "data": [
+        "views/account_move_views.xml",
         "views/assets.xml",
         "views/report_rma.xml",
         "views/rma_views.xml",
--- a/rma_sale/models/__init__.py
+++ b/rma_sale/models/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import account_move
 from . import res_company
 from . import res_config_settings
 from . import rma
--- a/rma_sale/models/account_move.py
+++ b/rma_sale/models/account_move.py
@@ -22,3 +22,9 @@ class AccountMove(models.Model):
             if rma.sale_line_id:
                 rma._link_refund_with_reception_move()
         return super().button_draft()
+
+    def unlink(self):
+        """If the invoice is removed rollback the quantities correction"""
+        for rma in self.invoice_line_ids.rma_id.filtered("sale_line_id"):
+            rma._unlink_refund_with_reception_move()
+        return super().unlink()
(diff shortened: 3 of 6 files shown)


pedrobaeza:
Investigating more, this is not a FIX, but an IMP. Initial text changed according this.

pedrobaeza:
My only doubt here about this improvement is that the sales statistics are changed time after you "close" such data. Imagine an RMA done 6 months after the sales for a broken stuff. There are no possible replacements, so you decide to refund. This will decrease the sold quantity of 6 months ago. It's correct, because you didn't have any revenue for that, but is it convenient?

One possible option is to have a system parameter (but it's not black or white), and another to query on each return if decrease sales statistics or not, but most users won't know what this means (even if we give extended instructions), as they are simple transactional users.

What community think about this?

chienandalu (commented):
I think this new behavior could break some the expected workflows for the current app users, so I vote to put it behind a company setting

pedrobaeza:
Please tell me what can break. This operation is atomic, so no intermediate inconsistent data.

chienandalu:
> Please tell me what can break.

The user expecting that the invoiced/delivered quantities won't change by the effect of RMAs

pedrobaeza:
Well, I see this as something expected (and the PR has been motivated by that), so saying users the "new behavior" should be enough, but masking it under an option deactivated by default will provoke the undiscoverability of the feature.

chienandalu (approved):
Let's go with it then :+1:

ernestotejeda (approved):

chienandalu (changes requested):
I've found that splitting the RMA leads to inconsistencies in delivered and invoiced quantities.

For example, if we receive an RMA for 5 units and return 2 to the customer and refund the remaining 3, we'll have a count of 0 delivered and 2 invoiced.

This is because once the reception is linked to the sale line we can't split those outcomes.

chienandalu:
> I've found that splitting the RMA leads to inconsistencies in delivered and invoiced quantities.
> 
> For example, if we receive an RMA for 5 units and return 2 to the customer and refund the remaining 3, we'll have a count of 0 delivered and 2 invoiced.
> 
> This is because once the reception is linked to the sale line we can't split those outcomes.

Given this issue, I think this should go at least behind a company setting as we discussed earlier.

pedrobaeza:
This fix/imp has been finished on 15.0 in #354, so anyone wanted to have it in this version, please cherry pick from that PR.

PR #339, opened 2023-02-22
[15.0][IMP] stock_production_lot_warranty: re-usable _get_warranty_exp_date

florentx:
Need a re-usable method for extensibility

### DIFF
--- a/stock_production_lot_warranty/models/stock_production_lot.py
+++ b/stock_production_lot_warranty/models/stock_production_lot.py
@@ -2,12 +2,11 @@
 # Copyright (C) 2021 Serpent Consulting Services
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from datetime import datetime, timedelta
-
 from dateutil.relativedelta import relativedelta
 
 from odoo import api, fields, models
-from odoo.tools import DEFAULT_SERVER_DATE_FORMAT
+
+DELTA_TYPES = ("day", "week", "month", "year")
 
 
 class StockProductionLot(models.Model):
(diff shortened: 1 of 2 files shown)


i-vyshnevska (approved):

stock_production_lot_warranty/models/stock_production_lot.py
```diff
@@ -15,34 +15,25 @@ class StockProductionLot(models.Model):
...
+            warranty_delta = timedelta(days=duration)
+        elif duration_type == "week":
+            warranty_delta = timedelta(weeks=duration)
+        elif duration_type == "month":
+            warranty_delta = relativedelta(months=+duration)
+        elif duration_type == "year":
+            warranty_delta = relativedelta(years=+duration)
+        else:
+            raise RuntimeError(f"Invalid warranty_type {duration_type!r}")
+        return (start_date + warranty_delta).strftime(DEFAULT_SERVER_DATE_FORMAT)
```
grindtildeath: Is the `strftime` needed here? IIRC Odoo 15.0 allows to write Date/Datetime type without the need to convert to string.

stock_production_lot_warranty/models/stock_production_lot.py
```diff
@@ -15,34 +15,25 @@ class StockProductionLot(models.Model):
 
     warranty_exp_date = fields.Date(string="Warranty Expiration Date")
 
+    def _get_warranty_exp_date(self, start_date):
```
grindtildeath:
```suggestion
    def _get_warranty_exp_date(self, start_date=None):
        if start_date is None:
            start_date = date.today()
```

stock_production_lot_warranty/models/stock_production_lot.py
```diff
@@ -15,34 +15,25 @@ class StockProductionLot(models.Model):
...
-                    datetime.now()
-                    + relativedelta(months=+self.product_id.product_tmpl_id.warranty)
-                ).strftime(DEFAULT_SERVER_DATE_FORMAT)
-            elif warranty_type == "year":
-                time = (
-                    datetime.now()
-                    + relativedelta(years=+self.product_id.product_tmpl_id.warranty)
-                ).strftime(DEFAULT_SERVER_DATE_FORMAT)
-            self.warranty_exp_date = time
+        self.warranty_exp_date = self._get_warranty_exp_date(datetime.now())
```
grindtildeath: IMO we can use a default value in the function for the start date (cf suggestion above).
Moreover, if the field is of Date type, we can use Date instead of Datetime.

stock_production_lot_warranty/models/stock_production_lot.py
```diff
@@ -15,34 +15,25 @@ class StockProductionLot(models.Model):
...
+        if not self.product_id:
+            return False
+        duration_type = self.product_id.product_tmpl_id.warranty_type
+        duration = self.product_id.product_tmpl_id.warranty
+        if not duration_type or not duration:
+            return False
+        if duration_type == "day":
+            warranty_delta = timedelta(days=duration)
+        elif duration_type == "week":
+            warranty_delta = timedelta(weeks=duration)
```
grindtildeath: Wouldn't it make more sense to harmonize delta types and use `relativedelta` instead of `timedelta`?

grindtildeath (commented):
Looks ok, some nitpicking though :)

grindtildeath (approved):
Nice :+1:

simahawk (approved):

PR #341, opened 2023-03-09
[13.0][FIX] rma_sale_mrp /rma : RMA reopening cases

chienandalu:
**[FIX] rma_sale_mrp: avoid reopening of kit RMAs** 

- As we rely on the kit_qty field have the kit/component relation an that's computed when the RMA is created, we want to avoid reopening the RMA to avoid inconsitencies. The user should create a new RMA.

**[FIX] rma: avoid reopening a cancelled RMA when the reception is deleted** 

- When we delete the reception for an RMA, we're setting it to draft automatically so we can confirm it again and create a new reception. This is unconvenient when the RMA is cancelled, as we don't wan't to reopen it automatically.

cc @Tecnativa TT41943

### DIFF
--- a/rma_sale_mrp/models/rma.py
+++ b/rma_sale_mrp/models/rma.py
@@ -62,3 +62,13 @@ class Rma(models.Model):
                     "state": "refunded",
                 }
             )
+
+    def action_draft(self):
+        if self.filtered(lambda r: r.state == "cancelled" and r.phantom_bom_product):
+            raise UserError(
+                _(
+                    "To avoid kit quantities inconsistencies it is not possible to convert "
+                    "to draft a cancelled RMA. You should do a new one from the sales order."
+                )
+            )
+        return super().action_draft()
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -28,7 +28,9 @@ class StockMove(models.Model):
         rma_receiver = self.sudo().mapped("rma_receiver_ids")
         rma = self.sudo().mapped("rma_id")
         res = super().unlink()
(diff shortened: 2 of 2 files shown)


rma_sale_mrp/models/rma.py
```diff
@@ -62,3 +62,13 @@ def action_refund(self):
                     "state": "refunded",
                 }
             )
+
+    def action_draft(self):
+        if self.filtered(lambda r: r.state == "cancelled" and r.phantom_bom_product):
+            raise UserError(
+                _(
+                    "To avoid kit quantities inconsitencies it is not possible to convert "
```
pedrobaeza:
```suggestion
                    "To avoid kit quantities inconsistencies it is not possible to convert "
```

rma_sale_mrp/models/rma.py
```diff
@@ -62,3 +62,13 @@ def action_refund(self):
...
+                    "to draft a cancelled RMA. You should do new one from the sales order."
```
pedrobaeza:
```suggestion
                    "to draft a cancelled RMA. You should do a new one from the sales order."
```

rma_sale_mrp/models/rma.py
chienandalu: Done

rma_sale_mrp/models/rma.py
chienandalu: Done

pedrobaeza (approved):

victoralmau (approved):

PR #342, opened 2023-03-14
[14.0][FIX] rma_sale_mrp /rma : RMA reopening cases

chienandalu:
Fw of #341 

**[FIX] rma_sale_mrp: avoid reopening of kit RMAs** 

- As we rely on the kit_qty field have the kit/component relation an that's computed when the RMA is created, we want to avoid reopening the RMA to avoid inconsitencies. The user should create a new RMA.

**[FIX] rma: avoid reopening a cancelled RMA when the reception is deleted** 

- When we delete the reception for an RMA, we're setting it to draft automatically so we can confirm it again and create a new reception. This is unconvenient when the RMA is cancelled, as we don't wan't to reopen it automatically.

cc @Tecnativa TT41943

### DIFF
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -35,7 +35,9 @@ class StockMove(models.Model):
         rma_receiver = self.sudo().mapped("rma_receiver_ids")
         rma = self.sudo().mapped("rma_id")
         res = super().unlink()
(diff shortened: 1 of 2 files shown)


victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #344, opened 2023-03-21
[13.0][FIX] rma: avoid refunding in picking return wizard

chienandalu:

cc @Tecnativa TT42021

ping @pedrobaeza

### DIFF
--- a/rma/wizard/stock_picking_return.py
+++ b/rma/wizard/stock_picking_return.py
@@ -29,6 +29,9 @@ class ReturnPicking(models.TransientModel):
                 [("company_id", "=", self.picking_id.company_id.id)]
             ).mapped("rma_loc_id")
             rma_loc_domain = [("id", "child_of", rma_loc.ids)]
+            # We want to avoid setting the return move `to_refund` as it will change
+            # the delivered quantities in the sale and set them to invoice.
+            self.product_return_moves.to_refund = False
         else:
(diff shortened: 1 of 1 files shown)


rma/wizard/stock_picking_return.py
```diff
@@ -29,6 +29,9 @@ def _onchange_create_rma(self):
                 [("company_id", "=", self.picking_id.company_id.id)]
             ).mapped("rma_loc_id")
             rma_loc_domain = [("id", "child_of", rma_loc.ids)]
+            # We want to avoid setting the return move `to_refund` as it will change
+            # the delivered quantities in the sale and set them to invoice.
+            self.product_return_moves.to_refund = False
```
pedrobaeza, line 34: Singleton error if not?

```suggestion
            self.product_return_moves.update({"to_refund": False})
```

ernestotejeda (approved):

rma/wizard/stock_picking_return.py
chienandalu: Not a problem at all as it's a chaing of recordsets. You could even do this with no harm:

```python
env["stock.return.picking"].product_return_moves.to_refund = False
```

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
Please fw-port it to upper versions

PR #345, opened 2023-03-27
[14.0][FIX] rma: avoid refunding in picking return wizard

chienandalu:
- fw of https://github.com/OCA/rma/pull/344

cc @Tecnativa TT42021

ping @pedrobaeza


pedrobaeza (approved):
Fast-tracking this one:

/ocabot merge patch

Please continue with the fw-port chain

chienandalu:
This merge got stuck

PR #346, opened 2023-03-31
[14.0][IMP] RMA - Replace Operation Create Delivery

Hardik-OSI:


rma/models/rma.py
```diff
@@ -980,6 +980,13 @@ def _create_receptions_from_product(self):
             ),
             view="stock.view_picking_form",
         )
+        if self.operation_id.name == "Replace":
+            self._action_launch_stock_rule(
```
rousseldenis, line 987: @Hardik-OSI I wouldn't have based my comparison on record name. Moreover as the record can be deleted (noupdate=1).

Adding an option on rma.operation that triggers the stock rule is better IMHO.

chienandalu (changes requested):
Adding to @rousseldenis comments. This is meant to be managed manually by the RMA team. After an RMA is confirmed and despite the original intention we could end in different turnouts.

rousseldenis:
> Adding to @rousseldenis comments. This is meant to be managed manually by the RMA team. After an RMA is confirmed and despite the original intention we could end in different turnouts.

Indeed, that's why I would have added that feature as an option (checked: automatic, unchecked: manual).

PR #348, opened 2023-04-24
[13.0][FIX] rma: linked replacement moves

chienandalu:
It's useless to link the replacement moves and the side effect is that we can't make a proper reservation as the move stays in `waiting` state forever because the origin move (the reception one) is already done.

cc @Tecnativa TT42739

please review @victoralmau @pedrobaeza

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -1135,7 +1135,6 @@ class Rma(models.Model):
         moves_before = self.delivery_move_ids
         self._action_launch_stock_rule(scheduled_date, warehouse, product, qty, uom)
         new_moves = self.delivery_move_ids - moves_before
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

pedrobaeza (commented):
I think the linked move should be for the return, not for the replacement.

chienandalu:
> I think the linked move should be for the return, not for the replacement.

Yes, it is: https://github.com/OCA/rma/blob/c0ad0bf6aae4f145dbe9c147fd8fc23c4e44dbd3/rma/models/rma.py#L1097

pedrobaeza:
Please fw-port it

chienandalu:
Looks like it's already addressed upstream

PR #349, opened 2023-04-26
[15.0][FIX] rma: avoid refunding in picking return wizard

chienandalu:
fw of #345 

cc @Tecnativa TT42021

please review @pedrobaeza @victoralmau


pedrobaeza (approved):
/ocabot merge patch

PR #350, opened 2023-05-16
[14.0][IMP] RMA - refactor to use procurement run

mt-software-de:
This PR replaces all lines where a stock.picking or stock.move is created, by using only procurement group runs. 
This helps the RMA modules to easier interact with other stock modules like https://github.com/OCA/wms/tree/14.0/stock_warehouse_flow

Depends on:
~~https://github.com/OCA/stock-logistics-warehouse/pull/1708~~
~~https://github.com/OCA/rma/pull/351~~


mt-software-de:
FYI: @jbaudoux

mt-software-de:
@pedrobaeza and @chienandalu i will redo the PR https://github.com/OCA/rma/pull/334
Using this PR as an starting point.

rma/models/rma.py
```diff
@@ -1068,63 +1113,99 @@ def _get_extra_refund_line_vals(self):
...
         if "rma_return_grouping" in self.env.context:
             group_returns = self.env.context.get("rma_return_grouping")
+        return group_returns
+
+    def _key_group_rma(self):
+        self.ensure_one()
+        return (self.partner_shipping_id.id, self.company_id.id, self.warehouse_id.id)
+
+    def _group_if_needed(self):
+        if not self._delivery_should_be_grouped():
```
jbaudoux, line 1152: Can you add a docstring here?

jbaudoux (approved):
Thanks
This approach is a lot better. Routes are now properly evaluated.

mt-software-de:
@pedrobaeza and @chienandalu could you please review this PR?

rma/__manifest__.py
```diff
@@ -1,16 +1,17 @@
...
 {
     "name": "Return Merchandise Authorization Management",
     "summary": "Return Merchandise Authorization (RMA)",
-    "version": "14.0.3.2.6",
+    "version": "14.0.4.0.0",
     "development_status": "Production/Stable",
     "category": "RMA",
     "website": "https://github.com/OCA/rma",
-    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "author": "Tecnativa, MT Software, Odoo Community Association (OCA)",
```
pedrobaeza: This change doesn't qualify you for co-author.

rma/hooks.py
```diff
@@ -61,11 +64,18 @@ def create_rma_picking_types(whs):
         whs.rma_in_type_id.return_picking_type_id = whs.rma_out_type_id.id
         whs.rma_out_type_id.return_picking_type_id = whs.rma_in_type_id.id
 
+    def create_rma_routes(warehouses):
```
pedrobaeza, line 67: Please add docstrings explaining the purpose.

rma/models/rma.py
```diff
@@ -202,11 +204,15 @@ def _domain_location_id(self):
         compute="_compute_warehouse_id",
         store=True,
     )
-    reception_move_id = fields.Many2one(
-        comodel_name="stock.move",
+    reception_move_ids = fields.Many2many(
```
pedrobaeza: Why are you changing this?

rma/models/rma.py
```diff
@@ -282,18 +288,15 @@ def _domain_location_id(self):
...
-            record.delivery_picking_count = mapped_data.get(record.id, 0)
+        for rma in self:
+            rma.delivery_picking_count = len(rma.delivery_move_ids.picking_id)
+
+    @api.depends("reception_move_ids")
+    def _compute_reception_move(self):
+        for rma in self:
+            rma.reception_move_id = (
+                rma.reception_move_ids and rma.reception_move_ids.ids[0]
+            )
```
```suggestion
            rma.reception_move_id = rma.reception_move_ids[:1]
```

rma/models/rma.py
```diff
@@ -436,10 +439,11 @@ def _compute_can_be_locked(self):
 
     @api.depends("location_id")
     def _compute_warehouse_id(self):
-        for record in self.filtered("location_id"):
-            record.warehouse_id = self.env["stock.warehouse"].search(
-                [("rma_loc_id", "parent_of", record.location_id.id)], limit=1
-            )
+        for rma in self:
```
pedrobaeza: Please don't do this change that is not adding value and reduce the diff.

rma/models/rma.py
```diff
@@ -575,12 +579,15 @@ def _send_draft_email(self):
 
     def _send_confirmation_email(self):
         """Auto send notifications"""
-        for rma in self.filtered(lambda p: p.company_id.send_rma_confirmation):
+        subtype = self.env.ref("rma.mt_rma_notification")
```
pedrobaeza: The same. Useless change. I know that it seems more performant, but it's less readable and the performance penalty thanks to the cache is barely.

rma/models/stock_move.py
```diff
@@ -10,30 +11,24 @@ class StockMove(models.Model):
 
     # RMAs that were created from the delivery move
     rma_ids = fields.One2many(
-        comodel_name="rma",
-        inverse_name="move_id",
-        string="RMAs",
+        "rma",
```
pedrobaeza: Don't change this. Useless diff and doing it less explicit removing keyword arguments.

rma/models/stock_move.py
```diff
@@ -47,9 +42,9 @@ def _action_cancel(self):
         # A stock user could have no RMA permissions, so the ids wouldn't
         # be accessible due to record rules.
         cancelled_moves = self.filtered(lambda r: r.state == "cancel").sudo()
-        cancelled_moves.mapped("rma_receiver_ids").write({"state": "draft"})
-        cancelled_moves.mapped("rma_id").update_received_state()
-        cancelled_moves.mapped("rma_id").update_replaced_state()
+        cancelled_moves.rma_receiver_ids.write({"state": "draft"})
```
pedrobaeza: Please reduce diff with useless changes.

pedrobaeza (changes requested):

rma/models/rma.py
jbaudoux: I don't agree. It's cleaner and better

rma/models/stock_move.py
jbaudoux: mapped is useless and has no reason to stay

rma/models/rma.py
pedrobaeza: Explain why. Anyway, what we want is to reduce the diff, so for better chances to get this merged, this PR should touch only the significant parts. For refactorings, do another PR.

rma/models/stock_move.py
pedrobaeza: Correct, but still the same reasoning: we don't want a big diff for something unrelated of the goal of the PR. They work the same having `mapped` or not, so that's why it's useless. Do that in another PR doing a refactor if you want.

rma/__manifest__.py
mt-software-de: What are the guidelines, when to change the author?

rma/models/rma.py
mt-software-de: Because i'm changing the create of the reception_move, it is now created with procurement run.
This is not returning a move. Thats why i'm doing this https://github.com/mt-software-de/rma/blob/00c8faaae4c4fb1bda0dafd56a964fd72680d6df/rma/models/rma.py#L698, https://github.com/mt-software-de/rma/blob/00c8faaae4c4fb1bda0dafd56a964fd72680d6df/rma/models/stock_move.py#L25 and https://github.com/mt-software-de/rma/blob/00c8faaae4c4fb1bda0dafd56a964fd72680d6df/rma/models/stock_move.py#L110

rma/models/rma.py
jbaudoux: I agree, this change is not linked to the commit and PR.

Cleaner as it is looping once on self instead of twice

jbaudoux:
@mt-software-de Can you drop the changes to generated files?

mt-software-de:
> @mt-software-de Can you drop the changes to generated files?

Done. What can we do with this one?
@pedrobaeza  all of your change request are solved.

victoralmau (commented):
The code looks correct (I have to test use cases), but are the `delivery_procurement_group_carrier` and `stock_helper` dependencies really necessary (I think not)?

pedrobaeza:
No extra dependencies must be added.

rousseldenis:
> The code looks correct (I have to test use cases), but are the `delivery_procurement_group_carrier` and `stock_helper` dependencies really necessary (I think not)?

stock_helper provide extra functions in order to reuse code. I think this is a sustainable process. If not, we need to reimplement the functions here.

Sometimes, blind-blocking for such features are non sense (for me :smiley: ).

pedrobaeza:
Sorry, but the modules are right now independent and we don't want to depend on other modules. The feature is interesting, but adding such dependencies not. I don't think the amount of code to repeat is very high.

mt-software-de:
stock_helper is in fact not needed, but `delivery_procurement_group_carrier` is needed. 
Otherwise it is not possible to set the carrier via the proc group. Please check the change in
[rma_delivery/models/rma.py](https://github.com/OCA/rma/pull/350/files#diff-1f0be814effba842b213ac1e77b08a9ff3572e62096674780e9d66ed04aac04c)

rousseldenis:
> we don't want to depend on other modules

A good answer would explain why :smiley:

pedrobaeza:
> A good answer would explain why

It's not the first time I have answered you about this. And you ask us about doing glue modules for certain things, but now want the reverse.

We don't want to have to add extra dependencies in all of our customers, and more being for such an old version. Such dependencies implies more code to maintain in our codebase, possible interactions in integration tests, etc. And they are not really adding value to us. Extra glue modules can be added for such needs.

> delivery_procurement_group_carrier is needed. Otherwise it is not possible to set the carrier via the proc group.

You should add that in a glue module IMO.

mt-software-de:
> > A good answer would explain why
> 
> It's not the first time I have asked you about this. And you ask us about doing glue modules for certain things, but now want the reverse.
> 
> We don't want to have to add extra dependencies in all of our customers, and more being for such an old version. Such dependencies implies more code to maintain in our codebase, possible interactions in integration tests, etc. And they are not really adding value to us. Extra glue modules can be added for such needs.
> 
> > delivery_procurement_group_carrier is needed. Otherwise it is not possible to set the carrier via the proc group.
> 
> You should add that in a glue module IMO.

Hmmm. To make this work we need to remove the carrier_id from the proc group and set it like it was done (just with a write)
Than the glue module needs to do the following. Change the rma_delivery method which sets the carrier by write -> to not set it. And set the carrier on the proc group which will then set automatically the carrier.

I don't think that this would be a good approach. 
And also this glue module needs to

TDu (approved):
Reviewed the last commit

victoralmau (changes requested):
Finally merged in v16 https://github.com/OCA/rma/pull/392 with some changes (e.g. Change `reception_move_ids` to `reception_move_id`), I think PR should be adapted to the same changes.

mt-software-de:
> some

mt-software-de:
> Finally merged in v16 #392 with some changes (e.g. Change `reception_move_ids` to `reception_move_id`), I think PR should be adapted to the same changes.

Thats good, that there is now finally some progress.
Thanks for change reception_move_ids back to reception_move_id. That's indeed better. 

May i raise some issue i have with your approach about how pushed an merged your PR through.
By squashing your commit it is quite hard to backport the changes. And also i can not agree on all changes you made. 
Some of them i have now some you to revert and then fw port it again 16. 

Why was this PR blocked for so long and your PR was pushed through i don't think that was a convenient way of doing it.

victoralmau:
I am trying to clarify everything a bit.

This was left pending from https://github.com/OCA/rma/pull/350#pullrequestreview-2014555447 until now (otherwise this would have been taken to later versions and adapted as necessary) in order to start with the revision.

These changes were done in v15 (indicating it was a 14.0 FWP) https://github.com/OCA/rma/pull/384 keeping the commit of yours and then adding extra ones with the changes. Finally it was needed already in v16 and that's why https://github.com/OCA/rma/pull/392 was done (there everything was simplified in 1 commit adding you as co-author to avoid that the diff could be reviewed in addition to other necessary changes and reduce the methods that were not necessary).

If there are things that you think you need to change I think it is best to do it in v16 and when everything is ok continue with backports (if you need them).

mt-software-de:
> I am trying to clarify everything a bit.
> 
> This was left pending from [#350 (review)](https://github.com/OCA/rma/pull/350#pullrequestreview-2014555447) until now (otherwise this would have been taken to later versions and adapted as necessary) in order to start with the revision.
> 
> These changes were done in v15 (indicating it was a 14.0 FWP) #384 keeping the commit of yours and then adding extra ones with the changes. Finally it was needed already in v16 and that's why #392 was done (there everything was simplified in 1 commit adding you as co-author to avoid that the diff could be reviewed in addition to other necessary changes and reduce the methods that were not necessary).
> 
> If there are things that you think you need to change I think it is best to do it in v16 and when everything is ok continue with backports (if you need them).

Yes i have seen what you have done, but exactly thats the point, which i think is not convenient. 
Because no i have to check what you changed, in my commit because there are no separate commits for it. 
As well you added on the commit message me as as co-author but also im the author if the commit. https://github.com/OCA/rma/pull/392/commits/08a492ca769f768093f35f07e1552c291609c449

The annoying thing for me is right now, that i have now more worked to do. Somehow i really don't know why this one was blocked. 
I would be much easier if we would have merged this one and changed your suggestions and then fw port it then.

lmignon:
@mt-software-de IMO it could be easier to revert the commit in V16 and reapply yours. Do you plan to make a PR  for reintroducing the missing pieces in V16? If not, we can do it on our own while adding any relevant improvements from https://github.com/OCA/rma/pull/384/commits/2179cc5a1b63d2a465c0bc3160648ee5cdfb7cb9 and the test from 9bbe9973f9ee7b4791054b5303f01790c079692e

mt-software-de:
> @mt-software-de IMO it could be easier to revert the commit in V16 and reapply yours. Do you plan to make a PR for reintroducing the missing pieces in V16? If not, we can do it on our own while adding any relevant improvements from [2179cc5](https://github.com/OCA/rma/commit/2179cc5a1b63d2a465c0bc3160648ee5cdfb7cb9) and then test from [9bbe997](https://github.com/OCA/rma/commit/9bbe9973f9ee7b4791054b5303f01790c079692e)

No i am not planing to do it. I also think implementing the relevant improvements here and the revert the changes would be the easiest way.

mt-software-de:
@sbejaoui could you please review this PR.

mt-software-de:
@jbaudoux @rousseldenis re-review this? 
@lmignon @TDu could you please review this?

It would be nice to merge this finally and then FW Port it do v16. https://github.com/OCA/rma/pull/403

victoralmau (changes requested):
https://github.com/OCA/rma/pull/403

PR #351, opened 2023-05-22
[14.0][FIX] rma_sale_mrp: _check_rma_invoice_lines_qty 

mt-software-de:
checks now only lines with a phantom_bom_product set

If we are trying to post a refund for rma without kit and a lower qty than defined in the rma. 
There will no error thrown, as designed here https://github.com/OCA/rma/blob/50d86aedd57f8fcd6605768f2c9e59c76c58dd9b/rma/models/account_move.py#L30

### DIFF
--- a/rma_sale_mrp/models/account_move.py
+++ b/rma_sale_mrp/models/account_move.py
@@ -1,4 +1,5 @@
 # Copyright 2021 Tecnativa - David Vidal
+# Copyright 2024 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from odoo import models
 from odoo.tools import float_compare
@@ -16,7 +17,8 @@ class AccountMove(models.Model):
         if lines:
             return lines.sudo().filtered(
                 lambda r: (
-                    r.rma_id.phantom_bom_product
+                    not r.rma_id.phantom_bom_product
+                    or r.rma_id.phantom_bom_product
                     and float_compare(r.quantity, r.rma_id.kit_qty, precision) < 0
                 )
             )
--- a/rma_sale_mrp/readme/CONTRIBUTORS.rst
+++ b/rma_sale_mrp/readme/CONTRIBUTORS.rst
@@ -1,3 +1,4 @@
 * `Tecnativa <https://www.tecnativa.com>`__:
 
   * David Vidal
+* Michael Tietz (MT Software) <mtietz@mt-software.de>


mt-software-de:
@pedrobaeza and @chienandalu could you please do the review.

pedrobaeza:
Can you please add steps to reproduce the problem?

Also a regression test would be welcome.

mt-software-de:
> Can you please add steps to reproduce the problem?
> 
> Also a regression test would be welcome.

You can reproduce the problem if you install rma_sale_mrp with demo data and the run the tests of the base rma module

pedrobaeza:
Tests are not re-entrant by default. We have this module installed, and our users do both type of RMA without problem, so I insist in a reproducible steps to reproduce the problem.

mt-software-de:
> Tests are not re-entrant by default. We have this module installed, and our users do both type of RMA without problem, so I insist in a reproducible steps to reproduce the problem.

Ok then you can reproduce the error like this.
If you have rma_sale_mrp installed
create a rma with a qty > 1 for product which is not a kit product
Then do the incoming picking and after that create the refund
on the refund change the qty to a lower qty than it is set on the rma.
And then post the refund by calling action_post.
This should thrown an error because of this design https://github.com/OCA/rma/blob/50d86aedd57f8fcd6605768f2c9e59c76c58dd9b/rma/models/account_move.py#L30
But the error will not be thrown.

pedrobaeza:
@chienandalu do you understand the problem? I don't.

rma_sale_mrp/models/account_move.py
```diff
@@ -16,7 +16,8 @@ def _check_rma_invoice_lines_qty(self):
         if lines:
```
chienandalu, line 17: Maybe this could be redesigned:

- If there are `lines` from the super we return them directly (thus the error will be raised)
- Otherwise we filter again over `self.invoice_line_ids` with the kits conditions.

chienandalu (commented):
Yes, I think is ok @pedrobaeza 

@mt-software-de thanks for the fix :)

A comment on the method override design:

rma_sale_mrp/models/account_move.py
mt-software-de: Yes a redesign would be good. Done :-)

mt-software-de: Sorry no its not possible to return the lines because they maybe contain kit lines

mt-software-de: @chienandalu could you remove the stale label and merge this one?

mt-software-de:
@chienandalu can you update your review and remove the stale label?

rma_sale_mrp/models/account_move.py
```diff
@@ -13,11 +13,12 @@ def _check_rma_invoice_lines_qty(self):
             "Product Unit of Measure"
         )
         lines = super()._check_rma_invoice_lines_qty()
-        if lines:
-            return lines.sudo().filtered(
-                lambda r: (
-                    r.rma_id.phantom_bom_product
-                    and float_compare(r.quantity, r.rma_id.kit_qty, precision) < 0
-                )
+        if not lines:
```
pedrobaeza: Please just add the new condition instead of changing code adding an extra return. It's not adding value and makes the code more confusing.

pedrobaeza (changes requested):

rma_sale_mrp/models/account_move.py
jbaudoux: @mt-software-de Can you do this small change so that this can get approved?

mt-software-de: done

victoralmau (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #354, opened 2023-07-06
[15.0][IMP] rma_sale: Link refund line with origin sale line

chienandalu:
FW of #338 

@pedrobaeza :

Steps to reproduce:

- Create a sales order with an storable product with invoicing policy on delivered quantities.
- Confirm it and deliver the product.
- Invoice the order.
- Do an RMA, return it, and refund it.
- Result: the delivered quantity is 1 instead of 0.

This is because the refund generated from the RMA is not linked to sales order line, nor the RMA reception move. This is done because other operations are performed:

- Be replaced.
- Be changed by other product.

And we don't also want that meanwhile the RMA is being performed, the sales order is pending to invoice.

But when the refund has been done, we have it clear, so let's link both and have the sales statistics correct.

@chienandalu :

I've found that splitting the RMA leads to inconsistencies in delivered and invoiced quantities.

For example, if we receive an RMA for 5 units and return 2 to the customer and refund the remaining 3, we'll have a count of 0 delivered and 2 invoiced.

This is because once the reception is linked to the sale line we can't split those outcomes.

Given this issue, I think this should go at least behind a company setting as we discussed earlier.

@Tecnativa TT41645

### DIFF
--- a/rma_sale/__manifest__.py
+++ b/rma_sale/__manifest__.py
@@ -13,6 +13,7 @@
     "depends": ["rma", "sale_stock"],
     "data": [
         "security/ir.model.access.csv",
+        "views/account_move_views.xml",
         "views/report_rma.xml",
         "views/rma_views.xml",
         "views/sale_views.xml",
--- a/rma_sale/models/__init__.py
+++ b/rma_sale/models/__init__.py
@@ -1,4 +1,5 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from . import account_move
 from . import res_company
 from . import res_config_settings
 from . import rma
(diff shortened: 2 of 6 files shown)


pedrobaeza:
@chienandalu I have applied the definitive patch, as we don't link neither the invoice nor the move to the sales order if the quantity is not the whole move one. That way, we don't touch delivered/invoiced quantities on the sales order if there's no guarantee that the consistency is kept. Thus, no need of configuration option.

pedrobaeza (approved):

CarlosRoca13 (approved):

chienandalu:
Let's go :)

PR #356, opened 2023-07-06
15.0

caalvinz:
This will help to fix the problem: while the quantity is not an integer (for example 0.83), the RMA receipt order can't be validated. 
image


caalvinz:
Updated in the latest commit.

pedrobaeza:
Thanks for the contribution.

- Please squash all the commits into one.
- Use a proper commit message: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message
- Use `float_compare` function instead of that.

caalvinz:
Yep, I will create a new one for this. Thanks for your kind feedback.

PR #357, opened 2023-07-10
[FIX] rma: Validation failed on products receiving while quantity is not an integer value

caalvinz:
Fix float value comparison with "float_compare" instead of "=".

### DIFF
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -59,7 +59,7 @@ class StockMove(models.Model):
         """
         for move in self.filtered(lambda r: r.state not in ("done", "cancel")):
             rma_receiver = move.sudo().rma_receiver_ids
-            if rma_receiver and move.quantity_done != rma_receiver.product_uom_qty:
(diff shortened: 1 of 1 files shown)


caalvinz:
@pedrobaeza Hi, could you give me hand? I don't know how to fix the pre-commit problems. I just modified one line code for the if condition and it worked well in my local Odoo instance.

pedrobaeza:
For pre-commit, you can install and run it locally:

https://github.com/OCA/maintainer-tools/wiki/Install-pre-commit

but if you check the log, the format suggestion is put:

https://github.com/OCA/rma/actions/runs/5512483969/jobs/10049374479?pr=357#step:7:72

```diff
-            if rma_receiver and float_compare(move.quantity_done, rma_receiver.product_uom_qty, 2)!=0:
+            if (
+                rma_receiver
+                and float_compare(move.quantity_done, rma_receiver.product_uom_qty, 2)
+                != 0
+            ):
```

About the tests errors, the error is `NameError: name 'float_compare' is not defined`, because you have forgotten to import the method: `from odoo.tools import float_compare`.

caalvinz:
Yep, you are right. Thank you very much.

PR #361, opened 2023-08-24
[16.0][MIG] rma: Migration to 16.0

pedrobaeza:
* Standard procedure.
* Transfer view groups to nodes.
* Adjusted upstream changed field names.
* Converted onchanges to computed writable fields.
* Replace `Form` by direct dictionary vals in record creation, as they don't handle now properly multiple existing fields in the view, and computed writable improve the compatibility on new values.
* Replace domain returned on onchange by static domain in field.
* Change maintainer.
    
@Tecnativa TT44213


pedrobaeza:
The pre-commit error is supposed to be fixed in pylint-odoo. Probably a copier update template is needed. I also want to change refund generation without Form and remove the domain returned by the onchange.

pedrobaeza:
Everything ready to be reviewed.

rma/models/stock_picking.py
```diff
@@ -0,0 +1,43 @@
...
+    )
+
+    def _compute_rma_count(self):
+        for rec in self:
+            rec.rma_count = len(rec.move_ids.mapped("rma_ids"))
+
+    def copy(self, default=None):
+        self.ensure_one()
+        if self.env.context.get("set_rma_picking_type"):
+            location_dest_id = default["location_dest_id"]
```
chienandalu: We can't rely on this key as it isn't always present: https://github.com/odoo/odoo/blame/5f1ad0c1e4c1e8ed41337ccd4c12e70c9c760c85/addons/stock/wizard/stock_picking_return.py#L124

This provokes an error when you try to create an RMA from the picking return wizard when storage locations aren't active

chienandalu (commented):
I stumbled upon an issue when testing with no locations group:

rma/models/stock_picking.py
pedrobaeza: OK, fixed, but this may be a symptom of other bigger thing. Please try again.

rma/wizard/rma_delivery_views.xml
```diff
@@ -0,0 +1,73 @@
...
+                            <field
+                                name="product_uom_qty"
+                                attrs="{'required': [('rma_count', '=', 1)]}"
+                            />
+                            <field
+                                name="product_uom"
+                                groups="uom.group_uom"
+                                attrs="{'required': [('rma_count', '=', 1)]}"
+                                domain="[('category_id', '=', uom_category_id)]"
+                            />
```
chienandalu, line 46: Another error that I'd say is related to the other one and the way the groups are handled for the final view rendering and the attributes we get.

```suggestion
                            <field
                                name="product_uom"
                                groups="uom.group_uom"
                                attrs="{'required': [('rma_count', '=', 1)]}"
                                domain="[('category_id', '=', uom_category_id)]"
                            />
                            <field name="product_uom" groups="!uom.group_uom" invisible="1"/>
```

The error with no uom group:

```
  File "/mnt/data/odoo-addons-dir/rma/wizard/rma_delivery.py", line 82, in action_deliver
    rma.create_replace(
  File "/mnt/data/odoo-addons-dir/rma/models/rma.py", line 1218, in create_replace
    self._action_launch_stock_rule(scheduled_date, warehouse, product, qty, uom)
  File "/mnt/data/odoo-addons-dir/rma/models/rma.py", line 1303, in _action_launch_stock_rule
    self.env["procurement.group"].run([procurement])
  File "/opt/odoo/addons/stock/models/stock_rule.py", line 446, in run
    float_is_zero(procurement.product_qty, precision_rounding=procurement.product_uom.rounding)
  File "/opt/odoo/odoo/tools/float_utils.py", line 123, in float_is_zero
    epsilon = _float_check_precision(precision_digits=precision_digits,
  File "/opt/odoo/odoo/tools/float_utils.py", line 29, in _float_check_precision
    assert precision_rounding is None or precision_rounding > 0,\
AssertionError: precision_rounding must be positive, got 0.0
```

rma/models/stock_picking.py
chienandalu: > this may be a symptom of other bigger thing

Yes, looks like we don't have those values anymore when the group  removes the field. Related:
https://github.com/OCA/rma/pull/361#discussion_r1305510978

rma/wizard/rma_delivery_views.xml
pedrobaeza: Yes, that one is the key: to add the fields protected with groups as invisible for the framework to prefetch them anyway and be able to operate with them in defaults, compute, etc. Rechecked now all the affected ones. Anyway, we leave the defensive code about the existence of the dictionary key.

rma/models/stock_picking.py
pedrobaeza: I have added a new line in migration guide about this:

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0/_compare/ad52290cdb5f5385ae25fad0c4862a4fa244d9f8...e2d0ba6cd1bfdc8cee787c3fd61b6feecfbe49cc

chienandalu (approved):
:+1:

victoralmau (approved):
Don't forget to remove the migrations folder.

PR #363, opened 2023-08-25
[FIX] rma_sale: set payment_term from order

mt-software-de:


pedrobaeza (changes requested):
That were done on purpose. The payment terms don't apply to refunds, and more to RMA ones. If you need to refund the money due to a faulty product, you can't expect the customer to accept a delayed payment according your "standard conditions".

If this is a frequent thing on B2B, you should put it under an option, not as default.

mt-software-de:
> That were done on purpose. The payment terms don't apply to refunds, and more to RMA ones. If you need to refund the money due to a faulty product, you can't expect the customer to accept a delayed payment according your "standard conditions".
> 
> If this is a frequent thing on B2B, you should put it under an option, not as default.

Thx, we shouldn't do this.
This was required due to a wrong usage of payment terms. 
So i will close this PR.

PR #374, opened 2023-11-01
[14.0][IMP] rma_sale: Set analytic_account_id on refund from sale.order

mt-software-de:


rma_sale/models/rma.py
```diff
@@ -116,3 +117,4 @@ def _prepare_refund_line(self, line_form):
         if line:
             line_form.discount = line.discount
             line_form.sequence = line.sequence
+            line_form.analytic_account_id = line.order_id.analytic_account_id
```
pedrobaeza: Only assign it if there's any. If not, analytic defaults will be overwritten.

rma_sale/tests/test_rma_sale.py
```diff
@@ -91,11 +92,18 @@ def test_create_rma_from_so(self):
             {"login": "test_refund_with_so", "name": "Test"}
         )
         order.user_id = user.id
+        order.analytic_account_id = self.env.ref(
+            "analytic.analytic_our_super_product"
```
pedrobaeza: Please don't use demo data, but create your own AA.

rma_sale/tests/test_rma_sale.py
```diff
@@ -91,11 +92,18 @@ def test_create_rma_from_so(self):
...
+        ).id
         rma.action_confirm()
         rma.reception_move_id.quantity_done = rma.product_uom_qty
         rma.reception_move_id.picking_id._action_done()
         rma.action_refund()
         self.assertEqual(rma.refund_id.user_id, user)
+        self.assertEqual(
+            rma.refund_id.invoice_line_ids.analytic_account_id,
```
pedrobaeza, line 104: For tests, use always hard-coded references. Here, you are reimplementing again the same logic of what you want to test.

rma_sale/tests/test_rma_sale.py
mt-software-de: Why do you think i shouldn't use it?
I'm not a fan of doing things twice.

rma_sale/tests/test_rma_sale.py
mt-software-de: What would be the difference in doing it like
```
analytic_account = some_account
order.analytic_account_id = analytic_account.id
rma.action_refund()
self.assertEqual(rma.refund_id.invoice_line_ids.analytic_account_id, analytic_account)
```
It is basically the same .

rma_sale/tests/test_rma_sale.py
pedrobaeza: It's a guideline even from Odoo. Demo data can be altered in any sample DB (or by other modules), and then this change may provoke an error in current tests, which it shouldn't. Odoo has been doing that for several versions, but finally recognizes the problem and switch to tests with "self-contained data". I don't remember if in v15 or v16. A notable problem is to use the standard pricelist that is changed from USD to EUR (or the reverse) when installing `crm` module, and provoke a lot of problems. This has been solved in v17 not changing that data though (not creating a new pricelist for each test).

rma_sale/tests/test_rma_sale.py
pedrobaeza: Both make the test green right now, but it's not a guideline of good practice. These are the reasons:

1. Readability. In its current form, it requires to "decode" the logic behind the data that you want to check, having to read upper lines until finding the line where the value is assigned, or debugging.
2. Scope. You don't have to test the implementation, but the business requirement, which is to have the analytic account in the refund line, so this is what you have to check.
3. Avoid side effects. If the implementation changes, you are still testing the final result, which is the business requirement.
4. Resiliency. If tomorrow the field name changes, you won't need to touch in 2 places, only in the first analytic assignation.

Do you understand now why it's a good practice?

mt-software-de: In this test case i think it is more readable, because what i want to test is.
If the refund line analytic account is the same as the order analytic account.
```
self.assertEqual(
    rma.refund_id.invoice_line_ids.analytic_account_id,
    order.analytic_account_id,
)
```
With this line
```
self.assertEqual(
    rma.refund_id.invoice_line_ids.analytic_account_id,
    analytic_account,
)
```
i have to look, where analytic_account is defined. 

You are right if the field would change, i have to change it in this test. Two times, but for me this would be fine.

rma_sale/tests/test_rma_sale.py
mt-software-de: Fine for me

rma_sale/tests/test_rma_sale.py
pedrobaeza: Most IDEs have the option to go to the definition, so you can easily navigate to `analytic_account` assignation, so that's not a big problem. On contrary, using a complex expression like `order.analytic_account_id` is not traceable and may vary if other code reassigns this value. Imagine a test failure that gets you crazy because there's an installed module that is auto-assigning the analytic account in the sales order according some criteria after you have made the refund.

As said, these are good practices for avoiding risks. That risks may not come, but if so, it can make you lose a lot of time debugging and trying to figure out things.

mt-software-de: I don't think `order.analytic_account_id` is complex. 
Also i don't think we should consider other modules which will maybe change the behavior als long as the not a dependency or a auto_install module of the tested one.
Also it could be the other way around, if some module changes the account on the sale.order within the action_refund before the main code is executed. Then it its fine that initial account which was set in the test is not the one we test against.

pedrobaeza: Well, I desist trying to explain why it's convenient. I'm not going to block the PR due to this.

pedrobaeza (approved):

chienandalu (approved):
Thanks @mt-software-de :+1:

pedrobaeza:
/ocabot merge patch

Can you please fw-port it to upper versions?

pedrobaeza:
https://github.com/OCA/rma/pull/373 should be finished to this be mergeable.

rma_sale/tests/test_rma_sale.py
nilshamerlinck: a bit counter-intuitive indeed :) ftr here's the PR on odoo side for that change of approach for tests since v14: https://github.com/odoo/odoo/pull/37312

chienandalu:
Rebase to get :green_circle:  CIs

mt-software-de:
> Rebase to get 🟢 CIs

done.

PR #383, opened 2024-04-24
[16.0][FIX] rma_sale_mrp: _check_rma_invoice_lines_qty

victoralmau:
FWP from 15.0: https://github.com/OCA/rma/pull/382

Checks now only lines with a phantom_bom_product set

If we are trying to post a refund for rma without kit and a lower qty than defined in the rma.
There will no error thrown, as designed here 

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #384, opened 2024-04-26
[15.0][IMP] RMA - refactor to use procurement run

victoralmau:
FWP from 14.0: https://github.com/OCA/rma/pull/350

Changes done:
- [x] Remove `rma_delivery` changes.
- [x] Adapt changes to 15.0
- [x] Change `reception_move_ids` to `reception_move_id` field
- [x] Add tests

@Tecnativa TT48789


victoralmau:
Ping @chienandalu

victoralmau:
Ping @pedrobaeza and @chienandalu

rma/models/rma.py
```diff
@@ -650,21 +642,125 @@ def action_rma_send(self):
         }
 
     def _add_message_subscribe_partner(self):
```
pedrobaeza, line 644: This method shouldn't change. Just put `self.ensure_one()` and call it one by one.

rma/models/rma.py
```diff
@@ -650,21 +642,125 @@ def action_rma_send(self):
...
+        return product.type in ["product", "consu"]
+
+    def _prepare_procurement_group_values(self):
+        return {
+            "move_type": "direct",
+            "partner_id": self and self.partner_shipping_id.id or False,
+            "name": self and ", ".join(self.mapped("name")) or False,
+        }
+
+    def _create_procurement_group(self, values=None):
```
pedrobaeza: This method shouldn't exist, as it's not going to be overridden. Instead, all the values should be handled in `_prepare_procurement_group_values` instead of that weird thing handling some of them in one method, and the rest in the other.

rma/models/rma.py
```diff
@@ -650,21 +642,125 @@ def action_rma_send(self):
...
+        return {
+            "company_id": self.company_id,
+            "group_id": group,
+            "date_planned": scheduled_date or fields.Datetime.now(),
+            "warehouse_id": warehouse or self.warehouse_id,
+            "partner_id": group.partner_id.id,
+            "priority": self.priority,
+        }
+
+    def _prepare_procurement(
```
pedrobaeza: The same about excessive hooks. I don't think they are needed.

rma/models/rma.py
```diff
@@ -650,21 +642,125 @@ def action_rma_send(self):
...
-                reception_move = self._create_receptions_from_picking()
-            else:
-                reception_move = self._create_receptions_from_product()
-            self.write({"reception_move_id": reception_move.id, "state": "confirmed"})
-            self._add_message_subscribe_partner()
-            self._send_confirmation_email()
+        self = self.filtered(lambda rma: rma.state == "draft")
+        if not self:
+            return
+        self._create_receptions()
```
pedrobaeza, line 734: I don't think we should extract this method too. When a real example requires it, then it can be extracted. Meanwhile, the code gets very complicated to be followed with all of this and with no real gain.

rma/models/rma.py
```diff
@@ -815,8 +911,7 @@ def action_preview(self):
             "url": self.get_portal_url(),
         }
 
-    def action_view_receipt(self):
-        """Invoked when 'Receipt' smart button in rma form view is clicked."""
+    def action_view_pickings(self, pickings):
```
pedrobaeza: A public method shouldn't require an extra argument. It should be private.

rma/models/rma.py
```diff
@@ -1233,86 +1331,27 @@ def create_replace(self, scheduled_date, warehouse, product, qty, uom):
...
-        rma.create_replace
-        """
-        self.ensure_one()
-        if self.product_id.type not in ("consu", "product"):
-            return
-        if not self.procurement_group_id:
-            self.procurement_group_id = (
-                self.env["procurement.group"]
-                .create(
+        for rma in self:
```
pedrobaeza, line 1300: Don't convert everything to multi-record to get still more diff. Let it `self.ensure_one()` and call it one by one.

pedrobaeza (commented):
I know that most of the changes are not yours, but the diff is barely impossible to follow due to unneeded changes like passing methods from `self.ensure_one` to multi-record with no reason, and excessive unneeded hooks. Please clean up for being able to check something.

rma/models/rma.py
victoralmau: This method was already extracted previously (_create_receptions_from_picking and _create_receptions_from_product).

victoralmau:
Changes done to reduce diff and methods.

victoralmau:
Superseed https://github.com/OCA/rma/pull/392

PR #392, opened 2024-05-29
[16.0][IMP] RMA - refactor to use procurement run

victoralmau:
FWP from 15.0: https://github.com/OCA/rma/pull/384

Extra changes:
- [x] Change reception_move_ids to reception_move_id
- [x] Add test_rma_replace_pick_ship
- [x] Code and method reduction to simplify logic
- [x] Set route_ids (in/out) from procurements

@Tecnativa TT48789


rma/migrations/16.0.1.4.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
...
+
+from openupgradelib import openupgrade
+
+# pylint: disable=W8150
+from odoo.addons.rma.hooks import post_init_hook
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    post_init_hook(env.cr, env.registry)
```
pedrobaeza: This should call only `create_rma_routes`

rma/migrations/16.0.1.4.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2024 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+# pylint: disable=W8150
+from odoo.addons.rma.hooks import post_init_hook
```
pedrobaeza:
```suggestion
from ...hooks import create_rma_routes
```

rma/models/rma.py
```diff
@@ -670,21 +661,94 @@ def action_rma_send(self):
...
+        return product.type in ["product", "consu"]
+
+    def _prepare_procurement_group_values(self):
+        return {
+            "move_type": "direct",
+            "partner_id": self and self.partner_shipping_id.id or False,
+            "name": self and ", ".join(self.mapped("name")) or False,
+        }
+
+    def _prepare_procurement_values(
```
pedrobaeza: For better understanding and use the usual suffix as well (applicable this second part to all the code):

```suggestion
    def _prepare_common_procurement_vals(
```

rma/models/rma.py
```diff
@@ -670,21 +661,94 @@ def action_rma_send(self):
...
+        group_model = self.env["procurement.group"]
+        for rma in self:
+            if not rma._product_is_storable():
+                continue
+            group = rma.procurement_group_id
+            if not group:
+                group = group_model.create(rma._prepare_procurement_group_values())
+            product = rma.product_id
+            values = rma._prepare_reception_procurement_values(group)
+            values.update({"route_ids": rma.warehouse_id.rma_in_route_id})
```
pedrobaeza: Do this inside `_prepare_reception_procurement_vals`

rma/models/rma.py
```diff
@@ -670,21 +661,94 @@ def action_rma_send(self):
...
+        procurements = []
+        group_model = self.env["procurement.group"]
+        for rma in self:
+            if not rma._product_is_storable():
+                continue
+            group = rma.procurement_group_id
+            if not group:
+                group = group_model.create(rma._prepare_procurement_group_values())
+            product = rma.product_id
+            values = rma._prepare_reception_procurement_values(group)
```
pedrobaeza: `vals` is the usual variable name. Better to not change it (the same to the rest of the code).

rma/models/rma.py
```diff
@@ -670,21 +661,94 @@ def action_rma_send(self):
...
+            "priority": self.priority,
+        }
+
+    def _prepare_reception_procurement_values(self, group=None):
+        values = self._prepare_procurement_values(group=group)
+        values.update(
+            {
+                "rma_receiver_ids": [(6, 0, self.ids)],
+            }
+        )
```
pedrobaeza: Better to not use `update` as it's uglier:

```suggestion
        vals["rma_receiver_ids"] = Command.set(self.ids)
```

rma/models/rma.py
```diff
@@ -670,21 +661,94 @@ def action_rma_send(self):
...
+        if self.move_id:
+            values.update(
+                {
+                    "origin_returned_move_id": self.move_id.id,
+                }
+            )
```
```suggestion
            vals["origin_returned_move_id"] = self.move_id.id
```

rma/models/rma.py
```diff
@@ -670,21 +661,94 @@ def action_rma_send(self):
...
-            self.write({"reception_move_id": reception_move.id, "state": "confirmed"})
-            self._add_message_subscribe_partner()
-            self._send_confirmation_email()
+        self = self.filtered(lambda rma: rma.state == "draft")
+        if not self:
+            return
+        procurements = self._prepare_reception_procurements()
+        if procurements:
+            self.env["procurement.group"].run(procurements)
+        self.reception_move_id.picking_id.action_assign()
```
pedrobaeza, line 738: Where is `reception_move_id` written?

rma/models/rma.py
```diff
@@ -819,25 +871,28 @@ def action_preview(self):
...
+                res_id=pickings.id,
+                view_mode="form",
+                view_id=False,
+                views=False,
+            )
         return action
 
+    def action_view_receipt(self):
+        """Invoked when 'Receipt' smart button in rma form view is clicked."""
+        return self._action_view_pickings(self.reception_move_id.picking_id)
```
pedrobaeza: Shouldn't this cover the possibility of having the RMA receipt in several steps?

rma/models/stock_move.py
```diff
@@ -22,7 +23,7 @@ class StockMove(models.Model):
         string="RMA receivers",
         copy=False,
     )
-    # RMA that create the delivery movement to the customer
+    # RMA that create the out move
```
pedrobaeza:
```suggestion
    # RMA that creates the out move
```

rma/readme/CONTRIBUTORS.rst
```diff
@@ -7,3 +7,4 @@
 
 * Chafique Delli <chafique.delli@akretion.com>
 * Giovanni Serra - Ooops <giovanni@ooops404.com>
+* Michael Tietz (MT Software) <mtietz@mt-software.de>
```
pedrobaeza, line 10: You should be also in the contributors for the adaptation work.

pedrobaeza (changes requested):
You have to extend my comments to the whole code.

rma/readme/CONTRIBUTORS.rst
victoralmau: I agree, but I am already added as a contributor.

rma/models/rma.py
victoralmau: This is done with the `rma_receiver_ids` key within _values_ in the procurement

rma/models/rma.py
victoralmau: Already changed. Clarify that to achieve multi-step RMA receipt it would be necessary to set the specific rules in the RMA: In route.  Is this approach “_appropriate_” or does it need to be made compatible with 2-step (or 3) incoming? 


If it has to be made compatible there would be several things to clarify beforehand:
- Do we use the already existing _reception_steps_ field of the warehouse or do we create a new one (`rma_reception_steps`). I think it should be a new one because you may want to have several steps active on one side and not the other.
- Do we auto-create the RMA: IN path according to the above field and auto-create the intermediate locations to get it?

The same will happen for customer returns with the RMA: Out rule.

pedrobaeza: We are not adding this for the moment. It's just that this method to be prepared when that happens (if ever). Add in ROADMAP the sentence "Add RMA reception and/or RMA delivery on several steps - 2 or 3 - like normal receptions/deliveries. It should be a separate option inside the warehouse definition.".

That's because you can have different flows for regular moves and RMA ones.

pedrobaeza (approved):

chienandalu (approved):
Nice :) Thanks!

PR #393, opened 2024-05-30
[14.0][ADD] rma_employee

FrankC013:
Creation of a new module to add the employee field within the RMA model


FrankC013:
@eantones Please review!

eantones (approved):

pedrobaeza:
Which is the goal of using employees instead of users? Isn't this module a bit stripped without no extra features except the field?

FrankC013:
The objective of this module is to allow the user to assign an RMA to an employee. It's a purely informational field. It's created in a separate module to avoid the installation of HR when someone installs the RMA.

pedrobaeza:
But I insist: why not assigning a user instead?

PR #397, opened 2024-06-25
[17.0][IMP] RMA - refactor to use procurement run

victoralmau:
FWP from 16.0: https://github.com/OCA/rma/pull/402 + https://github.com/OCA/rma/pull/392

Extra changes:
- Change reception_move_ids to reception_move_id
- Add test_rma_replace_pick_ship
- Code and method reduction to simplify logic
- Set route_ids (in/out) from procurements

@Tecnativa TT48789

### DIFF
--- a/rma/README.rst
+++ b/rma/README.rst
@@ -140,6 +140,9 @@ Known issues / Roadmap
 -  As soon as the picking is selected, the user should select the move,
    but perhaps stock.move \_rec_name could be improved to better show
    what the product of that move is.
+-  Add RMA reception and/or RMA delivery on several steps - 2 or 3 -
+   like normal receptions/deliveries. It should be a separate option
+   inside the warehouse definition.
 
 Bug Tracker
 ===========
@@ -174,6 +177,8 @@ Contributors
 -  `APSL-Nagarro <https://www.apsl.tech>`__:
 
    -  Antoni Marroig <amarroig@apsl.net>
+
+-  Michael Tietz (MT Software) mtietz@mt-software.de
 
 Maintainers
 -----------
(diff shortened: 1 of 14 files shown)


rma/models/stock_picking.py
```diff
@@ -30,8 +30,8 @@ def copy(self, default=None):
 
     def action_view_rma(self):
         self.ensure_one()
-        action = self.sudo().env.ref("rma.rma_action").read()[0]
-        rma = self.move_ids.mapped("rma_ids")
+        action = self.env["ir.actions.act_window"]._for_xml_id("rma.rma_action")
+        rma = self.move_lines.mapped("rma_ids")
```
chienandalu:
```suggestion
        rma = self.move_ids.mapped("rma_ids")
```

chienandalu (changes requested):

rma/models/stock_picking.py
pedrobaeza: In fact:

```suggestion
        rma = self.move_ids.rma_ids
```

chienandalu (approved):
:+1:

pedrobaeza:
Testing in the runboat http://oca-rma-17-0-pr397-605691345d56.runboat.odoo-community.org to return a delivery order marking the RMA option, I get:

imagen

victoralmau:
Reviewing the error you indicate, I see that it is a side effect of something that already happens currently in v17 and v16, if the user does NOT have the locations permission, the picking that is created does NOT have the correct data (location and picking_type_id).

This does not happen in tests because the user that is used does have this permission, I will review it to fix it.

pedrobaeza (approved):
/ocabot merge major

PR #407, opened 2024-07-12
[16.0][IMP] rma_sale: Display the allowed quantity to return in the RMA wizard

### LINKED ISSUE #406: RMA SO wizard quantity default to 0 and show max
When creating an RMA from an SO, it currently opens a wizard with all lines of the SO with the quantity preset to the SO delivered quantity (in fact returnable delivered quantity).

I have several customers that are managing large SOs and facing an RMA for one or few lines. This requires to delete all extra lines or set the quantity to 0 which is cumbersome.
In this case, what would be convenient is to have the quantity by default to 0 and have another column showing the max quantity that can be returned.

I can understand this change of approach may not be convenient to current users. Should this change be implemented in another module like `rma_sale_show_max_qty` or should this be a configuration option in the base `rma_sale` module?

sbejaoui:
fixes: #406 

- Display the allowed quantity to return
- Simplify data encoding by adding a button to select/deselect all lines
- Ensure that the quantity to return is always less than or equal to the allowed quantity

image

### DIFF
--- a/rma_sale/README.rst
+++ b/rma_sale/README.rst
@@ -106,6 +106,8 @@ Contributors
 
 * Chafique Delli <chafique.delli@akretion.com>
 * Giovanni Serra - Ooops <giovanni@ooops404.com>
+* Souheil Bejaoui - ACSONE SA/NV <souheil.bejaoui@acsone.eu>
+* Jacques-Etienne Baudoux - BCIM <je@bcim.be>
 
 Maintainers
 ~~~~~~~~~~~
--- a/rma_sale/controllers/sale_portal.py
+++ b/rma_sale/controllers/sale_portal.py
@@ -25,6 +25,7 @@ class CustomerPortal(CustomerPortal):
         except (AccessError, MissingError):
             return request.redirect("/my")
         order_obj = request.env["sale.order"]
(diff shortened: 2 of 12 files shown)


lmignon (approved):
LGTM (code review + functional tests)

jbaudoux (commented):
Can you perform the same in the stock picking return ?

jbaudoux (changes requested):
If you change the product in the wizard line, the allowed quantity is not recomputed.
As the check on the wizard validation is on the allowed quantity of the lines, it is not ensuring properly the constraint (as you may have changed the product).

If you add a line in the wizard and select a product, the allowed quantity is not recomputed.

jbaudoux (approved):
Approving the improvement on the SO rma wizard.
My previous remark on changing prepopulated quantity fields to computed fields would require a refactor that is better suited in another PR as this issue about wizard line removal and addition, or product change, is already there and not due to this PR

pedrobaeza:
Can you check the red CI?

sbejaoui:
> Can you check the red CI?

it's green now

rousseldenis (approved):

PR #409, opened 2024-07-15
[16.0][ADD] add rma_reason & sale_rma_reason

### LINKED ISSUE #405: RMA reason
### Add rma reason

To improve quality, it is good to understand why an RMA has been created.

For this I'd like to have a required RMA reason defined on the RMA. This would be also available and required on the SO wizard on each line (or globally)

Make a `rma_reason` module that depends on `rma`
- [ ] Define a model `rma.reason` with name, action, company + unique(name, company). Security: read for rma user, crud for rma manager 
- [ ] Allow to configure on the company if the reason is mandatory and check accordingly if the field is filled in `_ensure_required_fields` 
- [ ] Add a m2o field to reason on rma model and display on form view after `Requested operation`

Make a `rma_sale_reason` module that depends on `rma_sale` and `rma_reason`
- [ ] In the SO create rma wizard, add on each product line a field (optional or required based on company setting) to set the reason for the lines.
- [ ] In the SO create rma wizard, add before the list of products an optional field to set the default reason for all the lines. Make on all product lines the reason computed stored based on that default reason

sbejaoui:
Specifying the RMA reason when creating an RMA order is crucial for effective inventory management, quality control, and customer service. It helps categorize returned items accurately, identify recurring product issues, and improve overall product quality. This data enables customer service to address issues more effectively and enhances the customer experience.

This addon allows RMA managers to pre-configure possible reasons users can specify when creating the RMA order.

fixes #405 

needs: #410 

cc/ @jbaudoux , @lmignon , @rousseldenis 

image

image


rma_reason/models/rma_reason.py
```diff
@@ -0,0 +1,21 @@
...
+from odoo import fields, models
+
+
+class RmaReason(models.Model):
+
+    _name = "rma.reason"
+    _description = "Rma Reason"
+
+    name = fields.Char(required=True, translate=True)
+    description = fields.Text()
```
lmignon: Why name is `translatable` but not the `description`? Both should be translatable.

lmignon (changes requested):

rousseldenis (commented):
LGTM. Apart the @lmignon remark, it should be great to add it to portal template too:

image


Thanks

rousseldenis (approved):
LGTM

lmignon (approved):

rma_reason/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2024 Raumschmiede GmbH
+# Copyright 2024 BCIM
+# Copyright 2024 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+{
+    "name": "Rma Reason",
+    "version": "16.0.1.0.0",
+    "license": "AGPL-3",
+    "author": "Raumschmiede GmbH,BCIM,ACSONE SA/NV,Odoo Community Association (OCA)",
```
pedrobaeza, line 10: I think `Raumschmiede GmbH` should go in the CREDITS file, not as author, being the paying customer, isn't it?

rma_sale/views/sale_portal_template.xml
```diff
@@ -92,7 +92,10 @@
                             <th class="text-start">Product</th>
                             <th class="text-end">Quantity</th>
                             <th class="text-start">Delivery</th>
-                            <th class="text-start">Requested operation</th>
+                            <th
+                                class="text-start"
+                                name="operation"
```
pedrobaeza, line 97: And you can take the occasion to name all the elements.

```suggestion
                                name="th_operation"
```

rma_sale/views/sale_portal_template.xml
```diff
@@ -147,7 +150,7 @@
                                             t-att-value="data['picking'] and data['picking'].id"
                                         />
                                     </td>
-                                    <td class="text-start">
+                                    <td class="text-start" name="operation">
```
pedrobaeza, line 153:
```suggestion
                                    <td class="text-start" name="td_operation">
```

rma_reason/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,2 @@
+- Jacques-Etienne Baudoux - BCIM <je@bcim.be>
+- Souheil Bejaoui - ACSONE SA/NV <souheil.bejaoui@acsone.eu.com>
```
rousseldenis:
```suggestion
- Souheil Bejaoui - ACSONE SA/NV <souheil.bejaoui@acsone.eu>
```

@sbejaoui

pedrobaeza: For reference, I avoid to put the emails on the CONTRIBUTORS list for avoiding direct requests and spam. The email is already in the commit.

rma_reason/__manifest__.py
sbejaoui: @jbaudoux , fine with you?

rma_sale/views/sale_portal_template.xml
sbejaoui: is it really needed? As in the xpath, we must already specify the node

`<xpath expr="//th[@name='operation']" position="before">`

pedrobaeza: Adding the same name several times is not a good practice IMO. It's true that selectors are done usually with the HTML tag, but to get an absolute element, better with unique names.

lmignon: You are right @pedrobaeza in the case of the `id` element attribute since this one is expected to be unique across the full dom. But for the `name`, it seems to me to be redundant.

pedrobaeza: OK, it's not critical.

pedrobaeza: Just as reference, Odoo follows my same suggestion:

https://github.com/odoo/odoo/blob/9919961ffbc9d05d87f068ee8512728345d17bf1/addons/sale/report/ir_actions_report_templates.xml#L73

rousseldenis: I would say it should be at developer's discretion as both approaches provide a unique way of xpath anchor.

rma_reason/__manifest__.py
jbaudoux: They are the original author of the module, not the paying customer of this OCA port

rma_reason/readme/CONTRIBUTORS.md
jbaudoux: You can keep mine, I always put it

rma_reason/__manifest__.py
jbaudoux: But you should not add them at the top of the file that were not there originally.

jbaudoux:
cc @raumschmiede-joshuaL

jbaudoux (approved):
LG, not tested yet

jbaudoux:
@sbejaoui I would prefer to have the reason before the operation on the wizard as, in my opinion, it makes more sense to first choose a reason and then an operation. Some operations may not be valid for some reason

jbaudoux (commented):
Can you add the reason also in the stock picking return ? Like you did on SO wizard.

sbejaoui:
> Can you add the reason also in the stock picking return ? Like you did on SO wizard. You can also create RMAs from there



> Can you add the reason also in the stock picking return ? Like you did on SO wizard. You can also create RMAs from there

done

image

rma_reason/wizards/stock_return_picking.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="model">stock.return.picking</field>
+        <field name="inherit_id" ref="stock.view_stock_return_picking_form" />
+        <field name="arch" type="xml">
+            <xpath
+                expr="//field[@name='product_return_moves']//tree//field[@name='rma_operation_id']"
+                position="before"
+            >
+                <field
+                    name="rma_reason_id"
+                    attrs="{'column_invisible': [('parent.create_rma', '=', False)], 'required': [('is_rma_reason_required', '=', True)]}"
```
jbaudoux: Like you did for the operation
```suggestion
                    attrs="{'column_invisible': [('parent.create_rma', '=', False)], 'required': [('is_rma_reason_required', '=', True), ('parent.create_rma', '=', False), ('quantity', '>', 0)]}"
```

rma_sale_reason/wizards/sale_order_rma_wizard.xml
```diff
@@ -0,0 +1,30 @@
...
+                <field name="reason_id" />
+            </xpath>
+            <xpath
+                expr="//field[@name='line_ids']//field[@name='operation_id']"
+                position="before"
+            >
+                <field name="is_rma_reason_required" invisible="True" />
+                <field
+                    name="reason_id"
+                    attrs="{'required': [('is_rma_reason_required', '=', True)]}"
```
jbaudoux:
```suggestion
                    attrs="{'required': [('is_rma_reason_required', '=', True), ('quantity', '>', 0)]}"
```

rma_sale_reason/wizards/sale_order_rma_wizard.xml
```diff
@@ -0,0 +1,30 @@
...
+                />
+            </xpath>
+        </field>
+    </record>
+
+
+
```
jbaudoux:
```suggestion

```

jbaudoux (changes requested):

sbejaoui:
@jbaudoux , can you plz update your review

jbaudoux (approved):
Tested, looks good

lmignon (approved):
Code + tested. Works in production for months

PR #410, opened 2024-07-22
[16.0][IMP] rma_sale: operation required

### LINKED ISSUE #404: RMA - require operation
### Make the rma operation required

As multiple rma operations (return, refund, replace...) are available, I find it important that the user explicitly choose the operation for each product

In rma:
- [ ] Keep the field optional on the rma but require a value when confirming the rma (by adding to _ensure_required_fields) as this dictates what to do on confirmation

In rma_sale:
- [ ] In the SO create rma wizard, add before the list of products an optional field to set the default operation for all the lines. Make on all product lines the operation computed stored based on that default operation
- [ ] In the SO create rma wizard, on each product line, make the field required if qty > 0 as this wizard confirms the rma which require an operation to be set

sbejaoui:
The rma operation is required in the portal, and it makes sense for it to be required in the sales wizard as well

fixes: #404 

### DIFF
--- a/rma/static/description/index.html
+++ b/rma/static/description/index.html
@@ -522,6 +522,8 @@ If you spotted it first, help us to smash it by providing a detailed and welcome
 <li>Chafique Delli &lt;<a class="reference external" href="mailto:chafique.delli&#64;akretion.com">chafique.delli&#64;akretion.com</a>&gt;</li>
 <li>Giovanni Serra - Ooops &lt;<a class="reference external" href="mailto:giovanni&#64;ooops404.com">giovanni&#64;ooops404.com</a>&gt;</li>
 <li>Michael Tietz (MT Software) &lt;<a class="reference external" href="mailto:mtietz&#64;mt-software.de">mtietz&#64;mt-software.de</a>&gt;</li>
+<li>Jacques-Etienne Baudoux - BCIM &lt;<a class="reference external" href="mailto:je&#64;bcim.be">je&#64;bcim.be</a>&gt;</li>
+<li>Souheil Bejaoui - ACSONE SA/NV &lt;<a class="reference external" href="mailto:souheil.bejaoui&#64;acsone.eu">souheil.bejaoui&#64;acsone.eu</a>&gt;</li>
 </ul>
 </div>
 <div class="section" id="maintainers">
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -79,8 +79,11 @@ class TestRma(TransactionCase):
         cls.warehouse = cls.env.ref("stock.warehouse0")
         # Ensure grouping
         cls.env.company.rma_return_grouping = True
+        cls.operation = cls.env.ref("rma.rma_operation_replace")
 
-    def _create_rma(self, partner=None, product=None, qty=None, location=None):
+    def _create_rma(
+        self, partner=None, product=None, qty=None, location=None, operation=None
+    ):
         vals = {}
         if partner:
             vals["partner_id"] = partner.id
@@ -90,12 +93,16 @@ class TestRma(TransactionCase):
             vals["product_uom_qty"] = qty
         if location:
             vals["location_id"] = location.id
+        if operation:
+            vals["operation_id"] = operation.id
+        elif operation is None:
+            vals["operation_id"] = self.operation.id
         return self.env["rma"].create(vals)
 
     def _create_confirm_receive(
-        self, partner=None, product=None, qty=None, location=None
+        self, partner=None, product=None, qty=None, location=None, operation=None
     ):
-        rma = self._create_rma(partner, product, qty, location)
+        rma = self._create_rma(partner, product, qty, location, operation)
         rma.action_confirm()
         rma.reception_move_id.quantity_done = rma.product_uom_qty
         rma.reception_move_id.picking_id._action_done()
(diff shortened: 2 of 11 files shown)


rousseldenis (approved):
Code review. LGTM

rma_sale/wizard/sale_order_rma_wizard_views.xml
```diff
@@ -12,6 +12,7 @@
         <field name="arch" type="xml">
             <form>
                 <group>
+                    <field name="operation_id" />
```
pedrobaeza, line 15: Why putting it twice?

rma/tests/test_rma.py
```diff
@@ -157,7 +162,9 @@ def test_post_init_hook(self):
 
     def test_rma_replace_pick_ship(self):
         self.warehouse.write({"delivery_steps": "pick_ship"})
-        rma = self._create_rma(self.partner, self.product, 1, self.rma_loc)
+        rma = self._create_rma(
+            self.partner, self.product, 1, self.rma_loc, self.operation
```
pedrobaeza: Can we avoid to pass on each call the operation and change the method `_create_rma` to take `self.operation` if the value is `None`? It's for reducing all the diff in the test that makes it difficult to follow.

rma/wizard/stock_picking_return_views.xml
```diff
@@ -14,6 +14,10 @@
                         name="create_rma"
                         attrs="{'invisible': [('picking_type_code', '!=', 'outgoing')]}"
                     />
+                    <field
+                        name="rma_operation_id"
+                        attrs="{'invisible': [('create_rma', '=', False)], 'required': [('create_rma', '=', True)]}"
```
pedrobaeza: Not sure we should require it here.

rma_sale/wizard/sale_order_rma_wizard_views.xml
jbaudoux: It's the default operation that will apply to all lines. Below is the operation for the line

pedrobaeza: OK, understood.

rma/static/description/index.html
```diff
@@ -522,6 +522,8 @@ <h2><a class="toc-backref" href="#toc-entry-7">Contributors</a></h2>
 <li>Chafique Delli &lt;<a class="reference external" href="mailto:chafique.delli&#64;akretion.com">chafique.delli&#64;akretion.com</a>&gt;</li>
 <li>Giovanni Serra - Ooops &lt;<a class="reference external" href="mailto:giovanni&#64;ooops404.com">giovanni&#64;ooops404.com</a>&gt;</li>
 <li>Michael Tietz (MT Software) &lt;<a class="reference external" href="mailto:mtietz&#64;mt-software.de">mtietz&#64;mt-software.de</a>&gt;</li>
+<li>Jacques-Etienne Baudoux - BCIM &lt;<a class="reference external" href="mailto:je&#64;bcim.be">je&#64;bcim.be</a>&gt;</li>
+<li>Souheil Bejaoui - ACSONE SA/NV &lt;<a class="reference external" href="mailto:souheil.bejaoui&#64;acsone.eu">souheil.bejaoui&#64;acsone.eu</a>&gt;</li>
```
jbaudoux, line 526: Can you skip the update of auto generated files?
You can use `export SKIP=oca-gen-addon-readme`

rma/wizard/stock_picking_return_views.xml
sbejaoui: The RMA created from the return wizard is confirmed right away. The user will get an error if they don't fill in the operation. So, I think it's more user-friendly to make the field required in the view

rma/tests/test_rma.py
sbejaoui: Make sense

rma/static/description/index.html
sbejaoui: Done

rma/tests/test_rma.py
sbejaoui: Done

pedrobaeza (approved):

jbaudoux (approved):
LG, Thanks

jbaudoux (commented):
Can you perform the same in the stock picking return ? And add the operation in the lines.

sbejaoui:
> Can you perform the same in the stock picking return ? And add the operation in the lines. You can also create RMAs from there

done

image

sbejaoui:
@pedrobaeza , ready for merge

rma/wizard/stock_picking_return_views.xml
```diff
@@ -8,12 +8,22 @@
         <field name="model">stock.return.picking</field>
         <field name="inherit_id" ref="stock.view_stock_return_picking_form" />
         <field name="arch" type="xml">
-            <field name="product_return_moves" position="after">
+            <xpath expr="//field[@name='product_return_moves']//tree" position="inside">
+                <field
+                    name="rma_operation_id"
+                    attrs="{'column_invisible': [('parent.create_rma', '=', False)], 'required': [('parent.create_rma', '=', True)]}"
```
jbaudoux: Like you did on sales, rma operation is only required if you return that line
```suggestion
                    attrs="{'column_invisible': [('parent.create_rma', '=', False)], 'required': [('parent.create_rma', '=', True), ('quantity', '>', 0)]}"
```

jbaudoux (changes requested):
Picking return wizard to fix

pedrobaeza:
Please squash commits together. They don't add value as separated. And remember you can do `git commit --amend` and `git push -f`, as GitHub will put the diff of such forced push, so you avoid extra CI builds for nothing.

sbejaoui:
@jbaudoux , could you please update your review? We'd like to merge this PR since many others depend on it.

jbaudoux (approved):

PR #415, opened 2024-07-31
[16.0][FIX] rma_sale: Fix problems with link between RMA-sale.order and procurement.group-sale.order when creating from stock.picking.return wizard

TheClaud99:
## Module
rma_sale

## Describe the bug
There's some bugs in the rma_sale module when creating a return from the stock.picking.return wizard with the create_rma flag set.

## To Reproduce
**Affected versions:**

1. Create a sale.order with a picking
2. Validate the picking
3. Create a return with the create_rma flag set.

**Expected behavior**
The order_id on the RMA and sale_id on the new procurement.group are set

### DIFF
--- a/rma/wizard/stock_picking_return.py
+++ b/rma/wizard/stock_picking_return.py
@@ -80,10 +80,6 @@ class ReturnPicking(models.TransientModel):
     def _prepare_rma_vals(self):
         partner, partner_invoice, partner_shipping = self._prepare_rma_partner_values()
         origin = self.picking_id.name
-        vals = self.env["rma"]._prepare_procurement_group_vals()
-        vals["partner_id"] = partner_shipping.id
-        vals["name"] = origin
-        group = self.env["procurement.group"].create(vals)
         return {
             "user_id": self.env.user.id,
             "partner_id": partner.id,
@@ -92,7 +88,6 @@ class ReturnPicking(models.TransientModel):
             "origin": origin,
             "picking_id": self.picking_id.id,
             "company_id": self.company_id.id,
-            "procurement_group_id": group.id,
         }
 
     def _prepare_rma_vals_list(self):
@@ -129,6 +124,10 @@ class ReturnPicking(models.TransientModel):
                 )
             vals_list = self._prepare_rma_vals_list()
             rmas = self.env["rma"].create(vals_list)
+            proc_group_vals = rmas[:1]._prepare_procurement_group_vals()
+            proc_group_vals["name"] = self.picking_id.name
+            proc_group = self.env["procurement.group"].create(proc_group_vals)
+            rmas.write({"procurement_group_id": proc_group.id})
             rmas.action_confirm()
             picking = rmas.reception_move_id.picking_id
             picking = picking and picking[0] or picking
--- a/rma_sale/wizard/stock_picking_return.py
+++ b/rma_sale/wizard/stock_picking_return.py
@@ -17,8 +17,8 @@ class ReturnPicking(models.TransientModel):
             sale_order.partner_shipping_id,
         )
 
-    def _prepare_rma_values(self):
-        vals = super()._prepare_rma_values()
+    def _prepare_rma_vals(self):
+        vals = super()._prepare_rma_vals()
         sale_order = self.picking_id.sale_id
         if sale_order:
             vals.update(


pedrobaeza:
I don't get why you need this. Can you please explain the goal of having these fields set?

TheClaud99:
The field order_id is used for example as the inverse field in the rma_ids on sale.order, used for logic like the override of the function _get_invoiced. If you see my first commit 9787b8acbe0be5312a3bdce8bf17c05b5587d484 there is a bug with a function name that causes it to never be called when it should set the order_id.

The field sale_id on the procurement.group is used for computing the field picking_ids on sale.order and in the actual version should be set by the function _prepare_procurement_group_vals in the module rma_sale but actually it never sets it because it checks for self.order_id but the function is called with no records in self.

rma/wizard/stock_picking_return.py
```diff
@@ -77,12 +77,17 @@ def _prepare_rma_partner_values(self):
             partner_shipping_id and partner.browse(partner_shipping_id) or partner,
         )
 
+    def _prepare_procurement_group_vals(self):
```
pedrobaeza: No need of this extra method

rma/wizard/stock_picking_return.py
```diff
@@ -77,12 +77,17 @@ def _prepare_rma_partner_values(self):
...
+        return vals
+
     def _prepare_rma_vals(self):
         partner, partner_invoice, partner_shipping = self._prepare_rma_partner_values()
         origin = self.picking_id.name
-        vals = self.env["rma"]._prepare_procurement_group_vals()
+        vals = self._prepare_procurement_group_vals()
         vals["partner_id"] = partner_shipping.id
-        vals["name"] = origin
+
```
pedrobaeza: Don't add empty lines inside a method.

rma_sale/wizard/stock_picking_return.py
```diff
@@ -17,8 +17,19 @@ def _prepare_rma_partner_values(self):
             sale_order.partner_shipping_id,
         )
 
-    def _prepare_rma_values(self):
-        vals = super()._prepare_rma_values()
+    def _prepare_procurement_group_vals(self):
```
pedrobaeza: I don't see this method is adding too much value.

rma_sale/wizard/stock_picking_return.py
```diff
@@ -17,8 +17,19 @@ def _prepare_rma_partner_values(self):
...
+        vals = super()._prepare_procurement_group_vals()
+        sale_order = self.picking_id.sale_id
+        if sale_order:
+            vals.update(
+                {
+                    "sale_id": sale_order.id,
+                }
+            )
```
pedrobaeza: More readable:

```suggestion
            vals["sale_id"] = sale_order.id
```

pedrobaeza (changes requested):
OK, for the first commit. The second one seems weird to me and I don't get your explanation. If you see, the method is not covered by tests, so it seems still not called in usual flows. Other pending points:

You have to also check pre-commit CI, and the rest of the comments inline.

TheClaud99:
Sorry for my bad explanation, I try to be more clear. When the method _prepare_procurement_group_vals is called from here https://github.com/OCA/rma/blob/3436ec67afb63e438252a043e35a51173548bf0b/rma/wizard/stock_picking_return.py#L83

the recordset in self is empty, so this condition will never be true because self.order_id will always be false: 
https://github.com/OCA/rma/blob/3436ec67afb63e438252a043e35a51173548bf0b/rma_sale/models/rma.py#L167-L171
and this override in this case doesn't work. My commit tries to fix this case, but if it's not ok I'll try to find a cleaner solution.

claudio-mano:
Hi @pedrobaeza, I updated the code to be clearer, now the pre-commit CI works except for the tests. That's because the rma_sale module doesn't have a test method for the creation from stock.picking.return wizard like the one in the rma module, so the method I've added is not covered
https://github.com/OCA/rma/blob/3436ec67afb63e438252a043e35a51173548bf0b/rma/tests/test_rma.py#L679-L708

pedrobaeza:
Thanks @TheClaud99. Now the code and the problem is much clearer now. I think the main problem is this call with no RMA:

https://github.com/OCA/rma/blob/3436ec67afb63e438252a043e35a51173548bf0b/rma/wizard/stock_picking_return.py#L83

which is broken from the beginning. No partner is assigned as well.

What about creating the RMA, and then creating the procurement group from the RMA, so the rest of the code is valid? Putting also a `self.ensure_one()` in the prepare_vals will ensure no incorrect call is done.

TheClaud99:
The problem is that multiple RMAs could be created from the same stock.picking.return, which, according to the current logic, all share the same procurement.group:
https://github.com/OCA/rma/blob/3436ec67afb63e438252a043e35a51173548bf0b/rma/wizard/stock_picking_return.py#L98-L112
https://github.com/OCA/rma/blob/3436ec67afb63e438252a043e35a51173548bf0b/rma/wizard/stock_picking_return.py#L130-L131
Maybe the best solution could be to create the group by calling the method on the first created RMA and then set that group also to the others?

pedrobaeza:
Yes, I think the call should be done with a RMA in `self`. Then, we decide later when creating subsequent RMAs if the procurement group is the same and then just assign it from the previous create. If it's the same group for all unconditionally, then just assign it.

TheClaud99:
Hi @pedrobaeza, another solution I thought is to use the same logic as here (function _group_delivery_if_needed()):
https://github.com/OCA/rma/blob/09dd76a7a637f7a4c68945ff2bc5ce1098088649/rma/models/rma.py#L1121-L1124

I just also added this line
https://github.com/OCA/rma/blob/910e73a124cdafa8ddb972ed8132524970025db7/rma/wizard/stock_picking_return.py#L128
to maintain the same logic as before for the group name
https://github.com/OCA/rma/blob/09dd76a7a637f7a4c68945ff2bc5ce1098088649/rma/wizard/stock_picking_return.py#L101-L104

rma/wizard/stock_picking_return.py
```diff
@@ -129,6 +124,10 @@ def create_returns(self):
                 )
             vals_list = self._prepare_rma_vals_list()
             rmas = self.env["rma"].create(vals_list)
+            proc_group_vals = rmas._prepare_procurement_group_vals()
```
pedrobaeza: This must be called with only one records:

```suggestion
            proc_group_vals = rmas[:1]._prepare_procurement_group_vals()
```

TheClaud99:
@pedrobaeza I also made another fix that may be added to this pr: eeb9f7f.
The field sale_line_id on the move can be useful for populating the move_ids field on sale.order.line here:
https://github.com/odoo/odoo/blob/98408d48eecd7416275fbfc8b30769d844dbacba/addons/sale_stock/models/sale_order_line.py#L17
which is used for example to compute the delivered quantity on the sale line.
Can I merge the commit in this branch or should I create a new pr?

pedrobaeza:
This is done on purpose, as we don't want to affect the invoceability of the sales order. It's linked later.

TheClaud99:
Ok sorry, I didn't see that it's linked later

pedrobaeza (approved):

rma/wizard/stock_picking_return.py
```diff
@@ -129,6 +124,10 @@ def create_returns(self):
                 )
             vals_list = self._prepare_rma_vals_list()
             rmas = self.env["rma"].create(vals_list)
+            proc_group_vals = rmas[:1]._prepare_procurement_group_vals()
+            proc_group_vals["name"] = self.picking_id.name
+            proc_group = self.env["procurement.group"].create(proc_group_vals)
+            rmas.write({"procurement_group_id": proc_group.id})
```
sbejaoui, line 130: Hi @TheClaud99 ,

IMO, the issue is due to the non-unified of procurement groups creation.

When the picking return wizard is used, one procurement group is created for all RMAs. However, this is not the case when the sales order return wizard is used or when RMAs are created manually.

#423 addresses this issue by unifying the procurement group creation at RMA confirmation, regardless of the RMA origin.

I propose you revert these changes and keep the fix of the method name in rma_sale, 

what do you think?

sbejaoui (changes requested):

rma/wizard/stock_picking_return.py
TheClaud99: Hi @sbejaoui, I think it should be ok in your way as well (with also the advantage of having procurement.group creation unified), I can completely revert my commit 5f73bf6b5e962a30502b3d802c91142b56aae0f9 and wait for your pr to be merged

PR #417, opened 2024-08-15
[IMP] rma: configurable workflow

### LINKED ISSUE #413: RMA : configurable workflow
The RMA operation flow varies from company to company. Each company chooses its own process for managing returns, exchanges, and refunds.

The current version of the RMA base modules covers almost all use cases but lacks flexibility. More advanced use cases require a more configurable workflow. 

For instance:

1. **Refund without product return:** Some companies may offer refunds without requiring the return of the product, often in cases of low-value items or when the cost of return shipping is prohibitive.
2. **Return and exchange of the same product:** Customers may want to return a defective item and receive an identical replacement.
3. **Return of a non-ordered product:** Occasionally, customers receive items they did not order and need a process for returning these products. The delivered product don't figure on the sale order
4. **Return and exchange with a different product:** In some situations, customers may prefer to exchange the returned product for a different item, rather than receiving a refund or a direct replacement.

In addition to these use cases, when a company manages its own transportation, it may choose to pick up the returned product and deliver the replacement product simultaneously. This approach necessitates the creation of both the delivery order and the receipt order at the same time when the RMA is confirmed

If we summrize

| Use Case                            | Need return | Need delivery | Need refund |
|-------------------------------------|-------------|---------------|-------------|
| Refund without product return       | No | No | Yes |
| Return and exchange same product    | Yes | Yes | No |
| Return of non-ordered product       | Yes (return for a product different than the SO product) | No | No |
| Return and exchange with different product | Yes | Yes |No|

One way to manage these different use cases is to review the rigid flow and make it configurable based on the selected operation. We can imagine 

- Add three selection fields to rma.operation model that define when each action should happen:

| Action          | Timing Options   | Description                                                                                 |
|-----------------|------------------|---------------------------------------------------------------------------------------------|
| Create return   | - On confirm     | Initiate the return process as soon as the RMA is confirmed.                                |
|                 | - After receipt  | Create the return order only after the returned product is received.                        |
| Create delivery | - On confirm     | Start the delivery process immediately upon RMA confirmation.                               |
|                 | - After receipt  | Wait to create the delivery order until the returned product is received.                   |
| Create refund   | - On confirm     | Issue the refund as soon as the RMA is confirmed.                                           |
|                 | - After receipt  | Process the refund only after the product is received and inspected.                        |
|                 | - Update SO delivered quantity  | Adjust the sales order's delivered quantity based on the returned product. The refund will be generated from the SO.|

- Add a checkbox to the `rma.operation` model to define whether the product to be returned is different from the ordered product (use case 3).
- Add a checkbox to the `rma.operation` model to define whether the product to be delivered is different from the ordered product (use case 4).
- Add two fields to the `rma` model to define the product to be returned and the product to be delivered. By default, these two fields are hidden and have the same value as the ordered product. Their visibility depends on the selected operation.

I will appreciate your input and feedback on the proposed implementation.

cc/ @pedrobaeza , @jbaudoux , @lmignon , @rousseldenis , @phschmidt

sbejaoui:
As explained in #413, having a configurable RMA process helps cover more use cases by allowing the configuration of when the three main actions (reception, delivery, and refund) occur.

This PR adds configuration fields to the RMA operation, which allow managing the visibility of these actions in the RMA form. It also adds the possibility to automatically trigger these actions upon RMA confirmation or reception.

To minimize the impact on current users of this module, default values were set for these settings to maintain the existing workflow unchanged. After making these changes, there was no need to adjust the existing unit tests.

fixes: #413 

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -774,6 +774,7 @@ class Rma(models.Model):
         vals = self._prepare_common_procurement_vals(group=group)
         vals["route_ids"] = self.warehouse_id.rma_in_route_id
         vals["rma_receiver_ids"] = [(6, 0, self.ids)]
+        vals["to_refund"] = self.operation_id.action_create_refund == "update_quantity"
         if self.move_id:
             vals["origin_returned_move_id"] = self.move_id.id
         return vals
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -785,6 +785,8 @@ class Rma(models.Model):
         vals["to_refund"] = self.operation_id.action_create_refund == "update_quantity"
         if self.move_id:
             vals["origin_returned_move_id"] = self.move_id.id
+            if not self.operation_id.different_return_product:
+                vals["move_orig_ids"] = [(6, 0, self.move_id.ids)]
         return vals
 
     def _prepare_reception_procurements(self):
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -329,6 +329,14 @@ class Rma(models.Model):
     show_create_refund = fields.Boolean(
         string="Show Create refund Button", compute="_compute_show_refund_replace"
     )
+    return_product_id = fields.Many2one(
+        "product.product",
+        help="Product to be returned if it's different from the originally delivered "
+        "item.",
+    )
+    different_return_product = fields.Boolean(
+        related="operation_id.different_return_product"
+    )
 
     @api.depends("operation_id.action_create_receipt", "state", "reception_move_id")
     def _compute_show_create_receipt(self):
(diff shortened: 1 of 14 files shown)


rma/models/rma.py
```diff
@@ -317,6 +317,107 @@ def _domain_location_id(self):
...
+    )
+    show_create_replace = fields.Boolean(
+        string="Show Create replace Button", compute="_compute_show_create_replace"
+    )
+    show_create_refund = fields.Boolean(
+        string="Show Create refund Button", compute="_compute_show_refund_replace"
+    )
+
+    # Replace fields, used when the delivery is created automatically
+    replace_warehouse_id = fields.Many2one(
```
pedrobaeza: The computation of these fields will mean an overhead on existing installations. Why they are needed? This seems a new feature that is not linked with the configurable workflow.

rma/models/rma.py
```diff
@@ -317,6 +317,107 @@ def _domain_location_id(self):
         copy=False,
     )
```
pedrobaeza, line 319: Please avoid these empty lines for better readability of all the fields being shown together.

rma/models/rma.py
sbejaoui: In the operation, if the user selects 'Automatic Delivery' and 'Different Return Product,' the workflow will trigger the replacement action. Since the replacement action requires user input, the replacement information should be filled in the RMA at the time of its creation.

The computed fields for the replacement warehouse and replacement product UoM are intended to provide default values. A migration script can be added to create these columns, ensuring that their computation does not trigger on existing RMAs.

image

pedrobaeza: Please add the post_init_hook/migration script thus.

sbejaoui: @pedrobaeza ,

@jbaudoux  pointed out that replacing a product with a different one can be handled by processing an RMA reception with a refund and then creating a normal sale. I will remove this part I added to the workflow.

Another use case reported by the customer is receiving a different product than the one ordered. In this case, we need to receive a product that is not listed on the sale order and then redeliver the originally ordered product.

I am considering adding a new field, 'product to return,' which will be used if the operation has 'return a different product' set to True.
In this case, the reception will involve the different product. Otherwise, the standard behavior will apply, and the product_id will be used for both reception and delivery.

What are your thoughts on this?

jbaudoux (approved):

rousseldenis (approved):
Code review

PR #419, opened 2024-08-22
[16.0][ADD] add rma_lot

### LINKED ISSUE #416: RMA : Manage Lots
### Context
The customer want to return products tracked by lots to the seller. The seller want to ensure that only provided lots are returned.
#### Case 1
One reason are customers purchasing the same product from several suppliers to get it as quickly as possible, keep the first arrived package, and then returns the others to the sellers where it's the most convenient to make a return. The seller then gets back lots that have not been sent to the customer.
#### Case 2
The physically delivered lot is not always the one on the delivery order. This happens when lots are mixed on the stock location and the picker doesn't pay enough attention and doesn't pick the lot that is supposed to be picked

### Requirement
When the product is tracked by lot, the rma should require to select a lot that correspond to the selected delivery

### Solution proposal
* Add a lot_id on the rma, but don't make it required
* In the SO wizard and stock picking return wizard, populate the wizard with one line per lot and with the allowed quantity for this lot, when the product is tracked by lot. The lot on the line is required when the product is tracked by lot and is modifiable (but don't allow to create a lot). Ideally, set allowed quantity to 0 if the lot is expired
* When the SO wizard is validated, there is a check on the product quantity (that you don't try to return more than allowed). We want to keep this check at product level to allow to return a different lot (and provide some flexibility)
* In the portal, require the lot when the product is tracked by lot
* In the generated transfers, pre-fill the lot to receive on the return move created from the rma (using for example https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_restrict_lot). The purpose is also to pre-fill the lot on the move line (in case move lines creation is flagged in the picking type options). The receiver is able  to modify the move lines to choose a different lot (standard odoo)

@sbejaoui @pedrobaeza @chienandalu @lmignon @phschmidt

sbejaoui:
fixes: #416


rma/views/rma_views.xml
```diff
@@ -10,6 +10,7 @@
             <search>
                 <field name="name" />
                 <field name="user_id" />
+                <field name="product_id" />
```
peluko00: Why this change? It's better to inherit the view and add the field if it's necessary

rma_sale_lot/wizards/sale_order_rma_wizard.py
peluko00, line 1: Why this file it's empty?

peluko00:
Hi @pedrobaeza, can you please review that. I'm really interested in migrate this module to v17 because i need and it's probably very important for this repo

pedrobaeza:
Perform a full review and wait for the changes of the author.

peluko00:
The function not compute the product  when the lot is selected and the field product must be readonly.
image

peluko00:
Maybe @sbejaoui it's better to separate the two new modules in two prs

rma/views/rma_views.xml
sbejaoui: The product_id field is defined in the base module, so it makes sense to add it to the base search view

sbejaoui:
> The function not compute the product when the lot is selected and the field product must be readonly. 

I added the onchange functions so that the product and lot update when one or the other changes. I don't agree with making the product field readonly

rma_lot/models/rma.py
```diff
@@ -0,0 +1,35 @@
...
+        vals = super()._prepare_reception_procurement_vals(group=group)
+        vals["restrict_lot_id"] = self.lot_id.id
+        return vals
+
+    @api.onchange("product_id")
+    def _onchange_product(self):
+        self.update({"lot_id": False})
+
+    @api.onchange("lot_id")
+    def _onchange_lot(self):
```
rousseldenis: @sbejaoui Don't use onchanges. Prefer computed/stored/readonly=False field for lot.

Moreover, `product_id`is computed in `rma`module, so, override compute function instead

peluko00:
> > The function not compute the product when the lot is selected and the field product must be readonly.
> 
> I added the onchange functions so that the product and lot update when one or the other changes. I don't agree with making the product field readonly

Thanks but if you choose an lot_id it's link to product and doesn't associate to another. If lot_id it's not filled it's normal to choose product

peluko00:
> Maybe @sbejaoui it's better to separate the two new modules in two prs

What is your opinion in that case @rousseldenis?

rousseldenis:
> > Maybe @sbejaoui it's better to separate the two new modules in two prs
> 
> What is your opinion in that case @rousseldenis?

As both modules are new, I would say it's not really annoying. The problem with that is when changes in one module are accepted and not in the other. That couples the PR lifetime to the slowest.

@sbejaoui Could you do a PR for rma search view modification as the module version won't be updated when we will merge this (@peluko00 this is the more annoying case)

peluko00:
> > > Maybe @sbejaoui it's better to separate the two new modules in two prs
> > 
> > 
> > What is your opinion in that case @rousseldenis?
> 
> As both modules are new, I would say it's not really annoying. The problem with that is when changes in one module are accepted and not in the other. That couples the PR lifetime to the slowest.
> 
> @sbejaoui Could you do a PR for rma search view modification as the module version won't be updated when we will merge this (@peluko00 this is the more annoying case)

And related to the two new modules?

sbejaoui:
> > > Maybe @sbejaoui it's better to separate the two new modules in two prs
> > 
> > 
> > What is your opinion in that case @rousseldenis?
> 
> As both modules are new, I would say it's not really annoying. The problem with that is when changes in one module are accepted and not in the other. That couples the PR lifetime to the slowest.
> 
> @sbejaoui Could you do a PR for rma search view modification as the module version won't be updated when we will merge this (@peluko00 this is the more annoying case)

#430

rma/views/rma_views.xml
peluko00: what do you think @rousseldenis ?

rousseldenis: This is already merged. See #430

rousseldenis (approved):

peluko00 (approved):
LGTM, tested in runboat

peluko00:
Seems it's ready to merge @rousseldenis

mpascuall (approved):
LGTM, tested in runboat and working as expected

PR #420, opened 2024-08-28
[16.0][FIX] rma: fix returned quantity in picking return wizard

### LINKED ISSUE #418: rma: The returned quantity is not correctly computed when a reception is created through RMA
When performing multiple partial returns on the same picking, the expected behavior is for the return wizard to count down the already returned quantity (standard odoo).

However, this is not the case when receptions are created through an RMA. After a partial reception, the wizard incorrectly proposes the entire original quantity instead of the remaining quantity. 

## Module
rma
## To Reproduce
**Affected versions**: 16

Steps to reproduce:
1. Select a delivery
2. Perform a partial return
3. Validate the reception
4. Go back to the delivery
5. Open the return wizard

You'll notice that the wizard proposes the whole quantity again, rather than the remaining quantity that hasn't been returned yet.

**Expected behavior**
The wizard should propose the remaining quantity

**Delivery**
image
**Reception**
image
**Return wizard after reception**
image

cc/ @pedrobaeza , @jbaudoux , @lmignon , @rousseldenis , @phschmidt

sbejaoui:
fixes: #418

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -701,6 +701,7 @@ class Rma(models.Model):
         vals["rma_receiver_ids"] = [(6, 0, self.ids)]
         if self.move_id:
             vals["origin_returned_move_id"] = self.move_id.id
+            vals["move_orig_ids"] = [(6, 0, self.move_id.ids)]
         return vals
 
     def _prepare_reception_procurements(self):
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -848,3 +848,33 @@ class TestRmaCase(TestRma):
         )
         self.assertTrue(rma.name in mail_receipt.subject)
         self.assertTrue("products received" in mail_receipt.subject)
+
+    def _create_return_wizard(self, picking):
+        stock_return_picking_form = Form(
+            self.env["stock.return.picking"].with_context(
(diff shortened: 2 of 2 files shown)


pedrobaeza (commented):
AFAIK, this provokes some side effects in the "assign" method, not letting to fetch quants freely, but not sure if it applies here. Check also the effect in the delivered quantity, forecast, and so on.

sbejaoui:
> AFAIK, this provokes some side effects in the "assign" method, not letting to fetch quants freely, but not sure if it applies here. Check also the effect in the delivered quantity, forecast, and so on.

Quant reservation is restricted by the original move when the source location is internal. However, this restriction is not applied if the source location is a supplier location.

For the quantities on the sale order, a move is considered only if it's linked to the sale line. BTW, I pushed a [commit](https://github.com/OCA/rma/pull/417/commits/11fec6408d2242d7444515e938f7fff6ae79b90a) to allow this link if we want to have a standard refund after RMA reception.

pedrobaeza:
What happens if you have RMA reception in several steps? I think this won't be correct in that case.

sbejaoui:
> What happens if you have RMA reception in several steps? I think this won't be correct in that case.

I will add a test to be sure

PR #421, opened 2024-08-29
[16.0][FIX] rma: same procurement group for reception and delivery

sbejaoui:
The procurement group created for reception was not being correctly assigned to the RMA, resulting in a different group being generated for the delivery.

This fix ensures that the same procurement group is used for both operations.

cc/ @jbaudoux , @lmignon , @rousseldenis

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -820,6 +820,7 @@ class Rma(models.Model):
             group = rma.procurement_group_id
             if not group:
                 group = group_model.create(rma._prepare_procurement_group_vals())
+                rma.procurement_group_id = group
             product = self.product_id
             if self.different_return_product:
                 if not self.return_product_id:
@@ -1271,7 +1272,11 @@ class Rma(models.Model):
             key=lambda rma: [rma._delivery_group_key()],
         )
         for _group, rmas in grouped_rmas:
-            rmas = self.browse().concat(*list(rmas))
+            rmas = (
+                self.browse()
+                .concat(*list(rmas))
+                .filtered(lambda rma: not rma.procurement_group_id)
+            )
             if not rmas:
                 continue
             proc_group = self.env["procurement.group"].create(
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -164,18 +164,6 @@ class Rma(models.Model):
                 vals["sale_line_ids"] = [(4, line.id)]
         return vals
 
-    def _prepare_procurement_group_vals(self):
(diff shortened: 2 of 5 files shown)


rma_sale/models/rma.py
```diff
@@ -163,15 +163,3 @@ def _prepare_refund_line_vals(self):
...
-        vals = super()._prepare_procurement_group_vals()
-        if not self.env.context.get("ignore_rma_sale_order") and self.order_id:
-            vals["sale_id"] = self.order_id.id
-        return vals
-
-    def _prepare_delivery_procurements(self, scheduled_date=None, qty=None, uom=None):
-        self = self.with_context(ignore_rma_sale_order=True)
-        return super()._prepare_delivery_procurements(
-            scheduled_date=scheduled_date, qty=qty, uom=uom
-        )
```
sbejaoui: @pedrobaeza ,

I am not aware of the history behind this code, but I believe it was implemented this way to avoid linking the rma delivery pickings to the sales order, as in recent versions, this linkage creates a sales line if the procurement group is associated to a sale order without an existing line.

This approach has led to:

- Duplication of procurement groups, one for the reception and another for the deliveries.
- In the sales order, the RMA reception picking appears among the sales order pickings, but the delivery pickings do not.
- This unclear usage of context keys.

I propose to completely cut the link between the RMA procurement group and the sales order, maintaining a single procurement group for all RMA pickings.

In #417, I am working on a different approach to establish this link between the RMA stock moves and the sales line, allowing standard refunding to be configurable based on the operation.

rma/models/rma.py
```diff
@@ -1111,12 +1112,9 @@ def _group_delivery_if_needed(self):
...
         )
         for _group, rmas in grouped_rmas:
-            rmas = (
-                self.browse()
-                .concat(*list(rmas))
-                .filtered(lambda rma: not rma.procurement_group_id)
-            )
-            if not rmas:
+            rmas = self.browse().concat(*list(rmas))
+            if not rmas or len(rmas.procurement_group_id) == 1:
```
jbaudoux: But if some rma in the group don't have a procurement group, then you don't give them one anymore. I don't think you need to change the test

jbaudoux (changes requested):

rma/models/rma.py
jbaudoux: @sbejaoui Can you check this ?

pedrobaeza:
@sbejaoui I suppose this one can be closed, isn't it? Is #417 still applicable? If so, please reabase and fix conflicts.

PR #423, opened 2024-08-29
[16.0][FIX] rma: group rma receptions

### LINKED ISSUE #422: RMA: group reception
When returning an order with a large number of items, it can be challenging for the stock manager to handle multiple separate receptions for each RMA, especially when all the returned items are likely to arrive in a single package. 

Proposed Solution:
- Add a configurable setting at the company level that allows enabling or disabling the grouping of RMA receptions.
- If this option is enabled, all RMAs within the same batch will be grouped under a single procurement group.

sbejaoui:
fixes: #422

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -852,6 +852,10 @@ class Rma(models.Model):
         self = self.filtered(lambda rma: rma.state == "draft")
         if not self:
             return
+        if not self.procurement_group_id:
+            self.procurement_group_id = self.env["procurement.group"].create(
+                self._prepare_procurement_group_vals()
+            )
         self.write({"state": "confirmed"})
         for rma in self:
             rma._add_message_subscribe_partner()
@@ -1242,11 +1246,7 @@ class Rma(models.Model):
             key=lambda rma: [rma._delivery_group_key()],
         )
         for _group, rmas in grouped_rmas:
-            rmas = (
-                self.browse()
-                .concat(*list(rmas))
-                .filtered(lambda rma: not rma.procurement_group_id)
-            )
+            rmas = self.browse().concat(*list(rmas))
             if not rmas:
                 continue
             proc_group = self.env["procurement.group"].create(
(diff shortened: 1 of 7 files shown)


rma/models/res_company.py
```diff
@@ -26,6 +26,9 @@ def _default_rma_mail_draft_template(self):
         except ValueError:
             return False
 
+    rma_reception_grouping = fields.Boolean(
+        string="Group RMA receptions", default=False
```
rousseldenis: Maybe a help string should be great

rousseldenis (approved):
LGTM

jbaudoux (approved):

pedrobaeza:
Seeing the original problem, this fix shouldn't be an option, but act always this way, don't you think?

rma/models/rma.py
```diff
@@ -732,6 +732,10 @@ def action_confirm(self):
         self = self.filtered(lambda rma: rma.state == "draft")
         if not self:
             return
+        if self.env.company.rma_reception_grouping:
```
```suggestion
        if not self.procurement_group_id:
```

jbaudoux: Indeed, it should be the default behavior and then there is no need for an option. I don't see anything setting a value in `procurement_group_id` except `create_receipt` that you fixed here https://github.com/OCA/rma/pull/421/files#diff-3d6f82bd3aa96a98f1dd5cd6389cb926247243353dda39dc5bbf9924b53c98cbR715 and `create_replace` here https://github.com/OCA/rma/blob/16.0/rma/models/rma.py#L1212

So this fix should be obsolete as what Pedro is suggesting is already like that. Or there is something else setting a value here that needs to be overridden ?

pedrobaeza: You may set this value manually if you are interested to put a specific group, so let's keep this option open.

pedrobaeza: @sbejaoui any news here?

sbejaoui: I just got back from vacation today. I will check all RMA PRs this week.

pedrobaeza:
Please solve the conflicts

rousseldenis:
@pedrobaeza Ok for you ?

pedrobaeza (approved):
Sorry for the late reply.

Please fw-port it.

/ocabot merge patch

PR #429, opened 2024-10-16
[16.0][IMP] rma: add dashboard

sbejaoui:
image

### DIFF
--- a/rma/models/rma_operation.py
+++ b/rma/models/rma_operation.py
@@ -1,7 +1,14 @@
 # Copyright 2020 Tecnativa - Ernesto Tejeda
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from ast import literal_eval
+from collections import defaultdict
+
+from odoo import _, api, fields, models
+from odoo.osv.expression import AND
+
+PROCESSED_STATES = ["received", "refunded", "replaced", "finished"]
+AWAITING_ACTION_STATES = ["waiting_return", "waiting_replacement", "confirmed"]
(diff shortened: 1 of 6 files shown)


rma/models/rma_operation.py
```diff
@@ -10,7 +13,145 @@ class RmaOperation(models.Model):
 
     active = fields.Boolean(default=True)
     name = fields.Char(required=True, translate=True)
+    color = fields.Integer()
+    count_rma_draft = fields.Integer(compute="_compute_count_rma_draft")
```
pedrobaeza: Better to do this in one compute, and with `read_group` by `state`, obtain all the numbers and sum them according your needs.

sbejaoui: done

chienandalu (approved):
Cool :sunglasses: :+1:

rousseldenis:
@sbejaoui As requested operation is optional, I would have added a new card for those `awaiting operation`.

As those RMA don't appear in any current card.

What do you think ?

sbejaoui:
> @sbejaoui As requested operation is optional, I would have added a new card for those `awaiting operation`.
> 
> As those RMA don't appear in any current card.
> 
> What do you think ?

The operation is required to confirm the RMA. While it may be useful to show users RMAs awaiting operation, the working cost is much higher than the added value, in my opinion, as the dashboard is a Kanban view for the operation model and can't display a card for a non-existent record without diving deep into JS widgets.

pedrobaeza:
ping @sbejaoui

rousseldenis (approved):

PR #434, opened 2024-10-22
[IMP] rma: custom route for replace action

sbejaoui:
RMA replacement orders currently use the default delivery route, which results in their creation under the default delivery picking type. While this behavior works, it prevents users from tracking RMA replacement orders separately from regular deliveries.

This PR introduces a new field, `rma_out_replace_route_id`, to the warehouse. If set, this field specifies a custom route for RMA replacement actions, allowing users to separate RMA replacement orders from standard deliveries and manage them independently. If the field is not set, the default delivery route will continue to be used.

### DIFF
--- a/rma/models/stock_warehouse.py
+++ b/rma/models/stock_warehouse.py
@@ -30,6 +30,7 @@ class StockWarehouse(models.Model):
     )
     rma_in_route_id = fields.Many2one("stock.route", "RMA in Route")
     rma_out_route_id = fields.Many2one("stock.route", "RMA out Route")
+    rma_out_replace_route_id = fields.Many2one("stock.route", "RMA out Replace Route")
 
     def _get_rma_location_values(self, vals, code=False):
         """this method is intended to be used by 'create' method
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -2,6 +2,7 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
+from odoo import Command
 from odoo.exceptions import UserError, ValidationError
 from odoo.tests import Form, TransactionCase, new_test_user, users
 from odoo.tools import mute_logger
@@ -106,6 +107,20 @@ class TestRma(TransactionCase):
         rma.action_confirm()
         rma.reception_move_id.quantity_done = rma.product_uom_qty
         rma.reception_move_id.picking_id._action_done()
+        return rma
+
+    def _receive_and_replace(self, partner, product, qty, location):
+        rma = self._create_confirm_receive(partner, product, qty, location)
+        delivery_form = Form(
+            self.env["rma.delivery.wizard"].with_context(
+                active_ids=rma.ids,
+                rma_delivery_type="replace",
+            )
+        )
+        delivery_form.product_id = rma.product_id
+        delivery_form.product_uom_qty = qty
+        delivery_wizard = delivery_form.save()
+        delivery_wizard.action_deliver()
         return rma
 
     def _create_delivery(self):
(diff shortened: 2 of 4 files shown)


rma/tests/test_rma.py
```diff
@@ -848,3 +849,67 @@ def test_autoconfirm_email(self):
         )
         self.assertTrue(rma.name in mail_receipt.subject)
         self.assertTrue("products received" in mail_receipt.subject)
+
+    def _test_receive_and_replace(self):
```
pedrobaeza: I don't feel comfortable with helper methods starting with `_test`, as they seem "commented" methods to not be run, and more if they are not doing any assert.

pedrobaeza (approved):
We are going to need more routes than the a highway, hehe

rma/models/stock_warehouse.py
```diff
@@ -30,6 +30,7 @@ class StockWarehouse(models.Model):
     )
     rma_in_route_id = fields.Many2one("stock.route", "RMA in Route")
     rma_out_route_id = fields.Many2one("stock.route", "RMA out Route")
+    rma_out_replace_route_id = fields.Many2one("stock.route", "RMA out Replace Route")
```
rousseldenis, line 33: @sbejaoui Should'nt we use the mechanisms to create that route (see `_get_route_values()`) ?

sbejaoui: I believe the common use case is to use the default delivery route, which is why I didn't create or assign a custom route to the warehouse. If more people feel that having a custom route for RMA delivery by default is necessary, I will make the change.

jbaudoux (commented):
Why only for replace and not for all RMA deliveries?

While it's doing the job, I'm not sure it's a good approach. You can configure specific warehouse behavior by configuring a route on a product or product category. This is even more important in newer versions of odoo where you can configure a route on the carrier. If you force a route from rma, you are overriding those configurations.

From what I understood, you want to isolate the rma delivery from other deliveries. However, this is already the case as you have a dedicated procurement group for the rma delivery. The issue arise when you use the `stock_picking_group_by_partner_by_carrier` module that regroup deliveries for a same partner. In my opinion, a cleaner approach is to create a glue module with an option on the picking type to disable rma grouping.

sbejaoui:
> Why only for replace and not for all RMA deliveries?

The rma module differentiates between replacement and delivery actions. In the case of a replacement, the product is picked from stock, while for delivery, it’s picked from the RMA location.
To support this, there’s an existing route for delivery (`rma_out_route_id`). IMO, allowing the configuration of a separate route for replacements aligns it more closely with the delivery process and adds flexibility to manage both workflows.

> While it's doing the job, I'm not sure it's a good approach. You can configure specific warehouse behavior by configuring a route on a product or product category. This is even more important in newer versions of odoo where you can configure a route on the carrier. If you force a route from rma, you are overriding those configurations.

I agree, the use of routes in the rma module should be reconsidered to better align with the Odoo standard. While I don’t have specific ideas yet on how to achieve this alignment, my point is that this PR follows the existing logic in the module and doesn’t introduce any new approaches or changes to the current structure.

https://github.com/OCA/rma/blob/9f192814e4bcc719329b6939a728b3c3764c595d/rma/models/stock_warehouse.py#L23-L30

> From what I understood, you want to isolate the rma delivery from other deliveries. However, this is already the case as you have a dedicated procurement group for the rma delivery. The issue arise when you use the `stock_picking_group_by_partner_by_carrier` module that regroup deliveries for a same partner. In my opinion, a cleaner approach is to create a glue module with an option on the picking type to disable rma grouping.

One reason for grouping pickings is indeed the `stock_picking_group_by_partner_by_carrier`, but that's not the only reason.
Another important reason is to assign deliveries to the rma picking type, which provides them with a distinct sequence number, separate from the default delivery sequence. 
This differentiation is important for delivery managers, as it allows them to easily identify rma deliveries by their reference in delivery documents and in the tree view,..
Additionally, having separate sequences ensure accurate reporting, allowing rma deliveries to be excluded from standard sales delivery statistics if we analyse stock.picking data.

jbaudoux:
> The rma module differentiates between replacement and delivery actions. In the case of a replacement, the product is picked from stock, while for delivery, it’s picked from the RMA location. To support this, there’s an existing route for delivery (`rma_out_route_id`). IMO, allowing the configuration of a separate route for replacements aligns it more closely with the delivery process and adds flexibility to manage both workflows.

Thanks for the clarification. I had another look with your comment in mind and I do understand there is a `create_replace` method that creates the delivery for the replacement. However, I do not understand why this is not calling the already existing `_prepare_delivery_procurements` instead of `_prepare_replace_procurements` as this is for creating the delivery. There you can configure the route as you mentioned, so that would drop the need for another route in case of replace only.
In regards of where it is picked, it is up to you to configure the route if you want to use a specific one. The fact that the customer returns or not a faulty part should not impact how you will pick the new part from your stock, so I don't see why having 2 different routes. @mt-software-de Maybe you can clarify why you used 2 separate methods?

cc @phschmidt

mt-software-de:
> > The rma module differentiates between replacement and delivery actions. In the case of a replacement, the product is picked from stock, while for delivery, it’s picked from the RMA location. To support this, there’s an existing route for delivery (`rma_out_route_id`). IMO, allowing the configuration of a separate route for replacements aligns it more closely with the delivery process and adds flexibility to manage both workflows.
> 
> Thanks for the clarification. I had another look with your comment in mind and I do understand there is a `create_replace` method that creates the delivery for the replacement. However, I do not understand why this is not calling the already existing `_prepare_delivery_procurements` instead of `_prepare_replace_procurements` as this is for creating the delivery. There you can configure the route as you mentioned, so that would drop the need for another route in case of replace only. In regards of where it is picked, it is up to you to configure the route if you want to use a specific one. The fact that the customer returns or not a faulty part should not impact how you will pick the new part from your stock, so I don't see why having 2 different routes. @mt-software-de Maybe you can clarify why you used 2 separate methods?
> 
> cc @phschmidt

Sorry @jbaudoux , but as I already wrote, I will not be contributing to the RMA Reop any more for now.
As a little help, there are two methods because the requirements are different. This was already the case before I contributed to it. My customer never used create_replace.

pedrobaeza:
Please fw-port it to upper versions.

/ocabot merge minor

PR #437, opened 2024-11-14
[IMP] rma: add index for rma_id in stock.move model

sbejaoui:
Add an index to rma_id in the stock.move model to improve access performance for the one2many field delivery_move_ids


pedrobaeza (approved):

rousseldenis (approved):

rvalyi:
As per reviews,

/ocabot merge minor

PR #440, opened 2025-01-08
[IMP] rma: improve rma report layout

sbejaoui:
- Use a table instead of a div.
- Remove the move name, as it is not relevant to the customer and is redundant with the SO number and product reference.

**Before**
image
**After**
image

### DIFF
--- a/rma/views/report_rma.xml
+++ b/rma/views/report_rma.xml
@@ -95,31 +95,36 @@
                         </p>
                     </div>
                 </div>
-                <div class="row mt32 mb32" id="product_information">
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Which is the before and a summary of the changes?

sbejaoui:
> Which is the before and a summary of the changes?

I updated the PR description,

pedrobaeza (approved):
This may provoke problems with inherited modules extending the old report, but I don't think there should be many and it worths IMO.

chienandalu (approved):
Thanks :)

pedrobaeza:
/ocabot merge minor

Please fw-port it if applicable.

PR #441, opened 2025-02-03
[16.0][IMP] rma_sale: Avoid computing RMA move unnecessarily

sbejaoui:
In the RMA sale wizard, the move is computed after the wizard line is generated. This is inefficient since the line is created after iterating through the moves related to the sale order.

Assigning the move to the wizard line prevents another issue where all RMAs would be linked to a single move if the sale order line is associated with multiple moves.

### DIFF
--- a/rma_sale/models/sale.py
+++ b/rma_sale/models/sale.py
@@ -34,6 +34,7 @@ class SaleOrder(models.Model):
             "sale_line_id": data["sale_line_id"].id,
             "uom_id": data["uom"].id,
             "picking_id": data["picking"] and data["picking"].id,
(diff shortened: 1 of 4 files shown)


rousseldenis (approved):
Seems reasonable

FrancoMaxime (approved):
LGTM: code review

sbejaoui:
ready to merge

pedrobaeza (commented):
I don't like to remove the compute and "do things manually". Others may rely on such compute for getting the value populated. Why can't you restrict the compute scope for such painful cases?

PR #451, opened 2025-02-17
[FIX] rma_sale_lot: force ml reservation to the restricted lot

sbejaoui:


rma_sale_lot/models/rma.py
```diff
@@ -0,0 +1,15 @@
...
+
+from odoo import models
+
+
+class Rma(models.Model):
+
+    _inherit = "rma"
+
+    def _create_receipt(self):
+        res = super()._create_receipt()
```
rousseldenis: @sbejaoui Do you know why tests seem to not pass here ? See coverage

sbejaoui: _create_receipt is defined in #413 , is not merged yet.

rousseldenis (approved):

PR #452, opened 2025-02-20
[17.0][MIG] rma_sale: Migration to 17.0

victoralmau:
Migration to 17.0

Please @chienandalu can you review it?

@Tecnativa TT52353


rma_sale/views/sale_portal_template.xml
```diff
@@ -15,7 +15,7 @@
                     class="btn-close"
                     aria-label="Close"
                     data-bs-dismiss="modal"
-                >&amp;times;</button>
+                />
```
pedrobaeza, line 18: Why removing the word `times`?

victoralmau: This is to avoid showing the icon 2 times (I'm seeing that it happens also in v16, I'm going to fix it too).
ejemplo

pedrobaeza (approved):

rma_sale/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,33 @@
...
+                    class="col-12 col-lg-6 o_setting_box"
+                    title="Show portal RMA request in a single page"
+                >
+                    <div class="o_setting_left_pane">
+                        <field name="show_full_page_sale_rma" />
+                    </div>
+                    <div class="o_setting_right_pane">
+                        <label
+                            for="show_full_page_sale_rma"
+                            string="Single page RMA request"
```
chienandalu, line 18: I struggled to find this feature that I know existed :sweat_smile: 

Can you:

- Put it in the RMA config section
- Add a configuration README entry

rma_sale/static/src/js/rma_portal_form.esm.js
```diff
@@ -0,0 +1,125 @@
...
+            .map(function () {
+                return this.name.replace("-operation_id", "");
+            })
+            .get();
+        this.$submit = $("#form-request-rma button[type='submit']");
+        this.rows_ids = ids;
+        // We'll build an object that will ease the form check. It could be further
+        // extended with additional checks.
+        this.rows = {};
+        $.each(ids, (id) => {
```
chienandalu: Whatch out :eyes:  `$.each` isn't a drop in replacement for `_.each` and here it's behaving completly different and it was inducing errors (and as git wasn't catching the diff for this file it was tricky to spot why).

Here's the [right replacement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach), as we want to iterate over the array elements (it's recommend to try to replace the underscore stuff with vanilla js when possible)
```suggestion
        ids.forEach((id) => {
```

chienandalu (changes requested):

chienandalu (approved):
Thanks!

PR #455, opened 2025-03-19
[16.0][FIX] rma_sale: replacements create order lines

chienandalu:
When the RMA procurement group has a sale set it will propagate to the pickings. This is inconvenient for this operation as when we confirm the customer delivery a new order line will be created with the replaced option which will be set for invoicing.

cc @Tecnativa TT55461

Not sure about the fix though, as it could be argumented that the replacement could be invoiced with a price of 0 or whatever difference the user considers at the moment. In any case, 

This is the relevant core code: https://github.com/odoo/odoo/blob/b1d88c20a6cf2a2f283ed04f145cb9e7fc614755/addons/sale_stock/models/stock.py#L85-L117

In the other hand if there were automatic invoicing processes, wrong invoices would be sent to customers, causing confussions...

Opinions? @sbejaoui @pedrobaeza @victoralmau

### DIFF
--- a/rma_sale/README.rst
+++ b/rma_sale/README.rst
@@ -120,13 +120,13 @@ OCA, or the Odoo Community Association, is a nonprofit organization whose
 mission is to support the collaborative development of Odoo features and
 promote its widespread use.
 
-.. |maintainer-chienandalu| image:: https://github.com/chienandalu.png?size=40px
-    :target: https://github.com/chienandalu
-    :alt: chienandalu
+.. |maintainer-pedrobaeza| image:: https://github.com/pedrobaeza.png?size=40px
+    :target: https://github.com/pedrobaeza
+    :alt: pedrobaeza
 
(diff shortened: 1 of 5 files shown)


sbejaoui:
We had the same issue, but since we use stock_picking_group_by_partner_by_carrier, which replaces sale_id with sale_ids, we bypassed the problem with this workaround: https://github.com/OCA/stock-logistics-workflow/pull/1789.

chienandalu:
Thank you @sbejaoui we don't use those modules but as I see it's an issue for you as well and not a feature :sweat_smile: 

The only thing I'm worried about is that `sale_id` is a stored relation to the `group_id.sale_id` field and I'm not sure if that relation could be reset in the meantime of validating the picking... which I don't think is going to happen but it isn't impossible...

pedrobaeza:
I have added a regression tests, so let's go with this one for fixing the behavior.

/ocabot merge patch

PR #457, opened 2025-05-01
[17.0][FIX] rma_sale: replacements create order lines

pedrobaeza:
Forward-port of #455 

When the RMA procurement group has a sale set it will propagate to the pickings. This is inconvenient for this operation as when we confirm the customer delivery a new order line will be created with the replaced option which will be set for invoicing.

@Tecnativa TT55461


pedrobaeza:
Finishing the fw-port chain:

/ocabot merge patch

PR #464, opened 2025-07-30
[16.0][IMP] rma: Allow manual finish of confirmed RMAs with no further processing required

sbejaoui:
I came with a special case: a product was delivered by mistake but isn’t worth returning and can’t be invoiced. The operation is configured to prevent any further action (no receipt, no delivery, no refund). The goal is to track the case through an RMA, but still allow the user to close it manually when no follow-up is needed.

This PR enables users to manually finish RMAs after confirmation in such cases

cc/ @pedrobaeza , @rousseldenis , @lmignon

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -337,6 +337,24 @@ class Rma(models.Model):
     different_return_product = fields.Boolean(
         related="operation_id.different_return_product"
     )
+    requires_action = fields.Boolean(
+        compute="_compute_requires_action",
+        help="Indicates whether the RMA requires processing such as a receipt,"
+        "delivery, or refund.",
+    )
+
+    @api.depends("operation_id")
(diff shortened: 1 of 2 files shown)


pedrobaeza:
If I understood this correctly, now on all the previous done RMAs, the button "Finish" will appear?

sbejaoui:
> If I understood this correctly, now on all the previous done RMAs, the button "Finish" will appear?

No, it will only appear in this specific case where no action is set on the operation, and the RMA is stuck in the confirmed state. 
This change allows users to close such RMAs manually. All other RMAs with a regular operation that includes a follow-up action will not be affected — the "Finish" button will behave as it does currently.

pedrobaeza (approved):

lmignon (approved):

rousseldenis (approved):

PR #465, opened 2025-07-31
[16.0][IMP] rma: improve reception grouping

sbejaoui:
Group RMA receptions by partner, company, and warehouse

cc/ @pedrobaeza , @rousseldenis , @jbaudoux

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -3,6 +3,7 @@
 # Copyright 2023 Michael Tietz (MT Software) <mtietz@mt-software.de>
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 import logging
+import warnings
 from collections import defaultdict
 from itertools import groupby
 
@@ -878,7 +879,7 @@ class Rma(models.Model):
         self = self.filtered(lambda rma: rma.state == "draft")
         if not self:
             return
-        self._group_reception_if_needed()
+        self._assign_reception_procurement_group()
         self.write({"state": "confirmed"})
         for rma in self:
             rma._add_message_subscribe_partner()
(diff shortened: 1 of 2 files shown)


rma/models/rma.py
```diff
@@ -1261,6 +1258,26 @@ def _delivery_group_key(self):
         self.ensure_one()
         return (self.partner_shipping_id.id, self.company_id.id, self.warehouse_id.id)
 
+    def _reception_group_key(self):
```
pedrobaeza: I think calling methods by this convention makes better to understand its goal (`get`: return something):

```suggestion
    def _get_reception_group_key(self):
```

rma/models/rma.py
```diff
@@ -1261,6 +1258,26 @@ def _delivery_group_key(self):
...
+        self.ensure_one()
+        return (self.partner_id.id, self.company_id.id, self.warehouse_id.id)
+
+    def _group_reception_if_needed(self):
```
pedrobaeza: Isn't this better naming?

```suggestion
    def _assign_procurement_group(self):
```

sbejaoui: I agree, but I wanted to follow the same naming for delivery

https://github.com/OCA/rma/blob/09c8d9154a531806c4e448a5e521b175f226789d/rma/models/rma.py#L1043-L1052

pedrobaeza: Well, OK then.

lmignon (approved):
LGTM

phschmidt (approved):
Functionally tested

sbejaoui:
@pedrobaeza , can you update your review please

rma/models/rma.py
pedrobaeza: This comment is still pending.

sbejaoui: Agree, but I wanted to follow the same naming for delivery

https://github.com/OCA/rma/blob/e9e639f6fef7bd5069e80e8ba9d4016098b10c15/rma/models/rma.py#L1271-L1274

pedrobaeza: Well, if something is incorrect previously, let's do a `[REF]` changing it. We can keep the previous method name calling the new one for compatibility.

sbejaoui: Ok I will do,

pedrobaeza (approved):
Thanks for the changes.

/ocabot merge patch

pedrobaeza:
Please fw-port it to 17 and 18, or it will be a mess for other modules not having these new hooks/renamings.

PR #466, opened 2025-08-04
[16.0][FIX] rma: add access read to stock user

sbejaoui:
The rma_receiver_ids field is is in the list of fields for moves merge, which requires read access to the rma model.
This PR adds read access to stock users, preventing access errors during move merging


chienandalu (approved):

pedrobaeza (commented):
I'm a bit reluctant of using this solution, as it gives too much access to stock users. Can we find another solution for the merge problem?

phschmidt (approved):
Functionally tested

PR #468, opened 2025-08-11
[18.0][MIG] stock_production_lot_warranty

georgiedekker:
Migrate stock_production_lot_warranty to version 18.0

Related to #425

This PR migrates the `stock_production_lot_warranty` module from version 15.0 to 18.0.

## Changes Made
- Updated module version from `15.0.1.1.0` to `18.0.1.0.0`
- Applied pre-commit formatting and linting
- Added `pyproject.toml` for build configuration
- Converted readme files from .rst to .md format
- Module is fully compatible with Odoo 18 API

## Module Description
This module adds warranty expiration date functionality to stock production lots/serial numbers, automatically calculating the warranty expiration based on product warranty settings.

## Testing
- All pre-commit hooks pass
- Code is compatible with Odoo 18 APIs
- No deprecated methods or patterns detected


MarinaAForgeFlow:
I have found this module while searching for a solution for customer warranty dates in odoo 19. I was just wondering if it makes sense that the warranty date is initially set using the lot creation date (as it is done with an onchange on the product, which basically will trigger the method at creation). I understand that the intention of this module is in fact to specify customer warranty, since it is using the warranty field on the sales tab of the product. But then, wouldn't be most appropriate, based on the common behavior of warranties, that this date is set when selling the product, instead of at lot creation (which will probably be after manufacturing or receiving the product from the supplier).

Perhaps what would make more sense is to set the warranty date only when the lot is sold. And if you want to have **your** warranty date (the one that comes from the supplier) then another field could be created.

However, since this module is already used in several versions, I understand we can not change the behavior. So I am not sure how to approach this. Maybe creating a stock_production_lot_warranty_sale with the suggested logic will be fine?

What do you think? @osi-scampbell @max3903 @georgiedekker

georgiedekker:
@MarinaAForgeFlow I'm not actively developing on Odoo at the moment, but I did some conversions of proprietary modules to Odoo 18 and 19 which is why I created this PR, which hasn't found any reception since I created it in August, so I'm not sure that there is any active development in this repository. Happy to be of assistance if that would help you. The chance you describe sounds fair and doable.

MarinaAForgeFlow (approved):
Code looks good. The only issue is that the commit history was not preserved.

PR #470, opened 2025-08-11
[16.0][IMP] rma: operation setting for delivery grouping exception

sbejaoui:
Add a `prevent_delivery_grouping` flag on `rma.operation` to allow specific operations to bypass the company-level RMA delivery grouping configuration

cc/ @pedrobaeza , @rousseldenis , @lmignon


pedrobaeza (commented):
Why not switching the company flag as the default value for RMA operations and only put the `if` over this last one?

sbejaoui:
> Why not switching the company flag as the default value for RMA operations and only put the `if` over this last one?

It makes sense to me, the default behavior is to group deliveries the same way we did for receptions, so there’s no need for a company-level setting. Splitting deliveries is the exception, and we’ll keep the setting for it at the operation level.

If you agree, I’ll make the changes.

pedrobaeza:
OK, go ahead.

lmignon (approved):
LGTM (Code review)

sbejaoui:
@pedrobaeza, can you update your review plz

pedrobaeza (approved):
/ocabot merge minor

PR #472, opened 2025-08-13
[16.0][IMP] rma_lot: option to deliver same lot/serial as received

sbejaoui:
Add bool field `same_lot_as_received` on RMA operations to allow delivering or replacing with the exact lot/serial received from the customer. Useful for returning the same physical unit or a repaired item.

### DIFF
--- a/rma_lot/models/rma_operation.py
+++ b/rma_lot/models/rma_operation.py
@@ -1,0 +1,16 @@
+# Copyright 2025 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class RmaOperation(models.Model):
+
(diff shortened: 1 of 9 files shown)


marielejeune (approved):
LGTM

rma_lot/models/rma_operation.py
```diff
@@ -0,0 +1,16 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class RmaOperation(models.Model):
+
+    _inherit = "rma.operation"
+
+    deliver_same_lot = fields.Boolean(
```
rousseldenis, line 11: Shouldn't a default value coming from a configuration be great ?

phschmidt (approved):
Functionally tested

PR #478, opened 2025-09-15
[18.0][MIG] Oca port rma 16.0 improvements to 18.0 

samirGuesmi:


samirGuesmi:
@marielejeune

pedrobaeza:
Please check the CI and do the same for 17.0

sbejaoui (approved):
code review + functional test

sbejaoui:
@pedrobaeza , can you review please?

pedrobaeza:
17.0 already contains these ports?

sbejaoui:
> 17.0 already contains these ports?

I don't believe so

pedrobaeza:
Then this should be ported first to 17.0, and when merged, then this one can be merged, but we can't let one version in the middle orphaned of them.

yvaucher:
@pedrobaeza AFAIK we never enforced the need to forward port to the missing version, some contributors don't use some versions at all.

I agree it would make the ecosystem cleaner and it's a good practice to not skip version.

However, this shows the need for automated tools to forward port of commits.
I don't want to put the burden on the contributors themselves nor block a contrib for this reason.

pedrobaeza:
Well, as maintainer and main contributor of this module, I don't want to have problems on the migration path (I recently faced this exact problem on OCA/l10n-spain by some "selfish" contributors - see https://github.com/OCA/l10n-spain/pull/4453), so if they want this to come to 18, it should come first to 17. I'm not asking for something that requires too much, and it's something reasonable for the ecosystem.

Alternatives are:

- Do these on extension modules, not on the main module, so they can be migrated in any version they want.
- Keep the changes in their branch.

yvaucher:
> Keep the changes in their branch.

This is what we should want to avoid, IMO our aim as maintainer is to make it easier to contribute not the opposite.

Can any contribution be really seen as "selfish"? I prefer a lot of small contributions than a lack of big contributions.

I'm not asking to not ask for the forward port in v17, I'm just saying it shouldn't be a reason to block this.


I must add it is important to have the v18 branch up to date sooner than the v17 otherwise it will also be missing in v19 and it will double the work.

pedrobaeza:
No when merging the contribution means any problem in the migration path. And I say "selfish" (between quotation marks), because they are only seeing their specific case - migration with enterprise and only even versions -, but breaking the flows of the rest, so yes, merging something that is not harmless is not something I want, so I come back to my previous request.

sbidoul:
I agree with Yannick. 

Also at this point it is very possible that nobody will ever need the new features on 17. 

And it is not clear to me how merging this would make the migration more difficult since there is not even a migration script here.

rvalyi:
I will try to make a PR for v17 tonight with the same commits if not present to help fixing this.

rvalyi:
additionally, I believe the "port everything to all intermediate versions" could be a module or repo policy decided by the module authors or repo PSC. Because both policies make sense:

- In general it's better to port everything. Even if there is no migration script involved, another module could extend the feature assuming it has or doesn't have the new features. Or it could be just a compatibility issue.
- Now I totally understand there are repos/modules were the maintainers can not afford porting everything because it's just to much work on their shoulders. For instance, for OCA/l10-brazil we assume and warn that odd branches are just a transition versions that may not receive all the ports from even branches.

Eventually these these policies could be made explicit in the module manifest or repo so people understand what that can reasonably expect without adding load to the maintainers.

sbidoul:
Yes, I'm happy to let maintainers decide. My personal preference goes to getting the new features in the latest available version sooner, though, and backfill intermediate versions if I have time or contributors want to do it.

pedrobaeza:
IMO, that's the direct recipe for not doing that ever.

sbidoul:
> IMO, that's the direct recipe for not doing that ever.

Or doing it only when someone needs it, which is fine in my view.

pedrobaeza:
As shown in the example of OCA/l10n-spain, you may never know if not porting anything is going to cause a problem in the migration path of your customers, so going only to "the need of the feature" is simplistic IMO and it's advocated to get some failures that will require more time than doing the whole fw-port chain.

Note that my team performs always whole fw-port chaining (or even backports, in the supported versions), no matter if we are using them in that version, and it's barely impacting in the global times. Illustrative example: https://github.com/OCA/management-system/pulls?q=is%3Apr+archived%3Afalse+author%3Avictoralmau+is%3Aclosed

sbejaoui:
surpassed by #497 , #495 , #491 , #493

PR #482, opened 2025-10-14
[18.0][FIX] rma_sale: Can create from multiple products

eugenios73:
If you want to create multiple RMAs for multiple products at the same time, you cannot; the system shows you the following error. 

This solution allows you to create RMAs for one or several products at the same time without any issues.

image

image

image


pedrobaeza (changes requested):
- Please don't change manually the module version.
- The problem comes from the source module where it's calling with a multi-recordset. Patch that one.

pedrobaeza (approved):

victoralmau (approved):

PR #483, opened 2025-10-22
[18.0][IMP] rma: new icon

sbejaoui:
This is a proposal for a new icon aligned with Odoo’s latest visual style

image


pedrobaeza:
Thanks for the proposal. Although the icon aligned more with color style, I think the type of icon is not the same as the rest, so it still contrasts with the rest.

sbejaoui:
> Thanks for the proposal. Although the icon aligned more with color style, I think the type of icon is not the same as the rest, so it still contrasts with the rest.

Let's keep this open, maybe other contributors will come up with a better proposal

sbejaoui:
I asked chatGPT

image

image

what do you think @pedrobaeza ?

pedrobaeza (approved):
OK, I think it can serve, thanks.

rousseldenis (approved):
Great !

PR #485, opened 2025-10-22
[18.0][IMP] rma_sale: add name to th&td in sale_portal_template

sbejaoui:
needed to add rma reason before it, already added in https://github.com/OCA/rma/commit/9a3a65b5a2abe91203806ec9100fc95726776cb2

### DIFF
--- a/rma_sale/views/sale_portal_template.xml
+++ b/rma_sale/views/sale_portal_template.xml
@@ -90,10 +90,13 @@
                 <table class="table table-sm" id="request-rma-table">
                     <thead class="bg-100">
                         <tr>
-                            <th class="text-start">Product</th>
-                            <th class="text-end">Quantity</th>
-                            <th class="text-start">Delivery</th>
-                            <th class="text-start">Requested operation</th>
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Please name all the th elements as you are in there. And I know we have talked other times about this, and that it can be resolved as `//th[@name='operation']`, but I prefer to name it `th_operation` instead of just `operation` and have a unique name across whole arch.

sbejaoui:
> Please name all the th elements as you are in there. And I know we have talked other times about this, and that it can be resolved as `//th[@name='operation']`, but I prefer to name it `th_operation` instead of just `operation` and have a unique name across whole arch.

will do

sbejaoui:
> Please name all the th elements as you are in there. And I know we have talked other times about this, and that it can be resolved as `//th[@name='operation']`, but I prefer to name it `th_operation` instead of just `operation` and have a unique name across whole arch.

done

victoralmau (approved):

bjouini-acsone (approved):

PR #486, opened 2025-10-27
[17.0][FWP][IMP] rma_sale: Display the allowed quantity to return in the RMA wizard

sbejaoui:
- Display the allowed quantity to return
- Simplify data encoding by adding a button to select/deselect all lines
- Ensure that the quantity to return is always less than or equal to the allowed quantity

### DIFF
--- a/rma_sale/README.rst
+++ b/rma_sale/README.rst
@@ -78,11 +78,11 @@ Single page RMA request:
 Known issues / Roadmap
 ======================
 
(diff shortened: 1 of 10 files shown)


pedrobaeza:
Forward-port of #407

pedrobaeza (approved):

sbejaoui:
we can merge this one

PR #497, opened 2025-10-30
[18.0][FWP][IMP] rma: Allow manual finish of confirmed RMAs with no further processing required

sbejaoui:
FWP: #478


pedrobaeza (approved):
/ocabot merge minor

Issue #473, opened 2025-08-21
[18.0] RMA Various Enhancements


marielejeune:
Hi all, 

For a quite huge project here at Acsone, we are going to use RMA to manage our returns. The workflow is more complex than what currently exists in this repo, so we are going to provide most of all improvements to the OCA. Here is a list of all features we plan to add, please feel free to give all your suggestions and advice before we start. If you think something is too specific for the OCA, we can also discuss! Don't hesitate also to mention if (part of) a feature is already available.
We would like to focus on it during the OCA days here in Liège in one month, if anyone wants to join to our table.
The improvements will be done on version 18, and based on Souheil Bejaoui's improvements on the 16th version, that we are going to forward-port soon.

### Feature 1 - RMA requests

For now, as far as I know, RMA can only be created from sale orders. 
Our client would like to create RMA requests from scratch. Here are some of their reasons:
* The customer returns a bench of articles at one time, mixing several different orders. 
* They allow their customers to return articles that were buy somewhere else (but they refund less than the sales price of the product).

The solution we propose and we are going to implement is the following: 

1. Create a new model of RMA requests, where one can fill the customer, and a list of products (with quantities) he wants to return, as well as the requested RMA operations. 
2. Ask Odoo to link all RMA request lines to sale order lines were the product were bought. For this we imagine some rules: 
* Define an eligibility period on each RMA operation. 
* Odoo will associate the oldest sale order that is still under the eligibility period of the requested RMA operation
3. As far as all lines are linked to sale order lines and all the needed checks are done, create all the returns from this RMA request.
4. Provide tools to manage all exceptions by hand (Odoo couldn't find a sale order to associate with, or the sale orders were all expired for the eligibility period of the RMA operation,...)

### Feature 2 - Partially refund the customer

Up to now, if the RMA operation has to refund the customer, it will refund 100% of the price at which the product was bought. 
We propose to introduce on the RMA operation a refund percentage, combined with the eligibility period mentioned above: the product will be 100% refundable if still under the eligibility period, and only partially refundable after this period.

### Feature 3 - Allow (or not) a return that isn't linked to a sale order from this customer

As mentioned above, our company wants, under certain circumstances (depending on how big is the customer, what's is the bought product) allow to get back a product even if the customer didn't buy it at their shop. 
Add a configuration to allow or not validation of RMA requests (see Feature 1) if the RMA request line isn't linked to any sale order. Allow to only partially refund (see Feature 2) the customer in this case.

[...]

PR #499, opened 2025-10-31
[18.0][ADD] add rma_batch

sbejaoui:
In many after-sales operations, customers often return **multiple products at once**, sometimes coming from different sales orders or deliveries.

In the base addon **RMA**, each product return creates a separate RMA record. When a customer sends back several items together, this leads to multiple independent RMAs that must be processed, confirmed, and tracked one by one.

This fragmented approach makes it difficult to manage and validate grouped returns, especially for companies handling large volumes of RMAs.

The goal of this module is to introduce an **RMA Batch** — a container that groups several RMAs belonging to the same return event. It allows users to manage all related RMAs together, ensuring that shared information (customer, responsible, team, tags, date) stays synchronized and that batch actions apply consistently across all linked RMAs.

refs: #473


marielejeune (approved):
LGTM. Nice new feature.

Code review only. I plan to do a functional review soon.

marielejeune (approved):

StephaneMangin (approved):

PR #500, opened 2025-10-31
[18.0][ADD] rma_sale_auto_detect

sbejaoui:
In many business flows, a returned product must be linked back to the original sale in order to validate warranty conditions, refunds or exchanges.

Manually searching the correct sale order for each RMA is error-prone and time consuming, especially when:

- the customer has multiple past orders
- the product was delivered in several partial shipments
- the return period depends on the type of RMA operation (refund, warranty, lifetime, etc.)

This module introduces an **automatic matching engine** that links RMA records to the correct delivery moves of the original sale order, based on delivered quantities and eligibility period.

It avoids manual reconciliation and provides a deterministic, auditable match.

refs: #473


rma_sale_auto_detect/models/rma.py
```diff
@@ -0,0 +1,173 @@
...
+            ("state", "in", ["sale", "done"]),
+            ("order_id.date_order", ">=", oldest_date),
+            ("product_id", "=", product.id),
+        ]
+
+    @api.model
+    def _filter_sale_lines_by_delivery_move(self, sale_lines):
+        """get sale line only if the move is not linked"""
+        return sale_lines.filtered(
+            lambda sol: any(m.state == "done" and not m.rma_ids for m in sol.move_ids)
```
marielejeune: Do you only consider delivery moves that arrive to the customer here?

If my delivery is in 2 steps: Stock -> Pack -> Customer, and if the product has been picked but isn't packed yet, won't this method possibly return the related SOL, although the customer hasn't received the product yet?

rma_sale_auto_detect/models/rma.py
```diff
@@ -0,0 +1,173 @@
...
+from odoo import _, api, fields, models
+
+
+class Rma(models.Model):
+    _inherit = "rma"
+    has_sale_auto_detect_issue = fields.Boolean(readonly=True)
+    sale_auto_detect_note = fields.Text(readonly=True)
+
+    def action_link_rma_to_sale_line(self):
+        """automatically link RMAs to the most relevant sale order lines"""
```
marielejeune, line 52: I agree with how the code is written and organized. However I find it quite difficult to understand the logic when reading it.

I've seen the description in the README, but I don't know, shouldn't we add a bit more explanation here in the code? I'm open to discussion

rma_sale_auto_detect/tests/test_rma_sale_auto_detect.py
```diff
@@ -0,0 +1,212 @@
...
+        self.assertEqual(rma2.sale_line_id, sale_order.order_line)
+        self.assertEqual(rma3.sale_line_id, sale_order.order_line)
+        rma4 = self._create_rma(self.partner, self.product, 2, self.operation)
+        rma4.action_link_rma_to_sale_line()
+        # all delivered qty already linked, new rma should not be linked
+        self.assertFalse(rma4.sale_line_id)
+
+    def test_10(self):
+        """When user suggest the sale order and there is multiple delivery moves
+        the rma is split"""
```
marielejeune, line 200: Should we add a test "When user suggest the sale order and there are multiple sale order lines" or is this covered with the multiple delivery moves case?

rma_sale_auto_detect/models/rma.py
sbejaoui: sol.move_ids refer only to the delivery moves, the preparation moves are not directly linked to the sale order line, but to the delivery moves through move_orig_ids
the filter on state == 'done' ensures that we ignore cancelled deliveries and deliveries that have not yet been validated

rma_sale_auto_detect/tests/test_rma_sale_auto_detect.py
sbejaoui: yes, it is already covered by the other tests:

- If no sale line is found, the rma is flagged as having a linking issue
- If multiple sale lines are found, the process iterates line by line until the full rma quantity is satisfied

marielejeune (commented):
LGTM (code review). Functional tests planned soon.

rma_sale_auto_detect/models/rma.py
sbejaoui: Adding the eligibility period to the RMA simplified the code significantly, take another look and let me know what you think now

rma_sale_auto_detect/models/rma.py
```diff
@@ -195,3 +196,26 @@ def action_draft(self):
...
+                < 0
+            ):
+                raise ValidationError(
+                    _(
+                        "The quantity to return exceeds the remaining returnable "
+                        "quantity for this delivery."
+                    )
+                )
+
+        return super().action_confirm()
```
marielejeune, line 232: Is this a feature we want to keep in `rma_sale_auto_detect`? Shouldn't it be interesting in a more generic addon?

sbejaoui: good point, this can be in base module

marielejeune (approved):
LGTM (code + functional review). Working on a test database since several months.

rma_sale_auto_detect/models/rma.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _link_rma_to_sale_line(self, sale_lines):
+        """match between rmas and sale lines"""
+        if not sale_lines:
+            return False
+        sale_line_delivered_qty = self._get_sale_line_returnable_qty(sale_lines)
+        rmas = self.sorted("date")
+        sale_lines = sale_lines.sorted(lambda sol: (sol.order_id.date_order, sol.id))
+
+        rma_index = 0
```
rousseldenis, line 127: @sbejaoui Isn't using iterator better ?

sbejaoui: In this case, we’re doing an asymmetric loop over two sorted lists (sometimes advancing sales, sometimes RMAs, sometimes both), so using indices is easier to read.

Switching to iterators wouldn’t really simplify the code, and there’s no performance gain either, since we’d still need to keep track of the current rma and sale_line, initialize the iterators, and handle exhaustion explicitly.

sbejaoui:
@rousseldenis can you update your review plz

rousseldenis (approved):

PR #501, opened 2025-10-31
[18.0][ADD] rma_batch_sale_auto_detect

sbejaoui:
This module extends the batch RMA process by adding automatic sale order linking at batch level. It reuses the matching logic provided by the module `rma_sale_auto_detect`, but allows it to be executed on all RMAs contained inside an `rma.batch`.

A new batch state **Manual Treatment** is introduced. If at least one RMA in the batch fails to auto-match, the batch is moved to this state, so the user can review and fix the unmatched RMAs manually.

refs: #473


rma_batch_sale_auto_detect/tests/test_rma_batch_sale_auto_detect.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2025 ACSONE SA/NV
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.addons.rma_sale_auto_detect.tests.common import TestRmaSaleAutoDetectBase
+
+
+class TestRmaBatchSaleAutoDetect(TestRmaSaleAutoDetectBase):
```
marielejeune, line 10: Tests fail, I guess the class must be tagged as post-install

rma_batch_sale_auto_detect/tests/test_rma_batch_sale_auto_detect.py
```diff
@@ -0,0 +1,51 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.batch = cls.env["rma.batch"].create({"partner_id": cls.partner.id})
+
+    def _create_rma(self, partner, product, qty, operation):
+        rma = super()._create_rma(partner, product, qty, operation)
+        rma.batch_id = self.batch
```
marielejeune, line 18: What happens if you create a RMA batch for partner 1 but you add RMAs for both partner 1 and partner 2 inside this batch?

I've seen there is a compute on partner_id on `rma` model, to fill it with the partner of the batch. But what if the link is done in the other way: setting the partner on the RMA first, then adding the RMA to the batch?

sbejaoui: I will add a constraint to prevent users from mixing rmas from different partners in the same batch

sbejaoui: done in #499

rma_batch_sale_auto_detect/tests/test_rma_batch_sale_auto_detect.py
```diff
@@ -0,0 +1,54 @@
...
+        return rma
+
+    def test_0(self):
```
StephaneMangin, line 21: Maybe rename the method to be clearer about what it does with a quick docstring for the behavior's details ?

rma_batch_sale_auto_detect/tests/test_rma_batch_sale_auto_detect.py
```diff
@@ -0,0 +1,54 @@
...
+        rma2 = self._create_rma(self.partner, self.product2, 5, self.operation)
+        self.batch.action_ready()
+        self.assertEqual(self.batch.state, "manual")
+        self.assertTrue(rma.sale_line_id)
+        self.assertFalse(rma2.sale_line_id)
+        rma2.move_id = sale_order2.order_line.move_ids
+        self.batch.action_ready()
+        self.assertEqual(self.batch.state, "ready")
+
+    def test_1(self):
```
StephaneMangin, line 40: Maybe rename the method to be clearer about what it does with a quick docstring for the behavior's details ?

StephaneMangin (approved):
Non blocking comment added

StephaneMangin:
Preapproved. But the tests issues comes from a potentiel missing/invalid dependency.

rousseldenis (approved):
Code and functional review

marielejeune:
@sbejaoui could you have a look at non-blocking comments from Stéphane, and then we can merge this PR?

PR #502, opened 2025-11-01
[18.0][FIX] rma_sale: avoid wrong SO link when grouping receptions

sbejaoui:
ensure reception pickings are grouped even when RMAs come from different
sale orders, but without assigning the sale order to the procurement group
this prevents unexpected behavior where the return picking could be wrongly
linked to a sale order

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -166,7 +166,10 @@ class Rma(models.Model):
 
     def _prepare_procurement_group_vals(self):
         vals = super()._prepare_procurement_group_vals()
-        if not self.env.context.get("ignore_rma_sale_order") and self.order_id:
+        if (
+            not self.env.context.get("ignore_rma_sale_order")
+            and len(self.order_id) == 1
+        ):
             vals["sale_id"] = self.order_id.id
         return vals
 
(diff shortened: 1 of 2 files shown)


pedrobaeza (commented):
Although I understand the issue with returned quantities, not sure about this, as you may receive everything together. In fact, we have a specific case where this is happening: they make several order per month, and at the end, they return a quantity that may match more then one order. Don't we have other way to handle all the procurement group thing?

sbejaoui:
> Although I understand the issue with returned quantities, not sure about this, as you may receive everything together. In fact, we have a specific case where this is happening: they make several order per month, and at the end, they return a quantity that may match more then one order. Don't we have other way to handle all the procurement group thing?

It's a choice we need to make, if we want the module to group receptions from different sale orders we need detect this case and don't link the procurement group to the any of the sale orders

https://github.com/OCA/rma/blob/733adffcd944af92734280dba0ad8205b7821857/rma_sale/models/rma.py#L167-L171

this should be len(self.sale_order_id) == 1

pedrobaeza:
OK, that can be a good balance.

marielejeune (approved):
LGTM (code review)

pedrobaeza (approved):

PR #467, opened 2025-08-06
[16.0][IMP] BP rma_repair & align repair creation with other operation-level actions

sbejaoui:
- backport rma_repair
- align repair creation with other operation-level actions via new `action_create_repair` field:
  - manual_on_confirm
  - manual_after_receipt
  - automatic_on_confirm
  - automatic_after_receipt

- add `can_be_repaired` to control repair button visibility
- extend confirm and receipt flows to support automatic repair creation

PR #503, opened 2025-11-12
[18.0][MIG] rma_repair + rma_repair_lot: Migration to 18.0

victoralmau:
Migration to 18.0 + separate `rma_repair` into `rma_repair_lot` to avoid such a hard dependency (`rma_lot`).

Please @pilarvargas-tecnativa and @christian-ramos-tecnativa can you review it?

@Tecnativa TT57882


victoralmau:
I think the `rma_lot` dependency in https://github.com/OCA/rma/blob/18854fcc884e7d6b4926256f732b6469afdf97a5/rma_repair/models/rma.py#L58 is too strict (especially because of the dependencies that `rma_lot` has), so I can think of different possibilities:
- Do not add `rma_lot` as a dependency of `rma_repair` and if the `lot_id` field exists, define its value.
- Do not add `rma_lot` as a dependency of `rma_repair` and add an extra module (`rma_repair_lot`, for example) to add only that case.

What do you think, @pedrobaeza ?

rma_repair/models/rma.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2024 APSL-Nagarro Antoni Marroig
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, fields, models
```
eduezerouali-tecnativa:
```suggestion
from odoo import api, fields, models
```

pedrobaeza:
I would go for an extra module.

victoralmau:
> I would go for an extra module.

On second thought, I don't think it's worth the effort to separate that part into a new module just for two lines of code. In the future, the separation will have to be justified, and anyone using this module in previous versions would not be able to use it and would have to add the new one.

In any case, it will be added to the ROADMAP.

sbejaoui:
can you include these two commits please: 
- https://github.com/OCA/rma/pull/467/commits/c424d4f7fdf5a22af1acc2ef6ac577985a3d4755
- https://github.com/OCA/rma/pull/467/commits/f30284a206ae6829aef23a464035e085a6cd3ec2

christian-ramos-tecnativa:
> https://github.com/OCA/rma/blob/18854fcc884e7d6b4926256f732b6469afdf97a5/rma_repair/models/rma.py#L58

IMO if this is the only reason to add `rma_lot` to the dependencies I would prefer checking if the field exists directly, as  @victoralmau said, `rma_lot` adds many modules from `stock-logistics-workflow` unnecessary if you don't work with lots.

victoralmau:
FWP of changes https://github.com/OCA/rma/pull/503#issuecomment-3531837234 created at 17.0: https://github.com/OCA/rma/pull/507

Commits added from 17.0 (https://github.com/OCA/rma/pull/507) to this PR

Separated the `rma_lot` part into a new `rma_repair_lot` module to avoid that hard dependency

christian-ramos-tecnativa:
Shouldn't be `rma_repair_lot` autoinstalable?

victoralmau:
> `rma_repair_lot`

I'm not sure about auto-install. What do you think, @pedrobaeza ?

pedrobaeza:
Yeah, it sounds correct.

christian-ramos-tecnativa (approved):
LGTM

victoralmau:
Ping @pedrobaeza

pedrobaeza (approved):
/ocabot merge nobump

PR #512, opened 2025-12-10
[18.0][MIG] rma_delivery: Migration to 18.0

victoralmau:
Migration to 18.0

@Tecnativa TT57139


victoralmau:
Ping  @pedrobaeza and @carlos-lopez-tecnativa

pedrobaeza (approved):
Code review. The dependency change to `stock_delivery` is reasonable.

rma_delivery/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="res_config_settings_view_form" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="stock.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <block name="operations_setting_container" position="inside">
+                <setting id="rma_delivery_strategy" title="RMA delivery strategy">
```
carlos-lopez-tecnativa, line 8: I think it would be better to group all settings related to `RMA` into their own group or section. Why don’t you move this section to RMA, or, within Inventory, add these options in a new section instead of adding them to the `Operations` section?
image

rma_delivery/views/rma_views.xml
```diff
@@ -0,0 +1,24 @@
...
+                <field
+                    name="carrier_id"
+                    invisible="rma_delivery_strategy!='rma_method'"
+                    readonly="not show_create_refund and not show_create_replace and not show_create_return"
+                />
+                <field
+                    name="reception_carrier_id"
+                    invisible="rma_reception_strategy!='rma_method'"
+                    readonly="state!='draft'"
+                />
```
carlos-lopez-tecnativa, line 20: When you create the RMA from a sale, it is not possible to set this field because the RMA is in the `Confirmed` state and is read-only.
image

So, I think you need to ask the user for the `reception_carrier_id` field in the wizard, or this field can never be set from a sale and only when the RMA is created from its own menu. What do you think?
image

carlos-lopez-tecnativa (changes requested):

rma_delivery/views/res_config_settings_views.xml
victoralmau: This is the same as in v16, and although I also find it strange (I think it would be more appropriate to have it in the RMA section), I have chosen to keep it in the same place to avoid any confusion.

rma_delivery/views/rma_views.xml
victoralmau: I understand what you're saying and I agree, **BUT** the wizard you mention adds `rma_sale`, and to add that field in that wizard, it would have to be added as a dependency of this module (`rma_sale`), which I think would be incorrect. Therefore, the right thing to do would be to do that in a new module (auto_install) called `rma_delivery_sale`. What do you think @pedrobaeza

pedrobaeza: When being done from a sales order, the reception picking is already created, being confirmed, isn't it? The option is to put the carrier on that picking directly, or to add an option for creating the RMAs from sales in draft state.

victoralmau: I think it is important to clarify a few things:

- `rma_delivery` does **NOT** depend on `rma_sale`, therefore, the wizard for creating an RMA from sales does not exist (unless `rma_sale` is installed).
- The `reception_carrier_id` field is added to the RMA because it is also possible to create an RMA manually.
- The RMA reception picking is generated when the RMA is in confirmed status (this is when `reception_carrier_id` is used).
- The RMA will be automatically changed to confirmed as defined in the linked operation (`operation_id`): https://github.com/OCA/rma/blob/6a13fdd94f6bd5b94d1895559a78468f31c65009/rma/models/rma_operation.py#L37C5-L37C27

All of the above explains the process and the probable need for `rma_delivery_sale` to exist and for the `reception_carrier_id` field to be added to the sales order wizard.

pedrobaeza: The terminology is a bit confused: confirm the reception means to validate the picking or to create the picking? Anyway, I have already expressed my PoV in the previous comment, and it's not to create a new glue module. What do you think about it?

victoralmau: Confirming the RMA will simply create the delivery picking. No process (that I am aware of, at least) automatically done the delivery picking.

I do NOT understand exactly what you are proposing in https://github.com/OCA/rma/pull/512#discussion_r2682998868 (remember that the RMA creation wizard from a sales order only exists if `rma_sale` is installed, and this module does not depend on that module).

carlos-lopez-tecnativa: > Confirming the RMA will simply create the delivery picking. No process (that I am aware of, at least) automatically done the delivery picking.

The field `auto_confirm_reception` that you referred to automatically validates (marks as done) the reception picking.
https://github.com/OCA/rma/blob/6a13fdd94f6bd5b94d1895559a78468f31c65009/rma/models/rma.py#L789-L791

victoralmau: There are two different fields:

- The `action_create_receipt` (https://github.com/OCA/rma/blob/6a13fdd94f6bd5b94d1895559a78468f31c65009/rma/models/rma_operation.py#L24) field will auto-confirm the RMA and create the reception picking.
- The `auto_confirm_reception` (https://github.com/OCA/rma/blob/6a13fdd94f6bd5b94d1895559a78468f31c65009/rma/models/rma_operation.py#L37) field will auto-validate the reception picking.

This is something that exists in RMA, and I'm not sure exactly what you're proposing to change with this. Obviously, if the reception picking has already been created, the `reception_carrier_id` field cannot be changed, and if the reception picking has been done, it cannot be changed either.

pedrobaeza: Then I insist to keep it this way. If someone wants to modify something, they have to unmark the RMA auto-confirm, and when they generate the RMA from the sales order, it's kept in draft, they change whatever, and then confirm.

carlos-lopez-tecnativa: Sorry, but this is not working as expected.

My settings in the `Repair` operation:
image

When I create an RMA from a sale, the receipt is not created, which is OK. However, I cannot set the `Reception Carrier` field because the RMA is in the confirmed state.
image

So, my suggestion is that this field can be changed if the receipt does not exist yet. What do you think about this change?

```suggestion
                    name="reception_carrier_id"
                    invisible="rma_reception_strategy!='rma_method'"
                    readonly="state not in ['draft', 'confirmed'] or reception_move_id "
                />
```

carlos-lopez-tecnativa: > Then I insist to keep it this way. If someone wants to modify something, they have to unmark the RMA auto-confirm, and when they generate the RMA from the sales order, it's kept in draft, they change whatever, and then confirm.

Note that an RMA created from a sale is always confirmed. Therefore, I propose allowing the field to be edited if `reception_move_id` has not been created yet. This change covers my use case without adding a dependency on `rma_sale` or creating a glue module.
https://github.com/OCA/rma/blob/ac34475a3a909929c3682d142931f863d4fef257/rma_sale/wizard/sale_order_rma_wizard.py#L98

carlos-lopez-tecnativa (approved):
LGTM

PR #515, opened 2025-12-22
[18.0][IMP] rma: Add _get_location_final() method to be able to use it in other modules

victoralmau:
Add `_get_location_final()` method to be able to use it in other modules

@Tecnativa TT57139

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -1259,6 +1259,13 @@ class Rma(models.Model):
         vals["move_orig_ids"] = [(6, 0, self.reception_move_id.ids)]
         return vals
 
+    def _get_location_final(self):
+        """Similar to the _get_location_final() method in sale_stock.
+        The method represents the location_final_id field that will be defined
+        when run procurements; the location_dest_id field in stock move is a compute."""
+        self.ensure_one()
+        return self.partner_shipping_id.property_stock_customer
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Shouldn't this be called `_get_final_location_dest`?

victoralmau:
> Shouldn't this be called `_get_final_location_dest`?

Okay, changed. The name of the method `_get_location_final()` was similar to the one that exists in `sale_stock`: https://github.com/odoo/odoo/blob/aad8a4a25f5d251b48e4f1894ee113176abdc834/addons/sale_stock/models/sale_order_line.py#L308

pedrobaeza:
Ah, OK, didn't know that analogy. You can comment that on the commit message/code for knowing about your inspiration. Yet, I think Odoo didn't choose the best name for it, don't you think? Anyway, if you think it's better to put the same name, just add the comment. What you prefer.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

victoralmau:
IMO the name `_get_location_final()` is not incorrect, because it represents reality, the value that will be defined in `location_final_id`. I have added the appropriate comment in the method to clarify this.

PR #524, opened 2026-01-13
[16.0][IMP] rma : send receipt notification when receiving product in waiting_replacement state 

nihelgabsi-acsone:
This improvement ensures that the receipt notification email is sent to the customer when a product is received for an RMA that is already in the waiting_replacement state.

**Problem:**
When an RMA operation is configured to create the delivery order automatically on confirmation, the RMA moves directly to the waiting_replacement state without passing through the received state, because the delivery order is created immediately. As a result, when the product is later received, the receipt notification email is not sent to the customer.

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -344,6 +344,7 @@ class Rma(models.Model):
         "without requiring further processing such as a receipt, "
         "delivery, or refund.",
     )
+    receipt_confirmation_email_sent = fields.Boolean(default=False)
 
     @api.depends("operation_id", "reception_move_id.state")
     def _compute_manual_finish_allowed(self):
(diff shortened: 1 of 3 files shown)


nihelgabsi-acsone:
@sbejaoui

rma/models/rma.py
```diff
@@ -344,6 +344,8 @@ def _domain_location_id(self):
         "without requiring further processing such as a receipt, "
         "delivery, or refund.",
     )
+    # Used in unit tests to verify that the confirmation email has been sent
```
sbejaoui:
```suggestion
```

sbejaoui (changes requested):
use this field in _send_receipt_confirmation_email to avoid re-sending  the mail

rma/models/rma.py
nihelgabsi-acsone: The product can only be received once. This field is more useful for testing. I can remove the comment if it's useless

PR #525, opened 2026-01-15
[18.0][MIG] rma_sale_mrp: Migration to version 18.0

eduezerouali-tecnativa:
cc @Tecnativa TT57276

ping @victoralmau @pedrobaeza 

Depends on:

- [x] https://github.com/OCA/rma/pull/547


rma_sale_mrp/views/sale_order_portal_template.xml
```diff
@@ -34,7 +34,7 @@
                 name="t-att-value"
             >data.get('phantom_bom_product') and data['quantity'] or 0</attribute>
         </xpath>
-        <!-- TODO: We could give a clue about what's to be returned, with readonly detailed lines -->
+        <!-- TODO: We could give a clue about what's to be returned, with readonly detailed lines
```
victoralmau: Why is it necessary to comment on this now? If it is not necessary, the corresponding code should be removed and explained.

victoralmau (changes requested):

victoralmau (changes requested):
It is not possible to select kit products on the portal.

**v16**
Image

**v18**
Image

mariancuadranetkia:
It's not possible create RMA of a Kit, because we receive next error:
imagen

I'm not sure if this error could be because the 'delivery order' field isn't automatically filled in the kits, and that's why the validation is being skipped.:
imagen

eduezerouali-tecnativa:
@victoralmau sorry for the late replay, now it is working as it should.
image

eduezerouali-tecnativa:
@mariancuadranetkia first thank you to take the time for the review. Could you give more detail on how you got the error. Anyways i think it has to do with not valid the picking.

mariancuadranetkia:
@eduezerouali-tecnativa 
I created a sale with table kit and I finished picking.
After that, I returned to sale order, and I created RMA. Although I could choose product kit to create RMA, when I 'accept' to create it, I got error.
I attach a short video, where you can see the sale order and picking, and error when I try to create RMA. This sale order is created in order to test this feature.

https://github.com/user-attachments/assets/ebfc86cd-0178-46c8-a435-1730cefe4b8c

eduezerouali-tecnativa:
@mariancuadranetkia the main problem is that you are using a kit with one product that it is not trackable. Try again creating your own products and components and you will see that RMA is create correctly.

mariancuadranetkia:
> @mariancuadranetkia the main problem is that you are using a kit with one product that it is not trackable. Try again creating your own products and components and you will see that RMA is create correctly.

Mmm yes, but, this way customer can't create RMA to ask for a refund if a product is not trackable or a component of kit is not trackable? Because, I can create a return picking without RMA without problem of this product not trackable, problem it's only creating RMA.

I'm not sure if this is the expected behaviour, but, if I can return a product and I can refund it, I don't understand the reason to not allow create RMA to register this operation.

victoralmau:
Reviewing the behavior in v16, it should work correctly, i.e., even if the product kit (table kit, for example) does not have tracking set, if there are stock moves for the kit components, it should allow the corresponding RMA(s) to be created.

eduezerouali-tecnativa:
@victoralmau I have check on v18 to do a RMA from Sale Order and it is not possible to do it if the product is not `is_storable`

eduezerouali-tecnativa:
@mariancuadranetkia you were right it is fix in this PR https://github.com/OCA/rma/pull/547

mariancuadranetkia:
@eduezerouali-tecnativa ok, I tested this PR and I validated.
But I can't test this fix yet with kits, can I? I've tested in this runbot with same Kit but I have same error, I think that I should wait but I'm not sure.
Thanks!

eduezerouali-tecnativa:
@mariancuadranetkia @victoralmau rebase done, now it should work as expected.

mariancuadranetkia:
@eduezerouali-tecnativa I've checked, and, with new sales order if I create RMA from backend,, it works.
But, if I duplicate sales order, and I try to create RMA from e-commerce, I receive next error:
imagen

imagen



Thanks!!!

eduezerouali-tecnativa:
@mariancuadranetkia this is a clear bug. It is amend here https://github.com/OCA/rma/pull/548.
Thank you for your review.

pedrobaeza:
Please update this one when the other PR is merged to make use of the new method.

eduezerouali-tecnativa:
I will do, plus include this commit 367e60affe07f7426f71f0ed89e846249a553c19

mariancuadranetkia (approved):
Functional review ok!
Now create RMA with Kits or Consu products from backend/e-commerce is working fine.

victoralmau (approved):

PR #528, opened 2026-01-23
[18.0][IMP] rma + rma_reason: Remove the RMA text from some menus as it is redundant

victoralmau:
Remove the RMA text from some menus as it is redundant

If we are in the RMA > Settings > Teams menu, there is no need for that menu to be RMA Teams

Please @carlos-lopez-tecnativa and @pedrobaeza can you review it?

@Tecnativa

### DIFF
--- a/rma/i18n/de.po
+++ b/rma/i18n/de.po
@@ -713,8 +713,8 @@ msgstr "Erstellen Sie einen neuen RMA-Abschluss"
 
 #. module: rma
 #: model_terms:ir.actions.act_window,help:rma.action_rma_tag
-msgid "Create a new RMA tag"
-msgstr "Erstellen Sie ein neues RMA-Tag"
+msgid "Create a new tag"
+msgstr "Erstellen Sie ein neues Tag"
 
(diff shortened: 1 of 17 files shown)


sbejaoui (approved):

pedrobaeza (commented):
Please adapt also the translations.

pedrobaeza (commented):
Aren't translations like "RMA teams" still pending to adapt?

pedrobaeza (commented):
It still seems inconsistent, as now team is plural and .pot only contains one change.

victoralmau:
> It still seems inconsistent, as now team is plural and .pot only contains one change.

RMA Tags by Tags has not been changed in the translation files because Tags already exists (https://github.com/OCA/rma/blob/3d53d8d31be115650e5ee09b5bf8a628db819d79/rma/i18n/rma. pot#L2106), so when this is merged, the bot will update the translation files to put everything in the right place.

pedrobaeza (approved):
/ocabot merge patch

PR #529, opened 2026-01-23
[18.0][ADD] rma_sale_delivery: New module

victoralmau:
New module

Locked by:
- [x] `rma_delivery`: https://github.com/OCA/rma/pull/512

@Tecnativa TT57139


victoralmau:
Ping @carlos-lopez-tecnativa and @pedrobaeza

rma_sale_delivery/wizard/sale_order_rma_wizard.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    reception_carrier_id = fields.Many2one(
+        comodel_name="delivery.carrier", string="Reception Carrier"
+    )
+
+
+class SaleOrderLineRmaWizard(models.TransientModel):
+    _inherit = "sale.order.line.rma.wizard"
+
+    reception_carrier_id = fields.Many2one(
```
pedrobaeza: Why put it at line level, being able to read it from the header?

pedrobaeza (approved):

rma_sale_delivery/wizard/sale_order_rma_wizard.py
victoralmau: Although it could be set in the header (and it is transferred to the lines), I had considered the possibility of defining a different carrier for each line (each rma to be created), but on second thought, it is somewhat “unusual” or uncommon, so I will leave the data only in the header.

carlos-lopez-tecnativa (approved):
LGTM

PR #530, opened 2026-01-26
[18.0][IMP] rma: Add the appropriate depends to the compute methods + Possibility of return or replace if it has already been received and the operation is set to Manually on Confirm

victoralmau:
Add the appropriate depends to the compute methods.

Possibility of return or replace if it has already been received and the operation is set to Manually on Confirm.

Steps to reproduce:
- Modify an operation and set Delivery Action=Manually on Confirm
- Create an RMA and generate reception picking
- Validate the reception picking
- The RMA does not have the Return to customer or Replace button

@Tecnativa TT60352

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -394,7 +394,7 @@ class Rma(models.Model):
         for r in self:
             r.remaining_qty = r.product_uom_qty - r.delivered_qty
 
-    @api.depends("state")
+    @api.depends("state", "operation_id", "operation_id.action_create_refund")
     def _compute_can_be_refunded(self):
         """Compute 'can_be_refunded'. This field controls the visibility
         of 'Refund' button in the rma form view and determinates if
@@ -411,7 +411,9 @@ class Rma(models.Model):
                 and record.state == "confirmed"
             )
 
-    @api.depends("remaining_qty", "state", "operation_id.action_create_delivery")
+    @api.depends(
+        "remaining_qty", "state", "operation_id", "operation_id.action_create_delivery"
+    )
     def _compute_can_be_returned(self):
         """Compute 'can_be_returned'. This field controls the visibility
         of the 'Return to customer' button in the rma form
@@ -434,7 +436,7 @@ class Rma(models.Model):
                 )
             )
 
-    @api.depends("state")
+    @api.depends("state", "operation_id", "operation_id.action_create_delivery")
     def _compute_can_be_replaced(self):
         """Compute 'can_be_replaced'. This field controls the visibility
         of 'Replace' button in the rma form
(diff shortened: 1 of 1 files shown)


rma/models/rma.py
```diff
@@ -394,7 +394,7 @@ def _compute_remaining_qty(self):
         for r in self:
             r.remaining_qty = r.product_uom_qty - r.delivered_qty
 
-    @api.depends("state")
+    @api.depends("state", "operation_id", "operation_id.action_create_refund")
```
pedrobaeza, line 397: The third one shouldn't be put, as it will compute retroactively for all the existing RMAs if you change anything in the operation (the same for all).

pedrobaeza:
This PR contains 2 different things, so at least the title should reflect it.

rma/models/rma.py
victoralmau: I disagree, mainly because if you decide to change that setting in operations, those changes are not reflected in RMAs that have already been created; furthermore, the fields `can_be_refunded`, `can_be_returned`, and `can_be_replaced` are not store.

Furthermore, some of these computes already have the corresponding field of the operation as a depend (https://github.com/OCA/rma/blob/3d53d8d31be115650e5ee09b5bf8a628db819d79/rma/models/rma.py#L414), so I think that to be consistent, it should be the same for all of them, don't you think?

pedrobaeza: No, that's catastrophic! It's the same as if you put on `sale.order.line` that `qty_to_invoice` should depend on `product.invoicing_policy`. That means that if you change the invoicing policy, all the previous orders that you have already invoiced and finished will be recomputed (apart from the computation times), putting a lot of orders pending to be invoiced suddenly.

The thumb rule is that a configuration setting should always affect only from that moment onward, not from the past thing. Any previous error should be fixed manually.

If there are other fields with that depends, they should be removed IMO.

Let's involve @sbejaoui here, that was the original author of this part.

victoralmau: The `qty_to_invoice` field in `sale.order.line` is a compute store=True field, so it is NOT the same example.

If all this could generate debate, perhaps the best thing to do would be to “move” that commit to another PR so we can continue with the other issue. Do you agree?

pedrobaeza: If the field is not stored, then it's a different story, as we are only talking about records in cache, so OK, let's continue this way.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
I am not sure about this `depends` `operation_id.action_create_delivery`, because this can recompute all RMAs that use this operation type. In a database with many records, this could be slow, but I assume this is the expected behavior.

victoralmau:
This is ready to merge.

sbejaoui (approved):
LGTM
At first sight, I wasn’t sure about the dependency on operation_id.action_create_refund, but since the fields are not stored, it’s fine

Thanks

PR #532, opened 2026-02-05
[18.0][IMP] rma: Improve delivery_picking_count (include next transfers)

victoralmau:
Improve `delivery_picking_count` (include next transfers)

@Tecnativa TT60705

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -345,7 +345,10 @@ class Rma(models.Model):
 
     def _compute_delivery_picking_count(self):
         for rma in self:
-            rma.delivery_picking_count = len(rma.delivery_move_ids.picking_id)
+            moves = rma.delivery_move_ids | self.env["stock.move"].browse(
+                rma.delivery_move_ids._rollup_move_dests()
+            )
+            rma.delivery_picking_count = len(moves.picking_id)
 
     @api.depends(
         "delivery_move_ids",
@@ -1006,7 +1009,10 @@ class Rma(models.Model):
 
     def action_view_delivery(self):
         """Invoked when 'Delivery' smart button in rma form view is clicked."""
-        return self._action_view_pickings(self.mapped("delivery_move_ids.picking_id"))
+        moves = self.delivery_move_ids | self.env["stock.move"].browse(
+            self.delivery_move_ids._rollup_move_dests()
+        )
+        return self._action_view_pickings(moves.mapped("picking_id"))
 
     # Validation business methods
     def _ensure_required_fields(self):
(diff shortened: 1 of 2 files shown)


rma/models/rma.py
```diff
@@ -344,7 +344,9 @@ def _compute_show_refund_replace(self):
 
     def _compute_delivery_picking_count(self):
         for rma in self:
-            rma.delivery_picking_count = len(rma.delivery_move_ids.picking_id)
+            moves = rma.delivery_move_ids
+            moves += moves.move_dest_ids
```
carlos-lopez-tecnativa: What happens if you have more delivery steps? This code only works for up to two steps, but what if you configure the route or the warehouse with more steps?

I think this could be recursive to get the move_dest_ids by using this function [_rollup_move_dests](https://github.com/odoo/odoo/blob/652dc7c7f8e7028a6b8ac3a2036b8d1c700d2193/addons/stock/models/stock_move.py#L2451C9-L2451C27) . What do you think?

victoralmau: It returns OrderedSet(), so you would have to do some work after the call; I don't see any module that does that.

Another possible approach (I didn't want to prepare it and open up a rather “complex” debate) would be to use all the movements from the same `procurement.group` by defining an `rma_id` field, but that change would be much more complex.

carlos-lopez-tecnativa: > It returns OrderedSet(), so you would have to do some work after the call; I don't see any module that does that.
> 

I don’t understand your point. If this returns an `OrderedSet`, you just need to call `browse`() afterward. This method is used in other places.
image

> Another possible approach (I didn't want to prepare it and open up a rather “complex” debate) would be to use all the movements from the same `procurement.group` by defining an `rma_id` field, but that change would be much more complex.

I imagine this approach could include returns as well, right?

victoralmau: Ok, done.

rma/models/rma.py
```diff
@@ -1006,7 +1009,9 @@ def action_view_refund(self):
 
     def action_view_delivery(self):
         """Invoked when 'Delivery' smart button in rma form view is clicked."""
-        return self._action_view_pickings(self.mapped("delivery_move_ids.picking_id"))
+        moves = self.delivery_move_ids
+        moves += moves.move_dest_ids
```
carlos-lopez-tecnativa: I think you need to apply the same logic in the count here; otherwise, you may display a number of deliveries, and when clicking the smart button to open them, a different number of records may be shown, which can cause confusion.

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

PR #534, opened 2026-02-10
[18.0][ADD] rma_batch_reason: add new module to link batch and reason behaviors

StephaneMangin:


StephaneMangin:
Dependency on https://github.com/OCA/rma/pull/499

StephaneMangin:
Failed tests due to an indirect dependency to rma_sale which is not on this module dependency scope but in  the repository.

yvaucher:
For the tests I recommend to test if `sale` module might be installed and thus provide `sale_line_warn` value in that case to improve robustness of the tests.

StephaneMangin:
> For the tests I recommend to test if `sale` module might be installed and thus provide `sale_line_warn` value in that case to improve robustness of the tests.

Without the test-requirements for rma_batch which has just been merged, everything passed green !

yvaucher (approved):

rma_batch_reason/models/rma_batch.py
```diff
@@ -0,0 +1,24 @@
...
+    _inherit = "rma.batch"
+
+    reason_id = fields.Many2one(
+        "rma.reason",
+        string="Reason",
+        help="Reason for the return, applied to all RMAs in the batch.",
+    )
+
+    @api.onchange("reason_id")
+    def _onchange_reason_id(self):
```
rousseldenis, line 17: @StephaneMangin I would have used computed/stored/readonly=False field instead in order to avoid onchanges implementation.

PR #539, opened 2026-02-12
[18.0][FIX] rma + rma_sale: Remove unnecessary invisible field

victoralmau:
Remove unnecessary invisible field

Also, if it exists in the view, it should have `column_invisible` defined.

@Tecnativa


pedrobaeza (commented):
I think in this case, they can be directly removed due to viewtiverse

pedrobaeza (approved):
/ocabot merge patch

PR #540, opened 2026-02-12
[17.0][FIX] rma: Change to column_invisible so that the fields are not displayed

victoralmau:
Change to `column_invisible` so that the fields are not displayed

@Tecnativa


pedrobaeza (approved):
Another direct clear one

/ocabot merge patch

PR #544, opened 2026-03-02
[18.0][IMP] rma: Define the expected configuration (action_create_* fields) in operations

victoralmau:
Define the expected configuration (action_create_* fields) in operations

Replace: Not refund
Repair: Not refund
Refund: Not delivery

Add tracking=True to operation_id field + Do not allow the operation_id field to be modified if it is already "completed"

@Tecnativa

### DIFF
--- a/rma/data/rma_operation_data.xml
+++ b/rma/data/rma_operation_data.xml
@@ -2,11 +2,14 @@
 <data noupdate="1">
     <record id="rma_operation_replace" model="rma.operation">
         <field name="name">Replace</field>
+        <field name="action_create_refund" />
     </record>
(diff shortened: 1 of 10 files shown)


pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

I think the `operation_id` field should be read-only when the RMA flow is finished. Currently, I can change this field, which can be confusing, and no trace is left in the chatter. Please consider adding tracking to the field or making it read-only at some point in the flow.

Image

carlos-lopez-tecnativa (approved):

PR #547, opened 2026-03-09
[18.0][FIX] rma: create rma when product is consu

eduezerouali-tecnativa:
In v17 `_product_is_storable` checked the product type (`consu` or `product`)
to determine whether an RMA could be created.

In v18 the `type` field was removed, causing the logic to only allow
`is_storable` products (previously `product`). As a result, consumable
products (`consu`) could no longer be used to create RMAs.

This change restores the previous behavior by allowing consumable
products as well when creating RMAs.

cc @Tecnativa
@pedrobaeza @victoralmau 

affects: https://github.com/OCA/rma/pull/525

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -707,6 +707,10 @@ class Rma(models.Model):
             "context": ctx,
         }
 
+    def _product_is_storable(self, product=None):
+        product = product or self.product_id
+        return product.type == "consu"
+
     def _add_message_subscribe_partner(self):
         self.ensure_one()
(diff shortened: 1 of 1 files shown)


pedrobaeza (commented):
Please restore this v17 method:

https://github.com/OCA/rma/blob/17.0/rma/models/rma.py#L714

and check everywhere where it was used in 17.0 and replace it.

pedrobaeza (approved):

mariancuadranetkia (approved):
LGTM!

victoralmau (approved):

PR #548, opened 2026-03-10
[18.0][IMP] rma_sale: improve how to get allowed_quantity

eduezerouali-tecnativa:
This change is made for `rma_sale_mrp`, because components are not in sale order. In that case we can override that method to get the correct quantity. I think that getting `allowed_quantity` from `qty_delivery` is right for products, but not for components.

Related: https://github.com/OCA/rma/pull/525
Anyways this should be backport, I have check in v16 and it is the same.

cc @Tecnativa 
ping @pedrobaeza @victoralmau

### DIFF
--- a/rma_sale/controllers/sale_portal.py
+++ b/rma_sale/controllers/sale_portal.py
@@ -51,7 +51,7 @@ class CustomerPortal(CustomerPortal):
                 custom_vals.update({name: value})
         for vals in mapped_vals.values():
             sale_line = order_line_obj.browse(vals.get("sale_line_id")).sudo()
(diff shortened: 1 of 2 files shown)


mariancuadranetkia (approved):
LGTM!

victoralmau:
Can you check version 15.0? In that version, rma_sale_mrp works correctly.

eduezerouali-tecnativa:
@victoralmau did check works as expected. This https://github.com/OCA/rma/commit/822c0462de30662015ac94702fde5aa434d6c43b#diff-500e4b73533a6c2514bec67539d038959296196a460e354e3e55a7b18d6a95ab change was only apply from 16 to 18.

PR #553, opened 2026-03-18
[18.0][WIP] rma: Add RECEIVED_STATES to enable inheritance

victoralmau:
Add RECEIVED_STATES to enable inheritance

If other modules add specific statuses that also allow for Refund/Replace/Return to customer, it is necessary to have a mechanism that allows for extending other statuses

@Tecnativa TT61554


pedrobaeza (commented):
I think the way to extend this is not through a Python variable that is not multi-tenant resilient, but simply overriding the `_compute_can_be_refunded` and adding the new condition.

PR #554, opened 2026-03-18
[18.0][FIX] rma_sale_lot: Add domain to the lot_id field so that only the appropriate lots can be selected

victoralmau:
Add domain to the `lot_id` field so that only the appropriate lots can be selected

@Tecnativa TT61559

### DIFF
--- a/rma_sale_lot/wizards/sale_order_rma_wizard.py
+++ b/rma_sale_lot/wizards/sale_order_rma_wizard.py
@@ -1,4 +1,5 @@
 # Copyright 2024 ACSONE SA/NV
+# Copyright 2026 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import api, fields, models
@@ -18,8 +19,20 @@ class SaleOrderRmaWizard(models.TransientModel):
 class SaleOrderLineRmaWizard(models.TransientModel):
     _inherit = "sale.order.line.rma.wizard"
 
-    lot_id = fields.Many2one(comodel_name="stock.lot", string="Lot/Serial Number")
+    lot_id_domain = fields.Binary(compute="_compute_lot_id_domain")
+    lot_id = fields.Many2one(
+        comodel_name="stock.lot", string="Lot/Serial Number", domain="lot_id_domain"
+    )
     lots_visible = fields.Boolean(compute="_compute_lots_visible")
+
+    @api.depends("move_id", "product_id")
+    def _compute_lot_id_domain(self):
+        for rec in self:
+            smls = rec.move_id.move_line_ids.filtered(
+                lambda x: x.state == "done" and x.lot_id
+            )
+            domain = [("id", "in", smls.lot_id.ids)]
+            rec.lot_id_domain = domain
 
     @api.depends("product_id.tracking")
     def _compute_lots_visible(self):


rma_sale_lot/wizards/sale_order_line_rma_wizard.py
```diff
@@ -7,9 +8,21 @@
 class SaleOrderLineRmaWizard(models.TransientModel):
     _inherit = "sale.order.line.rma.wizard"
 
-    lot_id = fields.Many2one(comodel_name="stock.lot", string="Lot/Serial Number")
+    domain_lot_id = fields.Binary(compute="_compute_domain_lot_id")
```
pedrobaeza:
```suggestion
    lot_id_domain = fields.Binary(compute="_compute_lot_id_domain")
```

pedrobaeza (commented):
Not sure if the goal is to restrict it so much. @sbejaoui do you see any problem?

rma_sale_lot/wizards/sale_order_rma_wizard.py
```diff
@@ -7,9 +8,21 @@
...
     lots_visible = fields.Boolean(compute="_compute_lots_visible")
 
+    @api.depends("move_id", "product_id")
+    def _compute_lot_id_domain(self):
+        for rec in self:
+            smls = rec.move_id.move_line_ids.filtered(
+                lambda x: x.state == "done" and x.lot_id
+            )
+            domain = [("id", "in", smls.lot_id.ids)]
+            rec.lot_id_domain = domain
```
carlos-lopez-tecnativa, line 35: I agree with adding a domain, because before this commit, any lot could be selected, even lots that belong to other products, which is an error.
But I also think this domain can be too restrictive; I’m not sure, so I’ll wait for more opinions in this case.

victoralmau: In the context of the RMA creation wizard from a sales order, in my opinion, you should only be able to select the lots that have been used and not others, don't you think?

sbejaoui:
When opening the wizard, the return lines are automatically populated based on the delivery information, so the lot is already preselected

In our case, users rely on this mechanism to manage customer returns and rarely modify the proposed values

But, if a user does change the suggested lot, it is likely for a good reason, and he knows what he is doing, in this context, applying a restrictive domain could be counterproductive

A domain would be more relevant if the lot were not prefilled by the system, but that is not the case here

If there is still a need to control this field for specific users, a better approach would be to make it read-only for a particular user group so they only change the returned qty by lot

victoralmau:
In my opinion, even though it doesn't happen often because it's self-selecting, allowing users to select a lot that hasn't been used is a mistake (right now, lots aren't even filtered by product).

victoralmau:
> When opening the wizard, the return lines are automatically populated based on the delivery information, so the lot is already preselected
> 
> In our case, users rely on this mechanism to manage customer returns and rarely modify the proposed values
> 
> But, if a user does change the suggested lot, it is likely for a good reason, and he knows what he is doing, in this context, applying a restrictive domain could be counterproductive
> 
> A domain would be more relevant if the lot were not prefilled by the system, but that is not the case here
> 
> If there is still a need to control this field for specific users, a better approach would be to make it read-only for a particular user group so they only change the returned qty by lot

Although the field is automatically populated, it is currently possible to modify that data; this change is intended to ensure consistency and prevent users from selecting a lot that was not used in related stock move.

In my opinion, I don't think making the field read-only in some cases will solve the problem—it simply won't allow certain users to do so, while others can. For that reason, this change is intended to resolve any potential data inconsistencies (even if they occur by mistake).

sbejaoui (approved):
After all, I don't see why we shouldn't apply the domain, data consistency should be preserved at all times

pedrobaeza (approved):
OK, let's move with this and let's see if we find a practical example where limiting this way is a problem:

/ocabot merge patch

PR #555, opened 2026-03-18
[18.0][IMP] rma_sale_lot: Add lot data to sale order portal

victoralmau:
Add lot data to sale order portal

@Tecnativa TT61559


pedrobaeza (commented):
The same comment about putting specifically if "Lot" or "Serial Number(s)".

victoralmau:
> The same comment about putting specifically if "Lot" or "Serial Number(s)".

In this case, the text is generic for the table header; although I’m not sure how common this is, it’s possible that one product might use lot numbers while another uses unique serial numbers.

pedrobaeza (approved):
Yeah, I see...

carlos-lopez-tecnativa (changes requested):
Adding new columns causes the table layout to render incorrectly.
Image

I suggest using a higher number, such as colspan="99", to prevent this.

https://github.com/OCA/rma/blob/f64c1d09e111620e9a600d294359388a3d27644e/rma_sale/views/sale_portal_template.xml#L179

carlos-lopez-tecnativa (approved):
LGTM

PR #556, opened 2026-03-18
[18.0][IMP] rma_lot: Add lot to the portal's RMA view

victoralmau:
Changes done:
- [x] Add lot to the portal's RMA view
- [x] Display the correct string (Lot or Serial Number) based on the product's tracking information

@Tecnativa TT61559


rma_lot/views/rma_portal_template.xml
```diff
@@ -0,0 +1,14 @@
+<odoo>
+    <template id="portal_rma_page_lot" inherit_id="rma.portal_rma_page">
+        <xpath expr="//div[@t-if='rma.product_uom_qty']" position="before">
+            <div t-if="rma.lot_id" class="row mb-2 mb-sm-1">
+                <div class="col-12 col-sm-4">
+                    <strong>Lot/Serial Number</strong>
```
pedrobaeza: I think this text can be customized for putting only "Lot" or "Serial Number" depending on the tracking of the product.

rma_lot/views/rma.xml
```diff
@@ -7,10 +7,23 @@
         <field name="inherit_id" ref="rma.rma_view_form" />
         <field name="arch" type="xml">
             <xpath expr="//field[@name='delivered_qty']" position="before">
+                <field name="product_tracking" invisible="1" />
```
carlos-lopez-tecnativa: From v18+, fields used in attributes (readonly, invisible, required) are no longer required to be included in the view.

```suggestion
```

rma_lot/views/rma.xml
```diff
@@ -7,10 +7,23 @@
...
+                    name="lot_id"
+                    groups="stock.group_production_lot"
+                    invisible="not lots_visible or product_tracking != 'serial'"
+                    string="Serial Number"
+                />
+                <field
+                    name="lot_id"
+                    groups="stock.group_production_lot"
+                    invisible="not lots_visible or product_id"
                 />
```
carlos-lopez-tecnativa: I don’t understand why we need three fields. If it’s only to change the string, you can add a `<label>` element and modify the invisible attribute instead of adding the field multiple times, as in https://github.com/odoo/odoo/blob/11d142750637052f9fddd18e049d436078fb2bdf/addons/account/views/account_move_views.xml#L972-L975

carlos-lopez-tecnativa (changes requested):

carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #560, opened 2026-03-23
[FIX] rma_repair: auto delivery creation regardless of repair state

sbejaoui:
When delivery is automatically created after RMA reception, it shouldn't be restricted by the repair state

The dependency between delivery and repair state is relevant for manual flows but shouldn't apply to automatic workflows


lmignon (approved):
LGTM  Code review. Makes sense

jbaudoux (approved):

peluko00 (approved):
LGTM!

nicolas-delbovier-acsone:
(rebased)

rousseldenis (approved):

PR #562, opened 2026-03-24
[17.0][FIX] rma_sale: Wrong method called in stock_picking_return

aaladro-kernet:
Issue: https://github.com/OCA/rma/issues/561#issuecomment-4116455267

PR #563, opened 2026-03-26
[18.0][FIX] rma_sale: Wrong method called in stock_picking_return

victoralmau:
FWP from 17.0: https://github.com/OCA/rma/pull/562

Wrong method called in stock_picking_return

@Tecnativa


rma_sale/wizard/stock_picking_return.py
pedrobaeza: As you are here, you can short this:

```suggestion
            vals["order_id"] = sale_order.id
```

pedrobaeza (approved):
Fast-tracking the simple fw-port

/ocabot merge patch

PR #567, opened 2026-03-31
[18.0][IMP] rma_lot: Add lot_id field to Replace wizard

victoralmau:
Add `lot_id` field to Replace wizard

@Tecnativa TT61741


rma_lot/wizards/rma_delivery.py
```diff
@@ -0,0 +1,42 @@
...
+        if self.product_id and self.product_tracking:
+            rma_ids = self.env.context.get("active_ids")
+            rma = self.env["rma"].browse(rma_ids)
+            if len(rma) == 1:
+                self.lot_id = rma.lot_id
+        return res
+
+    def action_deliver(self):
+        if self.type == "replace" and self.lot_id:
+            self = self.with_context(replace_lot_id=self.lot_id)
```
pedrobaeza: I know you don't want to change signatures, but passing context keys is a weak mechanism, and more with so generic context key. Is the method overridden in a lot of places so considering adding the argument is too much? If yes, then at least use a more specific context key.

victoralmau: IMO, adding an argument to an existing method could potentially be problematic, and besides, I wouldn't know how to do it given how it's currently implemented: https://github.com/OCA/rma/blob/3b9fde3ca52f10be6a3d8aec3a08d4a5cc8270a6/rma/wizard/rma_delivery.py#L82

I’ve changed the context key to something more specific (`rma_replace_lot_id`) to avoid any potential issues with such a generic name.

pedrobaeza (approved):
OK, let's use it this way.

rma_lot/wizards/rma_delivery.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    product_tracking = fields.Selection(
+        selection="_selection_product_tracking", compute="_compute_product_tracking"
+    )
+
+    @api.depends("product_id")
+    def _compute_product_tracking(self):
+        for item in self:
+            item.product_tracking = item.product_id.tracking
+
```
carlos-lopez-tecnativa: Why not use a simple related field instead of a computed one?

carlos-lopez-tecnativa (approved):

PR #571, opened 2026-04-01
[18.0][ADD]: rma_sale_stock_picking_group_by_partner_by_carrier: glue module

StephaneMangin:
# Problem

`stock_picking_group_by_partner_by_carrier` replaces the standard `sale.order.picking_ids` (`One2many` via `sale_id`) with a `Many2many` computed from `move_ids.group_id.sale_ids`.

It also adds a `sale_ids` (`Many2many`) field on `procurement.group`.

However, `rma_sale._prepare_procurement_group_vals()` only sets `sale_id` (the `Many2one`) on the procurement group — it does **not** populate `sale_ids` (the `Many2many`).

As a result, `stock.picking._compute_sale_ids()` → `move_ids.group_id.sale_ids` returns an empty recordset, and the RMA return picking becomes invisible from the SO.

| | Standard | Changes |
|---|---|---|
| `sale.order.picking_ids` | `One2many('stock.picking', 'sale_id')` | `Many2many` (computed) |
| `stock.picking` → SO link | `sale_id` (computed from `group_id.sale_id`) | `sale_ids` (computed from `move_ids.group_id.sale_ids`) |
| `procurement.group` → SO | `sale_id` (`Many2one`) | `sale_id` + **`sale_ids`** (`Many2many`) |

`rma_sale` only writes `sale_id` → the M2M path is broken.


StephaneMangin:
An other problem with the module rma_sale_mrp occurs...

PR #573, opened 2026-04-10
[18.0][IMP] rma_delivery: Filter available carriers

victoralmau:
Filter by available carriers (similar to what happens in sale.order)  

@Tecnativa TT61753

### DIFF
--- a/rma_delivery/models/delivery_carrier.py
+++ b/rma_delivery/models/delivery_carrier.py
@@ -1,0 +1,98 @@
+# Copyright 2026 Tecnativa - Víctor Martínez
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    # TODO 19.0: The available_carriers_* methods should be removed, and the core
+    # methods should be reused and extended for RMA.
+    def available_carriers_picking(self, partner, picking):
+        return self.filtered(lambda c: c._match_picking(partner, picking))
+
+    def available_carriers_rma(self, partner, rma):
+        return self.filtered(lambda c: c._match_rma(partner, rma))
+
+    def _match_picking(self, partner, picking):
+        self.ensure_one()
+        return (
+            self._match_address(partner)
+            and self._match_must_have_tags_picking(picking)
+            and self._match_excluded_tags_picking(picking)
+            and self._match_weight_picking(picking)
+            and self._match_volume_picking(picking)
+        )
+
+    def _match_rma(self, partner, rma):
+        self.ensure_one()
+        return (
+            self._match_address(partner)
+            and self._match_must_have_tags_rma(rma)
+            and self._match_excluded_tags_rma(rma)
+            and self._match_weight_rma(rma)
+            and self._match_volume_rma(rma)
+        )
+
+    def _match_must_have_tags_picking(self, picking):
+        self.ensure_one()
+        return not self.must_have_tag_ids or any(
+            tag in picking.move_ids.product_id.all_product_tag_ids
+            for tag in self.must_have_tag_ids
+        )
+
+    def _match_must_have_tags_rma(self, rma):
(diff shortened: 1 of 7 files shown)


rma_delivery/models/delivery_carrier.py
```diff
@@ -0,0 +1,60 @@
...
+
+from odoo import models
+
+
+class DeliveryCarrier(models.Model):
+    _inherit = "delivery.carrier"
+
+    # TODO 19.0: In all the _match_* methods, you'll need to remove the
+    # if statement for stock.picking because it's already included in delivery
+    def _match_must_have_tags(self, source):
```
pedrobaeza: Better to not rename arguments. I don't think you should override these methods, but call new ones `available_carriers_picking`, `available_carriers_rma`, `_match_picking`, `_match_rma`, etc with similar prefixews, as anyway all the code is basically new. Avoid also use `mapped` when the target is a x2x field, as it's done automatically, and then it's clearer in code.

pedrobaeza (changes requested):

rma_delivery/models/delivery_carrier.py
victoralmau: I think it's important to provide a little more context.

I initially created new methods for these features (such as `_match_must_have_tags_from_picking()`), but upon reviewing the existing `delivery` code in 19.0 (https://github.com/odoo/odoo/blob/2c1c90deb07bbd4a7a372af6d8129db4bafaa2fa/addons/delivery/models/delivery_carrier.py#L208), I decided against it and to use the same methods so that the code is more readable and so that it is very clear what needs to be done during the migration to 19.0.

Renaming the argument is intentional to make it more readable and consistent (especially with 19.0 in mind).

Considering all this, do you think we should create specific methods (available_carriers_picking, available_carriers_rma, and all related ones)? I think it’s better to leave it as is, especially with 19.0 in mind, but that’s just my opinion.

pedrobaeza: Thanks for the context. Indeed, in that version, we will be able to reuse methods, but not yet in this version. Renaming argument should be avoided at all, as you may have one call in any other module using the positional argument as keyword argument (example: `carrier._match_must_have_tags(order=order)`), and this will crash with your renaming.

About reusing the same methods, as in this version it's not possible yet, I would do it separate methods, for having clear code in this version, but putting a comment saying about 19 and how to do it.

rma_delivery/models/delivery_carrier.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # TODO 19.0: The available_carriers_* methods should be removed, and the core
+    # methods should be reused and extended for RMA.
+    def available_carriers_picking(self, partner, picking):
+        return self.filtered(lambda c: c._match_picking(partner, picking))
+
+    def available_carriers_rma(self, partner, rma):
+        return self.filtered(lambda c: c._match_rma(partner, rma))
+
+    def _match_picking(self, partner, picking):
```
carlos-lopez-tecnativa, line 18: I think you need to rename this method to avoid conflicts with other modules, for example:
https://github.com/OCA/delivery-carrier/blob/94609c47c8346827d2297282d1461cb40138e56c/delivery_carrier_picking_valid/models/delivery_carrier.py#L10

carlos-lopez-tecnativa (changes requested):

rma_delivery/models/delivery_carrier.py
victoralmau: Interesting, I wasn't aware of that, BUT I think the `delivery_carrier_picking_valid` module isn't working because the `_match_picking()` method is never called; this won't be a problem in v19 because it's already “backward-compatible.”

carlos-lopez-tecnativa: Ok, let’s go with it.

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge minor

PR #574, opened 2026-04-13
[18.0][IMP] rma: Add Create RMA button to rma page

victoralmau:
Add Create RMA button to rma page

@Tecnativa TT61741


pedrobaeza:
Can you expand a bit the purpose of the PR?

victoralmau:
> Can you expand a bit the purpose of the PR?

Use case example:
- A sales order is created and the delivery picking is done
- An RMA 1 is created to manage item related to the sales order
- An RMA 2 is created to manage item related to RMA 1

This wizard is designed to simplify the process of creating RMA 2 directly from RMA 1.

pedrobaeza:
I'm afraid I still don't understand RMA from RMA case, but I don't want to block this.

victoralmau:
> I'm afraid I still don't understand RMA from RMA case, but I don't want to block this.

If a customer (me) purchases something, create a return (RMA 1), and wants to return the item they received, an RMA 2 must be created to process it.

pedrobaeza:
Do you mean like a second attempt after the first one was finished?

victoralmau:
> Do you mean like a second attempt after the first one was finished?

That's right; for example, the product being returned to the customer (in RMA 1) is broken.

pedrobaeza:
OK, but at the end, the RMA is linked to the original one. This is only an accelerator for not having to go to the original sales order, isn't it?

victoralmau:
> OK, but at the end, the RMA is linked to the original one. This is only an accelerator for not having to go to the original sales order, isn't it?

No, this is a new RMA. Since this wizard isn't available, and RMA 1 isn't linked to RMA 2, you need to go to the delivery picking (RMA 1) and use the “Return” button there to create RMA 2.

rma/views/res_config_settings_views.xml
```diff
@@ -156,6 +156,17 @@
                                 </div>
                             </div>
                         </div>
+                        <div
+                            class="col-12 col-lg-6 o_setting_box"
+                            title="Send automatic notification when the customer places an RMA"
```
carlos-lopez-tecnativa: Please update the title; this is not related to the purpose of the setting.
You can add some text in the setting to help users understand what happens when the setting is activated. The text is not clear.

Image

rma/models/res_company.py
```diff
@@ -65,6 +65,10 @@ def _default_rma_mail_draft_template(self):
         default=_default_rma_mail_draft_template,
         help="Email sent to the customer when they place " "an RMA from the portal",
     )
+    rma_new_rma_button_from_rma = fields.Boolean(
+        string="New RMA button",
+        help="New RMA button from rma page",
```
carlos-lopez-tecnativa: I’m not sure if this string and help text are really helpful.

rma_delivery/wizard/rma_rma_wizard_views.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="rma_rma_wizard_form_view" model="ir.ui.view">
+        <field name="model">rma.rma.wizard</field>
+        <field name="inherit_id" ref="rma.rma_rma_wizard_form_view" />
+        <field name="arch" type="xml">
+            <field name="operation_id" position="after">
+                <field name="reception_carrier_id" />
```
carlos-lopez-tecnativa: Why do you always display this field? In other places, it is only displayed based on `rma_delivery_strategy` or `rma_reception_strategy`. In this case, why isn’t the same logic applied?

rma/models/rma.py
```diff
@@ -681,6 +701,33 @@ def _send_receipt_confirmation_email(self):
...
     # Action methods
+    def action_create_rma(self):
+        self.ensure_one()
+        action = self.env["ir.actions.act_window"]._for_xml_id(
+            "rma.rma_create_rma_action"
+        )
+        # Force active_id to avoid issues when coming from smart buttons
+        # in other models
+        action["context"] = dict(self.env.context)
+        action["context"].update(active_id=self.id, active_ids=self.ids)
```
carlos-lopez-tecnativa:
```suggestion
        action["context"].update(active_model=self._name, active_id=self.id, active_ids=self.ids)
```

rma/views/res_config_settings_views.xml
```diff
@@ -156,6 +156,17 @@
...
+                        >
+                            <div class="o_setting_left_pane">
+                                <field name="rma_new_rma_button_from_rma" />
+                            </div>
+                            <div class="o_setting_right_pane">
+                                <label for="rma_new_rma_button_from_rma" />
+                            </div>
```
carlos-lopez-tecnativa: I think you need to update the view architecture; this changed in v17+.
Image

rma/wizard/rma_rma_wizard.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @api.model
+    def default_get(self, fields_list):
+        res = super().default_get(fields_list)
+        rma_id = self.env.context.get("active_id")
+        rma = self.env["rma"].browse(rma_id)
+        if rma:
+            res.update(
+                operation_id=rma.operation_id.id,
+            )
```
carlos-lopez-tecnativa:
```suggestion
            res.update(
                operation_id=rma.operation_id.id
            )
```

rma_delivery/wizard/rma_rma_wizard.py
```diff
@@ -0,0 +1,30 @@
...
+
+    @api.model
+    def default_get(self, fields_list):
+        res = super().default_get(fields_list)
+        rma_id = self.env.context.get("active_id")
+        rma = self.env["rma"].browse(rma_id)
+        if rma:
+            res.update(
+                reception_carrier_id=rma.reception_carrier_id.id,
+            )
```
carlos-lopez-tecnativa:
```suggestion
            res.update(
                reception_carrier_id=rma.reception_carrier_id.id
            )
```

rma/wizard/rma_rma_wizard.py
```diff
@@ -0,0 +1,74 @@
...
+        # Call the onchange event manually to avoid using Form()
+        return_wizard._onchange_create_rma()
+        return_wizard.product_return_moves.filtered(
+            lambda x: x.move_id not in delivery_moves
+        ).unlink()
+        res = return_wizard.action_create_returns_all()
+        new_picking = self.env[res["res_model"]].browse(res["res_id"])
+        new_rma = new_picking.move_ids.rma_receiver_ids
+        # Message
+        rma_msg = _(
```
carlos-lopez-tecnativa:
```suggestion
        rma_msg = self.env._(
```

carlos-lopez-tecnativa (changes requested):

rma/views/res_config_settings_views.xml
victoralmau: Okay, I'm going to create a pull request to fix this in v17 first.

victoralmau:

victoralmau:
Ping @carlos-lopez-tecnativa

rma_reason/wizards/rma_rma_wizard.py
```diff
@@ -0,0 +1,40 @@
...
+                reason_id=rma.reason_id.id,
+            )
+        return res
+
+    @api.depends("reason_id")
+    def _compute_domain_operation_id(self):
+        for item in self:
+            domain = []
+            if item.reason_id:
+                domain = [("id", "in", item.reason_id.allowed_operation_ids.ids)]
```
carlos-lopez-tecnativa: I think this domain should be applied only if the reason has `allowed_operation_ids`; otherwise, when I try to change the operation, no records are available to select.

Image


```suggestion
            if item.reason_id.allowed_operation_ids:
                domain = [("id", "in", item.reason_id.allowed_operation_ids.ids)]
```

carlos-lopez-tecnativa (changes requested):

rma_reason/wizards/rma_rma_wizard_views.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="rma_rma_wizard_form_view" model="ir.ui.view">
+        <field name="model">rma.rma.wizard</field>
+        <field name="inherit_id" ref="rma.rma_rma_wizard_form_view" />
+        <field name="arch" type="xml">
+            <field name="operation_id" position="after">
+                <field name="reason_id" />
+            </field>
```
carlos-lopez-tecnativa: If you apply a domain on operation_id based on reason_id, I think the reason should be displayed first, then the operation, similar to the RMA form view.
Image

Additionally, if you apply a filter and operation_id is already set, the selected reason may not be valid for that operation, making the data inconsistent. So, when you select a reason, if the current operation is not allowed, the field remains set instead of being cleared. I think you should handle this in an onchange on the reason.

rma/models/rma.py
```diff
@@ -681,6 +701,35 @@ def _send_receipt_confirmation_email(self):
...
+            action.update(
+                res_id=rma.id,
+                view_mode="form",
+                views=[],
+            )
+        else:
+            action["domain"] = [("id", "in", rma.ids)]
+        # reset context to show all related rma without default filters
+        action["context"] = {}
+        return action
```
carlos-lopez-tecnativa: You can use this function to handle 1 or more records
https://github.com/odoo/odoo/blob/d73e5662a0af7c549008661f743ba5d51f765339/odoo/addons/base/models/ir_ui_view.py#L2899


```suggestion
        rma = self.delivery_move_ids.mapped("rma_ids")
        action = rma._get_records_action()
        # reset context to show all related rma without default filters
        action["context"] = {}
        return action
```

rma/wizard/rma_rma_wizard.py
```diff
@@ -0,0 +1,72 @@
...
+        rma = self.create_rma()
+        rma.action_confirm()
+        action = self.env["ir.actions.act_window"]._for_xml_id("rma.rma_action")
+        action.update(
+            res_id=rma.id,
+            view_mode="form",
+            view_id=False,
+            views=False,
+        )
+        return action
```
carlos-lopez-tecnativa:
```suggestion
        return rma._get_records_action()
```

carlos-lopez-tecnativa (changes requested):

rma_reason/wizards/rma_rma_wizard_views.xml
victoralmau: OK, I've changed in the wizard included in this PR

carlos-lopez-tecnativa (approved):
LGTM

victoralmau:
Ping @pedrobaeza

pedrobaeza (commented):
Isn't possible to use the same RMA wizard that already exists for not duplicating so many code? AFAIK, the source is still a picking (the RMA delivery picking against the OUT picking), so most of the code should work. You only need to adjust the selection of such picking.

victoralmau:
So, if I understand correctly, you want the return wizard for a picking order to appear when the RMA button is clicked? I think that while some elements could be reused, it might be a bit confusing—for example, the “Create RMA” checkbox shouldn't be displayed, nor should the wizard lines, to avoid confusion. 

In any case, it might have been better to propose that approach from the start, rather than doing it now that all the code and extra modules have already been changed.

pedrobaeza:
The aim is to reduce the maintenance burden and the possibility to have desynchronized features in one wizard or the other. If you don't think is something valuable, we can continue this way.

victoralmau:
> The aim is to reduce the maintenance burden and the possibility to have desynchronized features in one wizard or the other. If you don't think is something valuable, we can continue this way.

Thinking again about the approach you mentioned, I don't think it's possible to do that because an RMA user without inventory permissions should be able to perform that process, and a user without inventory permissions doesn't have access to the `stock.return.picking` wizard.

pedrobaeza (approved):
OK, let's continue this way.

/ocabot merge minor

PR #582, opened 2026-05-04
[18.0][FIX] rma_sale_delivery: Use proper available carriers method

pedrobaeza:
We can't use the RMA one, as it doesn't exist yet. Not sure though if the sales order check is enough, or there should be something more, as this is a return.

@Tecnativa TT62250

### DIFF
--- a/rma_sale_delivery/wizard/sale_order_rma_wizard.py
+++ b/rma_sale_delivery/wizard/sale_order_rma_wizard.py
@@ -32,7 +32,7 @@ class SaleOrderRmaWizard(models.TransientModel):
                 carrier_model._check_company_domain(item.company_id)
             )
             item.available_reception_carrier_ids = (
-                carriers.available_carriers_rma(partner, item.order_id)
+                carriers.available_carriers(partner, item.order_id)
                 if partner
                 else carriers
             )


rma_sale_delivery/wizard/sale_order_rma_wizard.py
```diff
@@ -32,7 +32,7 @@ def _compute_available_reception_carrier_ids(self):
                 carrier_model._check_company_domain(item.company_id)
             )
             item.available_reception_carrier_ids = (
-                carriers.available_carriers_rma(partner, item.order_id)
```
carlos-lopez-tecnativa, line 35: Sorry, but I don’t understand what the error is.

This method is available at https://github.com/OCA/rma/blob/4b84f797be71519359203b412e070ebd0b6f9ac8/rma_delivery/models/delivery_carrier.py#L15

pedrobaeza: Because it expects a RMA, not a sales order...

carlos-lopez-tecnativa: Ahhh, Ok, good catch

carlos-lopez-tecnativa (approved):

victoralmau (approved):

PR #583, opened 2026-05-12
[18.0][FIX] rma + rma_sale: Full RMA workflow without additional permissions (Inventory or Sales)

victoralmau:
Full RMA workflow without additional permissions (Inventory or Sales)

@Tecnativa

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -435,7 +435,7 @@ class Rma(models.Model):
                 item.state in ("returned", "replaced")
                 and item.company_id.rma_new_rma_button_from_rma
                 and item.delivery_move_ids
-                and all(m.state == "done" for m in item.delivery_move_ids)
+                and all(m.state in ("done", "cancel") for m in item.delivery_move_ids)
                 and not any(m.rma_ids for m in item.delivery_move_ids)
             )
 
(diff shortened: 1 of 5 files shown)


pedrobaeza:
To which point is the RMA module functional without stock permissions? Should we go this way to restrict visibility, or should we grant equivalent stock permissions to RMA users?

victoralmau:
Although it’s not common (or at least that’s what I think) to use the RMA module without inventory permissions, it could happen; I don’t think inheriting permissions is appropriate, nor is adding ACLs for everything related to inventory, which is why these changes have been done.

There are some changes related to sales permissions that I could probably add in a separate commit or a separate PR if you prefer.

rma/models/rma.py
```diff
@@ -431,7 +435,7 @@ def _compute_can_be_new_rma(self):
                 item.state in ("returned", "replaced")
                 and item.company_id.rma_new_rma_button_from_rma
                 and item.delivery_move_ids
-                and all(m.state == "done" for m in item.delivery_move_ids)
+                and all(m.state in ("done", "cancel") for m in item.delivery_move_ids)
```
pedrobaeza, line 438: This patch should go in a separate commit explaining why this change.

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #586, opened 2026-05-19
[18.0][IMP] rma_lot: Add domain_lot_id field to lot_id field

victoralmau:
Add domain_lot_id field to lot_id field

Similar to `sale_order_lot_selection`

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT62586

### DIFF
--- a/rma_lot/models/rma.py
+++ b/rma_lot/models/rma.py
@@ -2,21 +2,59 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 from odoo import Command, api, fields, models
+from odoo.tools import float_compare
 
 
 class Rma(models.Model):
     _inherit = "rma"
 
+    domain_lot_id = fields.Binary(compute="_compute_domain_lot_id")
     lot_id = fields.Many2one(
         comodel_name="stock.lot",
         string="Lot/Serial Number",
-        domain="[('product_id', '=?', product_id)]",
+        domain="domain_lot_id",
         compute="_compute_lot_id",
         store=True,
         readonly=False,
     )
     product_tracking = fields.Selection(related="product_id.tracking")
     lots_visible = fields.Boolean(compute="_compute_lots_visible")
+
+    def _domain_lot_id_quant_domain(self):
+        """This method defines the domain that will be used to obtain the appropriate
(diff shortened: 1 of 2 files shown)


rma_lot/models/rma.py
```diff
@@ -2,22 +2,60 @@
...
+                # Only available lots should be displayed. In pickings, the
+                # corresponding stock.quant record is selected directly, so we
+                # use the same filters that are used.
+                quants = self.env["stock.quant"].search(
+                    rec._domain_lot_id_quant_domain()
+                )
+                available_quants = quants.filtered(
+                    lambda x, qty=rec.product_uom_qty: float_compare(
+                        x.available_quantity,
+                        qty,
```
pedrobaeza, line 51: In practical terms, this should be 1 for working.

pedrobaeza (approved):

rma_lot/models/rma.py
victoralmau: It will be 1 if tracking is serial; otherwise, the value may be higher. 
In any case, using the appropriate field covers all scenarios (although we generally use tracking=serial).

pedrobaeza: Yeah, I see that in this case, you haven't limited to `serial`, but to both. I see a potential problem that you make a RMA for >1 qty of a product with serial number. The module doesn't allow to select more than 1 lot.

victoralmau: Similar to what happens in `sale_order_lot_selection`, the domain applied does not distinguish based on the product's tracking field.

In a separate PR (similar to https://github.com/OCA/sale-workflow/pull/4336), after this one is merged, a restriction will be added to prevent “inconsistent actions.”

pilarvargas-tecnativa (approved):

victoralmau:
This is ready to merge, right?

pedrobaeza:
Not really, as the rules state that it should be 3 approvals or 2 and 5 working days, and this one not being our module, it's not fair to not let others to give their opinion being a new behavior. If we have the @sbejaoui's blessing, we can continue.

victoralmau:
Ping @carlos-lopez-tecnativa

sbejaoui (approved):
LGTM, thank you

PR #587, opened 2026-05-19
[18.0][IMP] rma_lot: Add quantity serial number validation when replace/return process

victoralmau:
Add quantity serial number validation when replace/return process

Similar to https://github.com/OCA/sale-workflow/pull/4336

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT62586

### DIFF
--- a/rma_lot/i18n/es.po
+++ b/rma_lot/i18n/es.po
@@ -89,6 +89,12 @@ msgid "Serial Number"
 msgstr "Número de serie"
 
 #. module: rma_lot
+#. odoo-python
+#: code:addons/rma_lot/models/rma.py:0
+msgid "The serial number %(serial)s is not available"
+msgstr "El número de serie %(serial)s no está disponible"
+
+#. module: rma_lot
 #: model:ir.model.fields,field_description:rma_lot.field_rma__product_tracking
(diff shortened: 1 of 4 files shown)


victoralmau:
Ping @pedrobaeza and @pilarvargas-tecnativa

pedrobaeza (commented):
I think returns should check in RMA/WH, and replaces, if it's other S/N, does it make sense to do the check?

victoralmau:
I think it's important (once again) to ensure data consistency, and that the replacement (out picking) wizard shouldn't allow users to select a kit that isn't available.

pedrobaeza:
Yeah, I'm not saying the contrary, but to look in both locations for both operations seems not right.

victoralmau:
> Yeah, I'm not saying the contrary, but to look in both locations for both operations seems not right.

It's an interesting question, but so far I haven't found a better way to do it; regardless of the RMA location specified, depending on the routes and the settings for each one, stock will be drawn from one location or the other—or both—which is why the system checks both locations.

pedrobaeza:
As the method is called from different places, you can distinguish the need from that calls, and only look on RMA/WH when returning, and the other when replacing.

PR #588, opened 2026-05-25
[18.0][FIX] rma: Fix can_be_replaced value

victoralmau:
Fix can_be_replaced value

@Tecnativa

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -477,12 +477,7 @@ class Rma(models.Model):
             r.can_be_replaced = (
                 r.operation_id.action_create_delivery
                 in ("manual_after_receipt", "automatic_after_receipt")
-                and r.state
-                in [
-                    "received",
-                    "waiting_replacement",
-                    "replaced",
-                ]
+                and r.state == "received"
             ) or (
                 r.operation_id.action_create_delivery
                 in ("manual_on_confirm", "automatic_on_confirm")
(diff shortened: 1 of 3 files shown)


carlos-lopez-tecnativa (commented):
I think you need to add a bit more context to explain the change or provide a real use case. Does this change come from a customer-specific need, or is it part of a general requirement/flow?

victoralmau:
This change is not based on a customer requirement; it is simply intended to ensure consistency:
- If the RMA is in the “Awaiting Replacement” status (the replacement has already been generated), the “Replace” button should not appear, and it should not be possible to create a new one.
- If the RMA is in the “Replaced” status (the replacement has already been generated and validated), the ‘Replace’ button should not appear, and it should not be possible to create a new one; the process is already “completed.”

pedrobaeza (commented):
Do you know any reason why it was that way before?

victoralmau:
No, I don't know why it's like that right now; maybe someone can explain it to us.

carlos-lopez-tecnativa (approved):
LGTM

pilarvargas-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

pedrobaeza:
@sbejaoui this change makes sense, but don't know if there's any of your flows that require the previous condition.

sbejaoui (commented):
LGTM thanks

PR #591, opened 2026-06-01
[18.0][IMP] rma + rma_sale: Add new rma compatibility

victoralmau:
Add new rma compatibility

Use case example:
- Create a sales order
- Confirm the sales order + done picking
- Create RMA 1 from the sales order
- Done reception picking of RMA 1 + return to the customer + done delivery picking for RMA 1
- Create RMA 2 from RMA 1
- Done reception picking of RMA 2
- Refund RMA 2
- Credit note must be linked to the sales order

@Tecnativa TT62775

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -137,6 +137,9 @@ class Rma(models.Model):
         """Inject fiscal_position_id + salesman from sales order (if any)"""
         vals = super()._prepare_refund_vals(origin=origin)
         order = self.sudo().order_id
+        if not order and self.move_id.rma_id:
+            # We use the RMA from which it was created
+            order = self.move_id.rma_id.sudo().order_id
         if order:
             vals["invoice_user_id"] = order.user_id.id
             # It is important to set the correct fiscal position for the sales order
@@ -156,6 +159,9 @@ class Rma(models.Model):
         """
         vals = super()._prepare_refund_line_vals()
         line = self.sudo().sale_line_id
+        if not line and self.move_id.rma_id:
+            # We use the RMA from which it was created
+            line = self.move_id.rma_id.sudo().sale_line_id
         if line:
             vals["product_id"] = line.product_id.id
             vals["price_unit"] = line.price_unit
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -710,6 +710,7 @@ class Rma(models.Model):
     # Action methods
     def action_create_rma(self):
         self.ensure_one()
+        self._ensure_can_be_new_rma()
         action = self.env["ir.actions.act_window"]._for_xml_id(
             "rma.rma_create_rma_action"
         )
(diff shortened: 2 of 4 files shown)


pedrobaeza (commented):
Isn't this one fix and one imp that are not related (so 2 commits)?

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

victoralmau:
I think this is ready to be merged, right?

pedrobaeza:
No 5 days have passed, and this kind of things on popular modules, I prefer to let other people to give their opinion.

pilarvargas-tecnativa (approved):

rma_sale/models/rma.py
```diff
@@ -156,6 +159,9 @@ def _prepare_refund_line_vals(self):
         """
         vals = super()._prepare_refund_line_vals()
         line = self.sudo().sale_line_id
+        if not line and self.move_id.rma_id:
+            # We use the RMA from which it was created
+            line = self.move_id.rma_id.sudo().sale_line_id
```
sbejaoui, line 164: Hi @victoralmau,

IMO, we should handle this use case differently

to summarize, here we retrieve the sale line from the original rma when a new rma is created from the delivery generated by an existing rma

the problem is that this link only works for one level, if another rma is created from the second delivery, `move_id.rma_id.sale_line_id` will be empty, making this lookup recursive would also add unnecessary complexity

I think a better approach would be to propagate the sale information when creating the new rma: if it originates from `move_id.rma_id`, inherit its sale order and sale line, this way, the original sale information is preserved across all subsequent rmas

another advantage is that we keep a direct link between the original sale order and all related rmas, making it easier for the manager to see when a product has been returned multiple times

we should check whether the returned quantity is computed elsewhere based on these links and adapt it if necessary

cc/ @pedrobaeza, @rousseldenis, @marielejeune

pedrobaeza: I don't think there's a problem to keep both links: the RMA one, and the sales one. @victoralmau do you see any problem with that?

victoralmau: It is not possible to do so for various reasons, but especially because it would also be conceptually incorrect

Example:
Product A with quantity 1 and delivery to WH/OUT/00001

We create RMA 1 (it doesn't matter whether it's from the sales order or from the “Return” button on the WH/OUT/00001 picking), and WH/RMA/IN/00001 is created.

RMA 1 is automatically linked to the sales order line https://github.com/OCA/rma/blob/54ca6cdd50568422621f3d0f63d9c2ca465124d9/rma_sale/models/rma.py#L37

We confirm the receipt of WH/RMA/IN/00001 and process the delivery (of the same product, repaired for example) using WH/RMA/OUT/00001.

Later, the customer indicates that the item it's broken. in WH/RMA/OUT/00001 is broken and wants to create an RMA.

Linking RMA 2 to the sales order would be incorrect or impossible for several reasons:
- The `sale_line_id` field is a related `move_id.sale_line_id`
- Conceptually and in terms of actual inventory, what needs to be done is to “return” WH/RMA/OUT/00001 (this is an automatic process handled by https://github.com/OCA/rma/pull/574) to create WH/RMA/IN/00002. It is not possible (nor correct) to return the item again from WH/OUT/00001.
- It would be conceptually incorrect to link RMA 1 and RMA 2 in the sales order’s smart button, because it would “cause confusion” if, for example, the order had multiple lines (or a quantity greater than 1) and multiple RMAs were created at different times—and because RMA 2 is not (and should not be) linked to the sales order; it is linked to RMA 1. Conceptually, the same should apply in the portal (RMA 2 should not be linked to the sales order).

Having clarified the above points:
A- It is true that an RMA 3 (and subsequent ones) is not currently provided for, although it is not common (if a customer returns an item twice, it is “unlikely” that they will create an RMA), it should still be taken into account; therefore, I will create a PR to fix this.
B- Display a button or links on the RMA 1 page of the portal to easily access RMA 2.

PR #597, opened 2026-06-11
[18.0][IMP] rma_sale: Refund compatibility with not invoiced orders

victoralmau:
Changes done:
- [x] Refund compatibility with not invoiced orders
- [x] Change refunded label state to Refunded/Not invoiceable

Use case example:
- Define product with Invoicing Policy: Delivered quantities
- Create sale order + Validate picking
- Create RMA + Validate incoming picking
- Refund action on the RMA
- Refund invoice will NOT be created

@Tecnativa TT62394

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -127,10 +127,23 @@ class Rma(models.Model):
 
         NOTE: The refund line is linked to the SO line in `_prepare_refund_line`.
         """
+        _self = self
+        # Do not generate a refund invoice if the sales line has not yet been invoiced
+        # or if it is an RMA created from another RMA with the same condition
+        not_refundable = self.filtered(
+            lambda x: (x.sale_line_id and not x.sale_line_id.invoice_lines)
+            or (
+                x.move_id.rma_id.sale_line_id
+                and not x.move_id.rma_id.sale_line_id.invoice_lines
+            )
+        )
+        self -= not_refundable
         res = super().action_refund()
-        for rma in self:
+        for rma in _self:
             if rma.sale_line_id:
                 rma._link_refund_with_reception_move()
+        # These RMAs should be set to this state anyway
+        not_refundable.state = "refunded"
         return res
 
     def _prepare_refund_vals(self, origin=False):
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -178,7 +178,7 @@ class Rma(models.Model):
             ("received", "Received"),
             ("waiting_return", "Waiting for return"),
             ("waiting_replacement", "Waiting for replacement"),
-            ("refunded", "Refunded"),
+            ("refunded", "Refunded/Not invoiceable"),
             ("returned", "Returned"),
             ("replaced", "Replaced"),
             ("finished", "Finished"),
--- a/rma_sale/tests/test_rma_sale.py
+++ b/rma_sale/tests/test_rma_sale.py
@@ -147,8 +147,13 @@ class TestRmaSale(TestRmaSaleBase):
         self.operation.action_create_delivery = "manual_after_receipt"
         self.operation.action_create_refund = "manual_after_receipt"
         self.company.rma_new_rma_button_from_rma = True
+        self.product_1.invoice_policy = "delivery"
         order = self.sale_order
         order.user_id = self.user_rma
+        order._create_invoices()
+        self.assertEqual(order.invoice_status, "invoiced")
+        self.assertEqual(self.order_line.qty_delivered, 5)
+        self.assertEqual(self.order_line.qty_invoiced, 5)
         wizard = self._rma_sale_wizard(order)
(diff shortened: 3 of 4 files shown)


rma_sale/models/rma.py
```diff
@@ -127,10 +127,19 @@ def action_refund(self):
 
         NOTE: The refund line is linked to the SO line in `_prepare_refund_line`.
         """
+        _self = self
+        # Do not generate a refund invoice if the sales line has not yet been invoiced
+        self = self.filtered(
```
pedrobaeza: Instead of doing this technique, it's better for readability to just do the `if` inside the loop instead of checking the inclusion in this recordset.

victoralmau: I don't quite understand what you're suggesting; I want to ensure that the `action_refund()` method is NOT called for the relevant records.

pedrobaeza: OK, got it. I missed the call to super.

pedrobaeza (approved):

carlos-lopez-tecnativa (commented):
I tested the flow, but I found an inconsistency. After processing the RMA refund, I cannot generate the invoice, so I was unable to continue the flow and generate the refund (account.move).

Please review this sales order. The delivered quantity is zero (1 delivered - 1 returned), so there is nothing to invoice.

I'm not sure whether this is the expected behavior. Since there is nothing to invoice, there is also nothing to refund. Is that correct?

http://oca-rma-18-0-pr597-74e093a1c3ed.runboat.odoo-community.org/odoo/sales/25

Image

Image

pedrobaeza:
Yes, it's correct. We may want to rephrase the state "Refunded" to something more suitable for both cases, or create another state "Not invoiceable", but the behavior is the expected one.

rma_sale/models/rma.py
```diff
@@ -127,10 +127,19 @@ def action_refund(self):
...
+            lambda x: not x.sale_line_id
+            or (x.sale_line_id and x.sale_line_id.invoice_lines)
```
pedrobaeza:
```suggestion
            or x.sale_line_id.invoice_lines
```

rma_sale/models/rma.py
```diff
@@ -127,10 +127,19 @@ def action_refund(self):
 
         NOTE: The refund line is linked to the SO line in `_prepare_refund_line`.
         """
+        _self = self
+        # Do not generate a refund invoice if the sales line has not yet been invoiced
+        self = self.filtered(
+            lambda x: not x.sale_line_id
```
pedrobaeza: I'm thinking you may want to refund free RMAs without sales lines, so remove this condition:


```suggestion
            lambda x:
```

victoralmau:
> Yes, it's correct. We may want to rephrase the state "Refunded" to something more suitable for both cases, or create another state "Not invoiceable", but the behavior is the expected one.

Perhaps the "finished" status is the right one in this case?

pedrobaeza:
Uhm, that one is too generic IMO.

pedrobaeza:
We can label `refunded` as "Refunded/Not invoiceable".

rma_sale/models/rma.py
victoralmau: This test https://github.com/OCA/rma/actions/runs/27356307617/job/80831403997?pr=597#step:8:386 confirms that this condition is necessary (although it is a strange flow)

pedrobaeza: But we should be able to refund standalone RMAs. Imagine that you just moved to Odoo, and you don't have in Odoo neither the SO nor the invoice, and you still want to refund that RMA...

victoralmau: That condition is specifically designed to allow for that.

**Note**: We will select the rma that we will call to super to generate the refund.

pedrobaeza: Oh, sorry, I was interpreting the contrary. Indeed, this is not `_self`, but `self`. And my usual way of thinking is:

- To select the group you want to handle specially.
- Sustract from `self` that group.
- Call `super`.
- Handle the special group.

On that mood, what do you think about recoding it this way (much simple IMO)?:

```python
        # Do not generate a refund invoice if the sales line has not yet been invoiced
        # or if it is an RMA created from another RMA with the same condition
        not_refundable = self.filtered(
            lambda x: not x.sale_line_id.invoice_lines
            and not x.move_id.rma_id.sale_line_id.invoice_lines
        )
        self -= not_refundable
        res = super().action_refund()
        for rma in self:
            rma._link_refund_with_reception_move()
        # These RMAs should be set to this state anyway
        not_refundable.state = "refunded"

victoralmau: This condition is still incorrect (related to https://github.com/OCA/rma/pull/597#discussion_r3397094283)

pedrobaeza (approved):

carlos-lopez-tecnativa (approved):
LGTM

PR #603, opened 2026-06-18
[18.0][IMP] rma_lot: Add methods to improve inheritance and extension

victoralmau:
Add methods to improve inheritance and extension

@Tecnativa

### DIFF
--- a/rma_lot/models/rma.py
+++ b/rma_lot/models/rma.py
@@ -20,17 +20,27 @@ class Rma(models.Model):
     product_tracking = fields.Selection(related="product_id.tracking")
     lots_visible = fields.Boolean(compute="_compute_lots_visible")
 
+    def _get_lot_id_quant_domain_locations(self):
+        """This method retrieves the location(s) that will later be used in
+        _domain_lot_id_quant_domain(). It will be useful to extend this method if,
+        for example, you want to search across different multi-warehouse
+        locations.
+        """
+        self.ensure_one()
+        warehouse = self.warehouse_id
+        return warehouse.lot_stock_id + warehouse.rma_loc_id
+
     def _domain_lot_id_quant_domain(self):
         """This method defines the domain that will be used to obtain the appropriate
         stock.quant values and is useful for extending to other modules.
         """
         self.ensure_one()
+        locations = self._get_lot_id_quant_domain_locations()
         return [
             ("product_id", "=", self.product_id.id),
             ("quantity", ">=", self.product_uom_qty),
             ("lot_id", "!=", False),
-            ("location_id.usage", "=", "internal"),
-            ("location_id.warehouse_id", "=", self.warehouse_id.id),
+            ("location_id", "child_of", locations.ids),
         ]
 
     @api.depends("product_id", "product_uom_qty", "warehouse_id")
--- a/rma_lot/wizards/rma_delivery.py
+++ b/rma_lot/wizards/rma_delivery.py
@@ -16,17 +16,27 @@ class RmaReDeliveryWizard(models.TransientModel):
     )
     product_tracking = fields.Selection(related="product_id.tracking")
 
+    def _get_lot_id_quant_domain_locations(self):
+        """This method retrieves the location(s) that will later be used in
+        _domain_lot_id_quant_domain(). It will be useful to extend this method if,
+        for example, you want to search across different multi-warehouse
+        locations.
+        """
+        self.ensure_one()
+        warehouse = self.warehouse_id
+        return warehouse.lot_stock_id + warehouse.rma_loc_id
+
     def _domain_lot_id_quant_domain(self):
         """This method defines the domain that will be used to obtain the appropriate
         stock.quant values and is useful for extending to other modules.
         """
         self.ensure_one()
+        locations = self._get_lot_id_quant_domain_locations()
         return [
             ("product_id", "=", self.product_id.id),
             ("quantity", ">=", self.product_uom_qty),
             ("lot_id", "!=", False),
-            ("location_id.usage", "=", "internal"),
-            ("location_id.warehouse_id", "=", self.warehouse_id.id),
+            ("location_id", "child_of", locations.ids),
         ]
 
     @api.depends("product_id", "product_tracking", "product_uom_qty", "warehouse_id")


rma_lot/models/rma.py
```diff
@@ -20,17 +20,29 @@ class Rma(models.Model):
...
+        """This method retrieves the location(s) that will later be used in
+        _domain_lot_id_quant_domain(). It will be useful to extend this method if,
+        for example, you want to search across different multi-warehouse
+        locations.
+        """
+        self.ensure_one()
+        warehouse = self.warehouse_id
+        locations = warehouse.lot_stock_id
+        locations += warehouse.rma_loc_id
+        return locations
```
pedrobaeza:
```suggestion
        return warehouse.lot_stock_id + warehouse.rma_loc_id
```

pedrobaeza (approved):

rma_lot/wizards/rma_delivery.py
```diff
@@ -16,17 +16,29 @@ class RmaReDeliveryWizard(models.TransientModel):
     )
     product_tracking = fields.Selection(related="product_id.tracking")
 
+    def _get_lot_id_quant_domain_locations(self):
```
pedrobaeza, line 19: Why not using the other method?

victoralmau: This separate method is very important if, for example, you want to add an additional location or if you have a multi-warehouse environment.

victoralmau: Sorry, I got that wrong. It's important to use a specific method here because you can specify a different warehouse in the wizard.: https://github.com/OCA/rma/blob/ba6e660d9e10808dfc49281a2716a68994af57bb/rma/wizard/rma_delivery.py#L31

pedrobaeza: OK, but then call from inside this one the other one in this base.

victoralmau: Maybe I'm not understanding what you mean, because in my opinion it's not possible:

https://github.com/OCA/rma/pull/603/changes#diff-7b933f9bf4286d07a401f46946531786b8cfdc49bbf9768ad98ded44ab8651a3R29

https://github.com/OCA/rma/pull/603/changes# diff-7b933f9bf4286d07a401f46946531786b8cfdc49bbf9768ad98ded44ab8651a3R30 (this represents the RMA's `warehouse_id`, whereas in the wizard it represents the value defined in the wizard).

pedrobaeza: OK, understood with the different warehouse explanation.

rma_lot/models/rma.py
```diff
@@ -20,17 +20,27 @@ class Rma(models.Model):
...
 
+    def _get_lot_id_quant_domain_locations(self):
+        """This method retrieves the location(s) that will later be used in
+        _domain_lot_id_quant_domain(). It will be useful to extend this method if,
+        for example, you want to search across different multi-warehouse
+        locations.
+        """
+        self.ensure_one()
+        warehouse = self.warehouse_id
+        return warehouse.lot_stock_id + warehouse.rma_loc_id
```
carlos-lopez-tecnativa, line 31: What if you pass the warehouse as a parameter? This way, you can reuse the same method, can't you?

```suggestion
    def _get_lot_id_quant_domain_locations(self, warehouse):
        """This method retrieves the location(s) that will later be used in
        _domain_lot_id_quant_domain(). It will be useful to extend this method if,
        for example, you want to search across different multi-warehouse
        locations.
        """
        return warehouse.lot_stock_id + warehouse.rma_loc_id
```

rma_lot/models/rma.py
```diff
@@ -20,17 +20,27 @@ class Rma(models.Model):
...
+
     def _domain_lot_id_quant_domain(self):
         """This method defines the domain that will be used to obtain the appropriate
         stock.quant values and is useful for extending to other modules.
         """
         self.ensure_one()
+        locations = self._get_lot_id_quant_domain_locations()
```
carlos-lopez-tecnativa, line 38:
```suggestion
        locations = self._get_lot_id_quant_domain_locations(self.warehouse_id)
```

rma_lot/wizards/rma_delivery.py
```diff
@@ -16,17 +16,27 @@ class RmaReDeliveryWizard(models.TransientModel):
...
+        """This method retrieves the location(s) that will later be used in
+        _domain_lot_id_quant_domain(). It will be useful to extend this method if,
+        for example, you want to search across different multi-warehouse
+        locations.
+        """
+        self.ensure_one()
+        warehouse = self.warehouse_id
+        return warehouse.lot_stock_id + warehouse.rma_loc_id
```
carlos-lopez-tecnativa, line 27:
```suggestion
```

rma_lot/wizards/rma_delivery.py
```diff
@@ -16,17 +16,27 @@ class RmaReDeliveryWizard(models.TransientModel):
...
+
     def _domain_lot_id_quant_domain(self):
         """This method defines the domain that will be used to obtain the appropriate
         stock.quant values and is useful for extending to other modules.
         """
         self.ensure_one()
+        locations = self._get_lot_id_quant_domain_locations()
```
carlos-lopez-tecnativa, line 34:
```suggestion
        locations = self.env["rma"]._get_lot_id_quant_domain_locations(self.warehouse_id)
```

carlos-lopez-tecnativa (changes requested):
I think this approach could be simpler. What do you think?

Or alternatively, move this method to `stock.warehouse`.

victoralmau:
> I think this approach could be simpler. What do you think?
> 
> Or alternatively, move this method to `stock.warehouse`.

I understand your suggestion, and I think it's a great one, **but**... it's not a viable option if, for example, we want to override the method to use all of the company's warehouses from an RMA.

carlos-lopez-tecnativa (approved):
> > I think this approach could be simpler. What do you think?
> > Or alternatively, move this method to `stock.warehouse`.
> 
> I understand your suggestion, and I think it's a great one, **but**... it's not a viable option if, for example, we want to override the method to use all of the company's warehouses from an RMA.

I don't fully understand what would change in another module. My suggestion was based only on the code shown here, where you are working with a single warehouse.

In any case, I don't think this PR should be blocked because of this. Let's move forward.

PR #604, opened 2026-06-22
[18.0][IMP] rma*: Improved portal and list views

victoralmau:
Changes done:

- [x] `rma` + `rma_sale`: Move Requested operation (operation_id) block to rma page
- [x] `rma`: Add name to th/td elements to rma list portal page
- [x] `rma` + `rma_sale`: Add Origin (order_id or origin) column to rma page portal list
- [x] `rma_lot`: Add Serial Number/Lot column to rma page portal list
- [x] `rma_reason`: Add Reason column to rma page portal list
- [x] `rma_sale`: Display the Sale Order field as a link on the RMA portal form
- [x] `rma`: Improvements to the RMA section of /my (text + icon)

**Before**
antes
ficha-antes
my-rma-antes

**After**
despues
ficha-despues
my-rma-despues

@Tecnativa


pedrobaeza:
Can you put some screenshots?

victoralmau:
> Can you put some screenshots?

Added to the PR description.

carlos-lopez-tecnativa (approved):
LGTM, thanks for these improvements. Could you add an icon to the RMA section, like the other sections in the portal?
Image

The module icon can be displayed in the portal.
Image

rma_sale/views/sale_portal_template.xml
```diff
@@ -266,24 +266,29 @@
             </span>
         </xpath>
     </template>
+    <template id="portal_my_rmas_sale" inherit_id="rma.portal_my_rmas">
+        <th name="th_date" position="after">
+            <th name="th_order_id">Order</th>
+        </th>
+        <td name="td_date" position="after">
+            <td name="td_order_id">
+                <span t-field="rma.order_id" />
```
pedrobaeza: Is this clickable to go to the SO?

pedrobaeza (commented):
Using AI having the existing icons as guidance, the proposal is:

Image

rma_sale/views/sale_portal_template.xml
victoralmau: Currently, the SO in the RMA list is not clickable. Personally, I think it might be “confusing” if it were, because right now the only clickable item is the RMA name, which allows you to view the details; in any case, if you think it's appropriate, I'll make the SO clickable.

victoralmau:
Currently, there is no “Origin” column in the RMA list, but I think it might be useful to have one, especially when creating RMAs from another RMA. 
What do you think? Where should it be added? (I think right after the date, before the SO).

pedrobaeza:
Put the "Origin" column for both RMA or sale, and make it clickable.

pedrobaeza:
I think this can be packed a bit more than 10 commits. You can join all the column improvements together, even if from different modules, as they are related in the logical block.

eduezerouali-tecnativa (approved):

PR #607, opened 2026-07-06
[18.0][FIX] rma: References block layout & spacing

jans23:
This fixes the column width which originally had no gap which could make reading of concatenated values difficult.

This is how it looked like before:

image

This is after the fix:

image

### DIFF
--- a/rma/views/report_rma.xml
+++ b/rma/views/report_rma.xml
@@ -47,23 +47,23 @@
                     <span t-field="doc.name" />
                 </h2>
                 <div class="row mt32 mb32" id="general_information">
-                    <div t-if="doc.origin" class="col-auto mw-100 mb-2">
+                    <div t-if="doc.origin" class="col">
                         <strong>Origin:</strong>
                         <p class="m-0" t-field="doc.origin" />
                     </div>
-                    <div class="col-auto mw-100 mb-2">
+                    <div class="col">
                         <strong>Date:</strong>
                         <p class="m-0" t-field="doc.date" />
                     </div>
-                    <div t-if="doc.deadline" class="col-auto mw-100 mb-2">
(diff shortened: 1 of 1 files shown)


rma/views/report_rma.xml
```diff
@@ -98,19 +98,19 @@
                 <table class="table table-sm o_main_table table-borderless mt-4">
                     <tbody>
                         <tr t-if="doc.picking_id" name="tr_picking">
-                            <td>Origin delivery</td>
+                            <td style="width: 40%">Origin delivery</td>
```
pilarvargas-tecnativa: Is the 40% based on a specific rendering issue? It looks a bit arbitrary, so I was wondering whether a different ratio would work equally well.

Also, if possible, I'd prefer using existing Bootstrap utility classes (`w-*` or a reusable CSS class if Bootstrap doesn't provide an equivalent) instead of an inline style, as it's generally easier to maintain and keeps the markup cleaner.

jans23: @pilarvargas-tecnativa Indeed 40% is an arbitrary choice and I changed it to `w-25` now.

pilarvargas-tecnativa (approved):
Code review 👍

pedrobaeza:
Please put a proper commit message: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

jans23:
@pedrobaeza like this?

pedrobaeza:
Well, it should have separator between the module name and the explanation, and be more concise. You can put `[FIX] rma: References block layout & spacing`.

jans23:
@pedrobaeza done.

pedrobaeza (approved):
/ocabot merge patch

PR #609, opened 2026-07-21
[18.0][FIX] rma_sale: Prevent the automatic creation of an extra sales line when validating a delivery picking

### LINKED ISSUE #608: [18.0] Customer return reuses receipt procurement group causing unexpected Sale Order Line creation
## Module
rma, rma_sale

## Describe the bug
When the company setting **"Group RMA returns by shipping address and warehouse"** is disabled, the customer return generated from an RMA reuses the same `procurement.group` as the receipt.

If the RMA is linked to a Sale Order (`rma_sale`), the reused procurement group keeps the `sale_id` of the original Sale Order.

As a consequence, when validating the outgoing delivery, Odoo's standard `sale_stock` logic creates a new `sale.order.line` because the picking is still linked to the Sale Order while the generated stock move has no `sale_line_id`.

When the company setting is enabled, a new procurement group is created for the customer return and the problem does not occur.

## To Reproduce
**Affected versions**: 18.0

1. Install `rma` and `rma_sale`.
2. Disable the company setting **"Group RMA returns by shipping address and warehouse"**.
3. Create and confirm a Sale Order.
4. Deliver the products.
5. Create an RMA from the Sale Order.
6. Process the reception.
7. Execute **Return to customer**.
8. Validate the outgoing delivery.

**Expected behavior**
Disabling the grouping option is expected to prevent grouping customer returns together.

It is unexpected that the customer return reuses the procurement group created for the reception, since this preserves the original `sale_id` and results in the creation of an additional sale order line by the standard `sale_stock` logic.

**Additional context**
The behaviour has been reproduced on a clean RunBoat instance using the current 18.0 branch with only the `rma`, `rma_delivery`, `rma_sale` and `rma_sale_mrp` modules installed. This suggests that the issue is reproducible in the standard OCA implementation and is not caused by local customizations.

victoralmau:
Prevent the automatic creation of an extra sales line when validating a delivery picking

Example use case:
- Disable the company setting "Group RMA returns by shipping address and warehouse".
- Create and confirm a Sale Order.
- Deliver the products.
- Create an RMA from the Sale Order.
- Process the reception.
- Execute Return to customer.
- Validate the outgoing delivery.

Fixes #608

@Tecnativa

### DIFF
--- a/rma_sale/models/rma.py
+++ b/rma_sale/models/rma.py
@@ -260,6 +260,14 @@ class Rma(models.Model):
             vals["sale_line_id"] = move.sale_line_id.id
         return vals
 
+    def _delivery_should_be_grouped(self):
+        # It is important to always return True if there is a linked sales order
+        # so that rma creates a new procurement.group, there by preventing an extra
+        # sales order line from being automatically created once the delivery
+        # picking is done.
+        res = super()._delivery_should_be_grouped()
+        return True if self.order_id else res
+
     def create_replace(self, scheduled_date, warehouse, product, qty, uom):
         # When the procurement group has the sale id set it will propagate to the
         # pickings. This is inconvenient for this operation as when we confirm the
--- a/rma_sale/tests/test_rma_sale.py
+++ b/rma_sale/tests/test_rma_sale.py
@@ -199,6 +199,8 @@ class TestRmaSale(TestRmaSaleBase):
         wizard = self._rma_sale_wizard(order)
         rma = self.env["rma"].browse(wizard.create_and_open_rma()["res_id"])
         self.assertEqual(rma.order_id, order)
+        rma_group = rma.procurement_group_id
+        self.assertEqual(rma_group.sale_id, order)
         rma.reception_move_id.quantity = rma.product_uom_qty
         rma.reception_move_id.picking_id.button_validate()
         self.assertEqual(rma.state, "received")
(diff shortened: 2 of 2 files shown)


mariancuadranetkia (approved):
LGTM!

carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (changes requested):
Not sure the best solution is to create a new procurement group, but to inhibit the creation of such line, as having several procurement groups for the same source (a sales order) seems awkward, don't you think?

And I think the regression test should be focus just on checking if the extra line is created. Isn't possible to reuse a existing one just adding that check?

victoralmau:
> Not sure the best solution is to create a new procurement group, but to inhibit the creation of such line, as having several procurement groups for the same source (a sales order) seems awkward, don't you think?
> 
> And I think the regression test should be focus just on checking if the extra line is created. Isn't possible to reuse a existing one just adding that check?

The creation of 2 procurement groups is something that already occurs when the “Group RMA returns by shipping address and warehouse” option is enabled (it is enabled by default); that is why this solution was implemented, and the tests verify precisely that, among other things.

Another test cannot be reused because none of them test the RMA workflow with `rma_return_grouping=False`.

rma_sale/models/rma.py
```diff
@@ -260,6 +260,14 @@ def _prepare_reception_procurement_vals(self, group=None):
             vals["sale_line_id"] = move.sale_line_id.id
         return vals
 
+    def _delivery_should_be_grouped(self):
+        # It is important to always return True if there is a linked sales order
+        # so that rma creates a new procurement.group, there by preventing an extra
+        # sales order line from being automatically created once the delivery
+        # picking is done.
+        res = super()._delivery_should_be_grouped()
+        return True if self.order_id else res
```
pedrobaeza, line 269: In the sake of conciseness, but nitpicking:

```suggestion
        return bool(self.order_id) or super()._delivery_should_be_grouped()
```

pedrobaeza (approved):
OK, not sure if having that check as True by default is the best, but let's move on.

/ocabot merge patch

Issue #523, opened 2026-01-13
[16.0] RMA - Unable to confirm an RMA through the overview


SupportSNDC:
## Module
rma

## Describe the bug

Unable to confirm an RMA through the preview, the following message appears:
The operation cannot be completed: another model requires the record to be deleted. If possible, archive it instead.

Model: Stock Move (stock.move)
Constraint: stock_move_operation_id_fkey

<img width="1451" height="216" alt="Image" src="https://github.com/user-attachments/assets/3574e75b-4c7c-421e-9062-4a1475a789f2" />

However, by going through the Orders menu, I was able to confirm the RMA without any problems.

<img width="562" height="189" alt="Image" src="https://github.com/user-attachments/assets/4e5e06de-9150-439b-84d8-ea504441fa50" />

<img width="1192" height="241" alt="Image" src="https://github.com/user-attachments/assets/2e9e58b5-520a-498b-bb20-b38aac1426ae" />

## To Reproduce
**Affected versions**: V16

SupportSNDC:
Hello, please apply the label "no stale".

victoralmau:
I haven't been able to reproduce the error you mentioned on an RMA 16.0 Runbot.

<img width="1481" height="750" alt="Image" src="https://github.com/user-attachments/assets/4793e44c-4957-4102-922e-2d30f6b9c65f" />

Can you provide the specific steps to reproduce the error? (if it still occurs)

SupportSNDC:
Hi @victoralmau, thanks for looking into it. I dug into this further and found the exact root cause — it's a genuine bug, but it only manifests under specific conditions, which is probably why it didn't show up on your Runbot test.

Precondition: the mrp app must be installed

stock.move doesn't have an operation_id field by default — it's added by the core mrp module (odoo/addons/mrp/models/stock_move.py):


operation_id = fields.Many2one(
    'mrp.routing.workcenter', 'Operation To Consume', check_company=True,
    domain="[('id', 'in', allowed_operation_ids)]")
rma only depends on stock_account, not mrp, so on a plain rma-only Runbot this field doesn't exist at all and the bug can't occur. It only shows up on databases where mrp is also installed (a very common combination in practice).

Root cause: a context key collision on operation_id

Two completely unrelated models both have a field literally named operation_id: rma.operation_id (→ rma.operation) and stock.move.operation_id (→ mrp.routing.workcenter, core mrp, unrelated to RMA).

[...]

PR #610, opened 2026-07-21
[16.0][FIX] rma: Clear default_operation_id from the context upon confirm to avoid side effects

victoralmau:

Steps to reproduce:
- Install `rma` + `mrp`
- Click the Confirm button

@Tecnativa

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -10,6 +10,7 @@ from itertools import groupby
 from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, ValidationError
 from odoo.tools import html2plaintext
+from odoo.tools.misc import clean_context
 
 from odoo.addons.stock.models.stock_move import PROCUREMENT_PRIORITIES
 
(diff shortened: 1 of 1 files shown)


carlos-lopez-tecnativa (approved):
LGTM, but the same situation can occur with any field whose name starts with `default_`. Therefore, you should clean all keys that start with `default_`. You can use the [clean_context](https://github.com/odoo/odoo/blob/b11bef8c724cc6c0c9d063d45be553caa4f92d24/odoo/tools/misc.py#L1000) method.

rma/models/rma.py
```diff
@@ -870,6 +871,11 @@ def action_create_receipt(self):
 
     def action_confirm(self):
         """Invoked when 'Confirm' button in rma form view is clicked."""
+        # It is important to "clear" the default_operation_id key from the context to
+        # prevent an attempt to set that value in the stock.move record that will be
+        # created for the operation_id field if mrp is installed.
+        # pylint: disable=W8121
```
pedrobaeza: Which one is this pylint? And you are disabling it for the whole method, not the specific line.

victoralmau: `rma/models/rma.py:877: [W8121(context-overridden), Rma.action_confirm] Context overridden using dict. Better using kwargs `with_context(**clean_context(self.env.context))` or `with_context(key=value)`

pedrobaeza: You can avoid the warning using `**self.env.context` thus

victoralmau: That's not what we want, because then any existing default_** keys would remain; we want to delete them.

pedrobaeza: OK, right, my fault.

pedrobaeza (approved):
/ocabot merge patch

PR #611, opened 2026-07-22
[17.0][FIX] rma: Clear context upon confirm to avoid side effects

victoralmau:
FWP from 16.0: https://github.com/OCA/rma/pull/610

Steps to reproduce:
- Install `rma` + `mrp`
- Click the Confirm button

@Tecnativa


pedrobaeza (approved):
/ocabot merge patch

PR #612, opened 2026-07-22
[18.0][FIX] rma: Clear context upon confirm to avoid side effects

### LINKED ISSUE #523: [16.0] RMA - Unable to confirm an RMA through the overview

victoralmau:
FWP from 17.0: https://github.com/OCA/rma/pull/611

Steps to reproduce:
- Install `rma` + `mrp`
- Click the Confirm button

@Tecnativa

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -13,6 +13,7 @@ from markupsafe import Markup
 from odoo import _, api, fields, models
 from odoo.exceptions import AccessError, ValidationError
 from odoo.tools import html2plaintext
+from odoo.tools.misc import clean_context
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
/ocabot merge patch

PR #613, opened 2026-07-22
[18.0][IMP] rma: allow to confirm picking without rma access rights

twalter-c2c:
When RMA is installed, stock users without read access on rma model can hit an access error while confirming pickings, even if they never use RMA directly.

### DIFF
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -109,6 +109,12 @@ class StockMove(models.Model):
             "rma_receiver_ids",
         ]
 
+    def _merge_moves(self, merge_into=False):
+        self = self.sudo()
+        if merge_into:
+            merge_into = merge_into.sudo()
+        return super()._merge_moves(merge_into=merge_into)
+
     def _prepare_move_split_vals(self, qty):
         """Intended to the backport of picking linked to RMAs propagates the
         RMA link id.
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -22,6 +22,11 @@ class TestRma(BaseCommon):
             login="user_rma",
             groups="rma.rma_group_user_own,stock.group_stock_user",
         )
+        cls.user_stock = new_test_user(
+            cls.env,
+            login="user_stock",
+            groups="stock.group_stock_user",
+        )
         cls.res_partner = cls.env["res.partner"]
         cls.product_product = cls.env["product.product"]
         cls.company = cls.env.user.company_id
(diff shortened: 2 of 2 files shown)


rma/models/stock_move.py
```diff
@@ -106,7 +118,7 @@ def _prepare_merge_moves_distinct_fields(self):
         """
         return super()._prepare_merge_moves_distinct_fields() + [
             "rma_id",
-            "rma_receiver_ids",
+            "rma_receiver_ids_key",
```
grindtildeath: Can't you just use `sudo` on `_merge_moves` instead of defining the new field?

twalter-c2c: Thank you for the suggestion. It works equally well and is much cleaner.

rma/models/stock_move.py
```diff
@@ -29,6 +29,12 @@ class StockMove(models.Model):
         comodel_name="rma", string="RMA return", copy=False, index=True
     )
 
+    @api.depends("rma_receiver_ids")
+    def _compute_rma_receiver_ids_key(self):
+        for move in self:
+            receiver_ids = sorted(move.sudo().rma_receiver_ids.ids)
+            move.rma_receiver_ids_key = ",".join(map(str, receiver_ids)) or False
```
grindtildeath: Still need to drop that from your previous version

grindtildeath (approved):
Pre approving, LG

victoralmau:
Another approach to fix the same problem: https://github.com/OCA/rma/pull/614

rma/models/stock_move.py
twalter-c2c: Sorry, I missed that part.

twalter-c2c:
> Another approach to fix the same problem: #614

It is difficult to say which approach should be used. I would appreciate @pedrobaeza's opinion on that.

pedrobaeza:
Yes, I'm asking for just the sudo in #614, so I think this one seems correct. Not seen both tests to see which one is more focused and concise for the exact problem though.

victoralmau:
Personally, I don't like this solution, because even though it works (that method is always called with `sudo()`), it simply hides the real problem (if the user call the action doesn't have access to one of the fields, an error occurs), which happens in this case with RMA but could happen with others as well (I'll try to check if there are any similar cases, for example, with mrp).

pedrobaeza:
@victoralmau I think this solution is acceptable. There's no hiding at all, as we just want to confirm the picking without errors. Is there any drawback that I'm not seeing?

victoralmau:
> @victoralmau I think this solution is acceptable. There's no hiding at all, as we just want to confirm the picking without errors. Is there any drawback that I'm not seeing?

The solution is correct; we'll simply be applying `sudo()` to all those moves, and while that might have some side effects, I don't think they'll be significant.

pedrobaeza (approved):
Let's move with this. Thanks to both for keeping a constructive discussion about it.

/ocabot merge patch

PR #614, opened 2026-07-24
[18.0][IMP] rma: Allow users without RMA permissions to confirm stock moves not related to RMA

victoralmau:
Allow users without RMA permissions to confirm stock moves not related to RMA

@Tecnativa

### DIFF
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -103,11 +103,16 @@ class StockMove(models.Model):
     def _prepare_merge_moves_distinct_fields(self):
         """The main use is that launched delivery RMAs doesn't merge
         two moves if they are linked to a different RMAs.
+        Only add the fields if they are set, to prevent an error caused
+        by a stock user without RMA permissions when confirming a picking.
         """
-        return super()._prepare_merge_moves_distinct_fields() + [
-            "rma_id",
-            "rma_receiver_ids",
-        ]
+        extra_fields = []
+        _self = self.sudo()
+        if any(move.rma_id for move in _self):
+            extra_fields.append("rma_id")
+        if any(move.rma_receiver_ids for move in _self):
+            extra_fields.append("rma_receiver_ids")
+        return super()._prepare_merge_moves_distinct_fields() + extra_fields
 
     def _prepare_move_split_vals(self, qty):
         """Intended to the backport of picking linked to RMAs propagates the
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -1092,6 +1092,25 @@ class TestRmaCase(TestRma):
         self.assertEqual(len((rma1 | rma2).reception_move_id.picking_id), 1)
         self.assertEqual(len((rma1 | rma2 | rma3).reception_move_id.picking_id), 2)
 
+    def test_stock_user_can_confirm_picking_without_rma_acl(self):
+        picking_form = Form(
+            self.env["stock.picking"].with_context(
+                default_picking_type_id=self.warehouse_company.out_type_id.id
+            )
+        )
+        picking_form.partner_id = self.partner
+        with picking_form.move_ids_without_package.new() as move:
+            move.product_id = self.product
+            move.product_uom_qty = 1
+        picking = picking_form.save()
+        user_stock = new_test_user(
+            self.env,
+            login="test-user_stock",
+            groups="stock.group_stock_user",
+        )
+        picking.with_user(user_stock).action_confirm()
+        self.assertIn(picking.state, ("confirmed", "assigned", "waiting"))
+
     def test_copy_operation(self):
         operation = self.env.ref("rma.rma_operation_refund")
         new_operation = operation.copy()


rma/models/stock_move.py
```diff
@@ -103,11 +103,16 @@ def _action_done(self, cancel_backorder=False):
...
+        Only add the fields if they are set, to prevent an error caused
+        by a stock user without RMA permissions when confirming a picking.
         """
-        return super()._prepare_merge_moves_distinct_fields() + [
-            "rma_id",
-            "rma_receiver_ids",
-        ]
+        extra_fields = []
+        _self = self.sudo()
+        if _self.rma_id:
```
pedrobaeza: Is `self` multi-record? And if one to merge has RMA and the other not?

victoralmau: Changed to ensure proper compatibility with multi-record.

In my opinion, it's not possible for some items to be linked to an RMA while others aren't; the picking (confirmed) is created from the RMA (inbound or outbound), and the user has RMA permissions.

rma/models/stock_move.py
```diff
@@ -103,11 +103,16 @@ def _action_done(self, cancel_backorder=False):
...
+        Only add the fields if they are set, to prevent an error caused
+        by a stock user without RMA permissions when confirming a picking.
         """
-        return super()._prepare_merge_moves_distinct_fields() + [
-            "rma_id",
-            "rma_receiver_ids",
-        ]
+        extra_fields = []
+        _self = self.sudo()
+        if any(move.rma_id for move in _self):
```
pedrobaeza, line 111: But why you need to do this? Simply add them always.

victoralmau: Use case example:

- A user with inventory permissions but WITHOUT RMA permissions creates a picking (for outbound, for example) and attempts to confirm it.

The following error occurs:
```

File "/opt/odoo/auto/addons/stock/models/stock_picking.py", line 1177, in action_confirm
    self.move_ids.filtered(lambda move: move.state == 'draft')._action_confirm()
  File "/opt/odoo/auto/addons/stock/models/stock_move.py", line 1577, in _action_confirm
    moves = self._merge_moves(merge_into=merge_into)
  File "/opt/odoo/auto/addons/stock/models/stock_move.py", line 1219, in _merge_moves
    for __, g in groupby(candidate_moves, key=self._merge_move_itemgetter(distinct_fields)):
  File "/opt/odoo/custom/src/odoo/odoo/tools/misc.py", line 1216, in groupby
    groups[key(elem)].append(elem)
  File "/opt/odoo/auto/addons/stock/models/stock_move.py", line 1183, in <lambda>
    return lambda move: base_getter(move) + tuple(_get_formatted_float_fields(move, f_name, float_precision) for f_name in float_fields)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 7083, in __getitem__
    return self._fields[key].__get__(self)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 4673, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 3111, in __get__
    return super().__get__(records, owner)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 1272, in __get__
    recs._fetch_field(self)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4120, in _fetch_field
    self.fetch(fnames)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4158, in fetch
    fetched = self._fetch_query(query, fields_to_fetch)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4269, in _fetch_query
    field.read(fetched)
  File "/opt/odoo/custom/src/odoo/odoo/fields.py", line 4688, in read
    lines = comodel.search_fetch(domain, field_names)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 1778, in search_fetch
    query = self._search(domain, offset=offset, limit=limit, order=order or self._order)
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 5812, in _search
    self.browse().check_access('read')
  File "/opt/odoo/custom/src/odoo/odoo/models.py", line 4444, in check_access
    raise result[1]()
odoo.exceptions.AccessError: You are not allowed to access 'RMA' (rma) records.
```

In my opinion, the best way to fix this issue is to ensure that if the record (`stock.move`) does not have any RMA data associated with it, those fields are not added to the `_prepare_merge_moves_distinct_fields()` method. This will prevent users without RMA permissions from performing processes that are NOT RMA-related.

The test that has been added specifically tests this use case.

rma/models/stock_move.py
```diff
@@ -103,11 +103,16 @@ def _action_done(self, cancel_backorder=False):
...
         two moves if they are linked to a different RMAs.
+        Only add the fields if they are set, to prevent an error caused
+        by a stock user without RMA permissions when confirming a picking.
         """
-        return super()._prepare_merge_moves_distinct_fields() + [
-            "rma_id",
-            "rma_receiver_ids",
-        ]
+        extra_fields = []
+        _self = self.sudo()
```
pedrobaeza, line 110: You can do: `self = self.sudo()`. It's a local variable.

rma/models/stock_move.py
pedrobaeza: But I thought that the previous `sudo` already handles that. And now

pedrobaeza, line 111: Thinking twice, `if self.rma_id:` serves as well, as it's just non falsy if one or several records exist.

pedrobaeza: But coming again to the main concept, I think this is done for seeing if a move can be merged with other existing ones, so isn't possible that one has `rma_id` and the other not? You may do a return without RMA.

victoralmau: Even so, only if one of them (`stock.move`) has any RMA data set will it be important to distinguish it (by adding the corresponding fields to the `_prepare_merge_moves_distinct_fields()` method) so that they aren't merged; otherwise, it won't matter if those fields aren't in the method, right?

pedrobaeza: Not sure how the comparison is done. I'm just pointing possible failure reasons. And also because I don't think this will prevent access error. Have you tried without this `if` if it fails?

grindtildeath (changes requested):
Having a look here since I approved #613 .

IMO this is not right because `_prepare_merge_moves_distinct_fields` is an `api.model` function.

victoralmau:
Conclusions:
- Calling the `_prepare_merge_moves_distinct_fields()` method with `super()` won't do anything; it just returns the list of fields.
- The field that's actually causing the problem is `rma_receiver_ids`
- The `mrp` module does something similar to what was proposed here: https://github.com/odoo/odoo/blob/7971be18e4a33ad712fd11d5e7cf597f8c04459f/addons/mrp/models/stock_move.py#L761
- The actual problem is in https://github.com/odoo/odoo/blob/7971be18e4a33ad712fd11d5e7cf597f8c04459f/addons/stock/models/stock_move.py#L1183, specifically in `base_getter(move)`, and the issue is limited to the `rma_receiver_ids` field (I assume because it’s a One2many field). 

The way to fix this in `stock` would be to change `base_getter(move)` in https://github.com/odoo/odoo/blob/7971be18e4a33ad712fd11d5e7cf597f8c04459f/addons/stock/models/stock_move.py#L1183 to `base_getter(move.sudo())` in https://github.com/odoo/odoo/blob/7971be18e4a33ad712fd11d5e7cf597f8c04459f/addons/stock/models/stock_move.py#L1183, but I’m not sure if odoo will accept that change.

Another solution would be to override the `_merge_move_itemgetter()` method and call it with `super`, but the underlying problem would still exist.

grindtildeath:
Hello @victoralmau 

> The way to fix this in stock would be to change base_getter(move) in https://github.com/odoo/odoo/blob/7971be18e4a33ad712fd11d5e7cf597f8c04459f/addons/stock/models/stock_move.py#L1183 to base_getter(move.sudo()) in https://github.com/odoo/odoo/blob/7971be18e4a33ad712fd11d5e7cf597f8c04459f/addons/stock/models/stock_move.py#L1183, but I’m not sure if odoo will accept that change.

Did you propose this change to Odoo?

Although I agree it would be cleaner, IMO it's not significantly different that just applying `sudo` on `_merge_moves` as it's done in #613 to have this module working.

I'd suggest to mention the Odoo PR as a comment in the fix of #613 so that we can merge it instead of keeping 2 open PRs on this repo trying to fix the same issue...

What do you think?

ping @pedrobaeza

pedrobaeza:
Let's go with #613 for avoiding an incorrect usage of an `@api.model` (although it's called with `self` having the recordset, but it can't be assured).

PR #616, opened 2026-08-12
[19.0][MIG] rma: Migration from 18.0 to 19.0

eugenios73:
Additional changes included during migration
============================================

The migration preserves the existing functional behavior of the module. However,
during the adaptation, some pre-existing issues were identified and fixed, and
test coverage was added for behaviors that could suffer regressions in Odoo 19.

Fix in multiple receptions preparation
--------------------------------------

The method that prepares reception procurements iterates over each RMA in the
recordset, but some values were obtained from ``self`` instead of using the
``rma`` record being processed in each iteration.

This behavior did not usually affect single RMA confirmations, but it could trigger
singleton errors or use data from an incorrect RMA when confirming multiple
returns with different products or operations together.

The preparation now consistently uses the product, return product configuration,
and alternative product from the RMA being processed in each iteration. This fixes
the existing confirmation and batch grouping flow without introducing new
functionality.

Explicit account assignment on refund lines
-------------------------------------------

Credit note lines created directly from an RMA now explicitly receive the income
account configured for the product in the RMA's company.

The module creates these lines programmatically without going through the invoice
form view. Resolving the account during value preparation avoids relying on
implicit environment defaults and ensures that the generated line has a valid,
company-consistent accounting configuration.

The test configuration also explicitly creates a sales journal and an income
account. This avoids relying on demo data, which is no longer installed by
default in Odoo 19. The refund flow test additionally verifies that the expected
account is assigned to the resulting line.

Fix in reception grouping test coverage
---------------------------------------

The test verifying reception grouping was accidentally defined inside another test
method. As a result, the framework did not discover or execute it.

The test has been converted into an independent method to ensure that two
compatible RMAs are grouped into the same picking.

An assertion that compared the reference of the first RMA against itself (and thus
always returned a positive result) has also been fixed. Now, it verifies that RMAs
that should be grouped share the same reference, and that an RMA belonging to
another customer receives a different reference.

These changes do not modify the grouping rules: they ensure that the existing
behavior is properly covered against regressions.

Suggested recipient coverage in chatter
----------------------------------------

A test has been added to verify that the RMA customer continues to appear among
the suggested recipients when composing a message.

This coverage protects the existing functional behavior after adapting the
customization to Odoo 19's new suggested recipients API. The goal is not to change
which contacts are proposed, but to guarantee that the customer remains available
as a recipient after migration.

Picking dates coverage in portal
--------------------------------

A dedicated HTTP test has been added for the RMA portal. The test creates an RMA
with automatic reception and delivery, initially accesses its portal page, validates
the related pickings, and accesses the page again.

This test covers portal rendering for both pending and completed pickings. Its
purpose is to catch template errors when rendering the scheduled date before
transfer and the effective date after validation.

This coverage accompanies the replacement of the old ``stock.picking.date`` field:
pending pickings show ``scheduled_date`` and completed ones show ``date_done``.
Dates are displayed only when available and without including the time component,
keeping portal information aligned with the actual transfer state.

Functional scope
----------------

Beyond the described fixes, no new business functionality is intended to be
introduced. The remaining changes respond to framework, model, and API updates
required to run the module on Odoo 19.

Coverage Note
=============

Functional tests pass successfully with both Odoo and OCB.

The failure of ``codecov/project`` does not represent an actual regression in coverage. The base branch ``19.0`` did not yet contain the ``rma`` module, and its report only measured 71 lines—mostly belonging to ``product_warranty``—with a coverage of 98.59%.

With the addition of ``rma``, Codecov now measures 2,360 lines. The resulting overall coverage is 89.74%, which appears as an 8.85% drop even though the compared sets have completely different scopes. In practice, the coverage of an almost empty branch is being compared to the full coverage of the migrated module.

The ``codecov/patch`` check is independent of this scope difference and currently shows 64%. Targeted functional tests are being added for the lines modified during the migration to improve this coverage, without lowering thresholds or excluding code from analysis.

Therefore:

* ``codecov/project`` mainly reflects the addition of a complete module that did not yet exist in the base branch.
* ``codecov/patch`` reflects the coverage of the changes in this migration and is the indicator currently being addressed with additional tests.
* Test suites with Odoo and OCB execute successfully.


rma/migrations/19.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,61 @@
...
+            (new_reference_id, old_reference_id),
+        )
+
+    util.create_fk(
+        cr,
+        "rma",
+        "stock_reference_id",
+        "stock_reference",
+        on_delete_action="SET NULL",
+    )
```
victoralmau: According to the `stock` OpenUpgrade migration script, almost everything is already done (https://github.com/OCA/OpenUpgrade/tree/19.0/openupgrade_scripts/scripts/stock/19.0.1.1).

The code you'll likely need to write will look something like this:

```
from openupgradelib import openupgrade

_renamed_fields = [
    ("stock.reference", "rma", "procurement_group_id", "stock_reference_id"),
]


@openupgrade.migrate()
def migrate(env, version):
    openupgrade.rename_fields(env, _renamed_fields)
```

Perhaps the correct name for the field is `reference_id` (Odoo uses `reference_ids` in many places in v19). Ping @pedrobaeza

rma/models/rma.py
```diff
@@ -1262,7 +1255,11 @@ def _prepare_refund_line_vals(self):
         rma.action_refund
         """
         self.ensure_one()
+        income_account = self.product_id.with_company(
+            self.company_id
+        )._get_product_accounts()["income"]
         return {
+            "account_id": income_account.id,
```
victoralmau: Why is this necessary? In my opinion, the appropriate value in the `account_id` field should be automatically set based on the product data and the defined chart of account from company.

rma/models/stock_move.py
```diff
@@ -151,5 +147,6 @@ def _get_custom_move_fields(self):
             "origin_returned_move_id",
             "move_orig_ids",
             "rma_receiver_ids",
+            "to_refund",
```
victoralmau, line 151: Why is this necessary now, when it wasn't necessary before (v18)?

rma/migrations/19.0.1.0.0/pre-migrate.py
pedrobaeza: `stock` prefix may make sense in foreign models that are not `stock...`, so I would leave it that way.

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+        delivered_qty_done: represents the quantity delivered and done.
+        For each 'done' move in delivery_move_ids the quantity done is
+        taken. This field is used to control when the RMA cam be set
+        to 'delivered' state.
+        """
+        for record in self:
+            delivered_qty = 0.0
+            for move in record.delivery_move_ids.filtered(
+                lambda r: r.state != "cancel" and not r.scrap_id
+            ):
```
Reyes4711-S73:
```suggestion
            for move in record.delivery_move_ids:
                if move.state == "cancel" or r.scrap_id:
                    continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+    def _compute_warehouse_id(self):
+        for record in self.filtered("location_id"):
+            record.warehouse_id = self.env["stock.warehouse"].search(
+                [("rma_loc_id", "parent_of", record.location_id.id)], limit=1
+            )
+
+    @api.depends("user_id")
+    def _compute_team_id(self):
+        self.team_id = False
+        for record in self.filtered("user_id"):
```
Reyes4711-S73:
```suggestion
        for record in self:
            if not record.user_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+        for r in self:
+            r.can_be_locked = r.remaining_qty > 0 and r.state in [
+                "received",
+                "waiting_return",
+                "waiting_replacement",
+            ]
+
+    @api.depends("location_id")
+    def _compute_warehouse_id(self):
+        for record in self.filtered("location_id"):
```
Reyes4711-S73:
```suggestion
        for record in self:
            if not record.location_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+                        ("company_id", "child_of", record.company_id.ids),
+                    ],
+                    limit=1,
+                )
+            )
+
+    @api.depends("partner_id")
+    def _compute_partner_shipping_id(self):
+        self.partner_shipping_id = False
+        for record in self.filtered("partner_id"):
```
Reyes4711-S73:
```suggestion
        for record in self:
            if not record.partner_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+            address = record.partner_id.address_get(["delivery"])
+            record.partner_shipping_id = address.get("delivery", False)
+
+    @api.depends("partner_id")
+    def _compute_partner_invoice_id(self):
+        self.partner_invoice_id = False
+        for record in self.filtered("partner_id"):
```
Reyes4711-S73:
```suggestion
        for record in self
            if not record.partner_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+            address = record.partner_id.address_get(["invoice"])
+            record.partner_invoice_id = address.get("invoice", False)
+
+    @api.depends("picking_id")
+    def _compute_move_id(self):
+        """Empty move on picking change, but selecting the move in it if it's single."""
+        self.move_id = False
+        for record in self.filtered("picking_id"):
```
Reyes4711-S73:
```suggestion
        for record in self:
            if not record.picking_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+            if len(record.picking_id.move_ids) == 1:
+                record.move_id = record.picking_id.move_ids.id
+
+    @api.depends("move_id")
+    def _compute_product_id(self):
+        self.product_id = False
+        for record in self.filtered("move_id"):
```
Reyes4711-S73:
```suggestion
        for record in self:
            if not record.move_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+            record.product_id = record.move_id.product_id.id
+
+    @api.depends("move_id")
+    def _compute_product_uom_qty(self):
+        self.product_uom_qty = False
+        for record in self.filtered("move_id"):
```
Reyes4711-S73:
```suggestion
        for record in self:
            if not record.move_id:
                continue
```

rma/models/rma.py
```diff
@@ -0,0 +1,1640 @@
...
+            ).message_post_with_source(
+                rma.company_id.rma_mail_draft_confirmation_template_id.get_external_id()[
+                    rma.company_id.rma_mail_draft_confirmation_template_id.id
+                ],
+                subtype_xmlid="rma.mt_rma_notification",
+            )
+
+    def _send_confirmation_email(self):
+        """Auto send notifications"""
+        for rma in self.filtered(lambda p: p.company_id.send_rma_confirmation):
```
Reyes4711-S73:
```suggestion
        for rma in self
            if not rma.company_id.send_rma_confirmation:
                continue
```
Please, do it in the rest of similar cases

rma/models/stock_move.py
eugenios73: @victoralmau this is necessary in version 19 because `stock.move.to_refund` now defaults to `True`. RMA explicitly sets this value to `False` for operations other than `update_quantity`, but provisioning values ​​are only propagated to the generated movement when the field is included in `_get_custom_move_fields()`.

Without this input, the explicit `False` value is discarded, and the new v19 default value is used, which modifies existing RMA behavior. Existing tests cover both cases: regular refund operations should result in `to_refund=False`, while `update_quantity` should result in `to_refund=True`. Please correct me if this is incorrect.

rma/migrations/19.0.1.0.0/pre-migrate.py
eugenios73: Thanks for the review, @victoralmau  & @pedrobaeza . You're completely right—that was my mistake in the pre-commit file. I've updated it, and the new version should be good now. Could you please take a look and confirm the refactor? Thanks!

victoralmau:
Sorry, but I don't understand the suggestions to remove `filtered` to achieve the same result conceptually. 
I think it's simply a matter of code design preferences and isn't related to the migration.

Reyes4711-S73:
@victoralmau 
Both options are `O(n)`, but a `for` loop with `continue` is usually slightly more efficient because it does not create an intermediate recordset or execute a `lambda`:

```python
for line in self.line_ids:
    if not line.is_valid:
        continue
    # Process line
```

Using `filtered()` can be more expressive:

```python
for line in self.line_ids.filtered("is_valid"):
    # Process line
```

My rule of thumb:

- Use `continue` if you only need the filtering for that loop or the condition is complex.
- Use `filtered()` if it clearly improves readability or you will reuse the filtered recordset.
- If you can avoid loading unnecessary lines by using a `search()` with a domain, that will have a much greater impact than choosing between these two options.

In practice, the difference is negligible for small recordsets, so prioritize readability. For large loops, `continue` has a slight performance advantage.

Also note that `filtered()` filters the recordset in Python; it does not execute a SQL query.

(thanks codex)

Reyes4711-S73 (approved):

pablo-cort-s73 (approved):
LGTM

victoralmau:
> @victoralmau Both options are `O(n)`, but a `for` loop with `continue` is usually slightly more efficient because it does not create an intermediate recordset or execute a `lambda`:
> 
> ```python
> for line in self.line_ids:
>     if not line.is_valid:
>         continue
>     # Process line
> ```
> 
> Using `filtered()` can be more expressive:
> 
> ```python
> for line in self.line_ids.filtered("is_valid"):
>     # Process line
> ```
> 
> My rule of thumb:
> 
>     * Use `continue` if you only need the filtering for that loop or the condition is complex.
> 
>     * Use `filtered()` if it clearly improves readability or you will reuse the filtered recordset.
> 
>     * If you can avoid loading unnecessary lines by using a `search()` with a domain, that will have a much greater impact than choosing between these two options.
> 
> 
> In practice, the difference is negligible for small recordsets, so prioritize readability. For large loops, `continue` has a slight performance advantage.
> 
> Also note that `filtered()` filters the recordset in Python; it does not execute a SQL query.
> 
> (thanks codex)

What do you think @pedrobaeza  ?

pedrobaeza:
@Reyes4711-S73 @victoralmau there's still a more important thumb rule: on a migration, you should reduce the diff at maximum, so it's better to keep the previous code style even if it doesn't match your preferences unless there's an important reason for changing it (for example, an important performance draining). And even though, it can go into a separate commit `[REF]` or `[PERF]` one for proper diff reading.

Due to the scope of the `filtered`, that doesn't happen, and even more important, your thumb rule with the new ORM cache system is not so important, as data is fetched in batch only once, but using `filtered` or just the loop over `self`. And accessing the fields for the condition of the filter or the if takes the same time. In previous versions, there was some kind of double processing, but this is no longer true.

Conclusion: please revert those changes and keep migration commit minimal.

And another thing: the commit message shouldn't contain the prefix `[19.0]`.

eugenios73:
@pedrobaeza @victoralmau @Reyes4711-S73 The filtered changes have been reverted and the [19.0] prefix has been removed from the commit message and increased test coverage. Please, can you review?

victoralmau:
Please, cherry-pick https://github.com/OCA/rma/pull/624 to commit history before migration commit.

eugenios73:
@victoralmau I've already added the cherry-pick

rma/models/stock_move.py
victoralmau: After reviewing the code for `sale_stock` and `stock_account` in v18 versus v19 in detail, I don't think it's necessary to add that field.

Although the `to_refund` field in `stock. move` is now True by default, it is not necessary to add it to this method, because `sale_stock` already adds it (https://github.com/odoo/odoo/blob/4973903252865a6a7a2da235bc3a01675dbbff4d/addons/sale_stock/models/stock.py#L180) -just like in v18-.

rma/models/stock_picking.py
```diff
@@ -0,0 +1,32 @@
...
+        if len(rma) == 1:
+            action.update(
+                res_id=rma.id,
+                view_mode="form",
+                view_id=False,
+                views=False,
+            )
+        else:
+            action["domain"] = [("id", "in", rma.ids)]
+        return action
```
victoralmau:
```suggestion
        return self.move_ids.rma_ids._get_records_action()
```

rma/views/rma_finalization_views.xml
```diff
@@ -0,0 +1,74 @@
...
+                        name="web_ribbon"
+                        title="Archived"
+                        bg_color="bg-danger"
+                        invisible="active"
+                    />
+                    <group>
+                        <field name="name" />
+                        <field name="company_id" groups="base.group_multi_company" />
+                        <field name="company_id" invisible="1" />
+                        <field name="active" invisible="1" />
```
victoralmau:
```suggestion
```

victoralmau (changes requested):
Functional review OK.

Image

Can you review this and correct it? (Project has something similar with the colors in the Kanban view)

eugenios73:
@victoralmau thank you very much for your review. I have corrected the approach for the width adjustment.

The previous rule was applied to `.o_rma_kanban .o_kanban_renderer`, modifying the width of the kanban card, but not the button’s dropdown menu. In Odoo 19, this menu is rendered as an overlay with the class `.o-dropdown--kanban-record-menu`, outside the `.o_rma_kanban` container.

To limit the change exclusively to the RMA dashboard:

- The `o_rma_kanban_menu` class has been added to the menu content.
- A width of `185px` has been applied to the overlay using `:has()`.
- An adaptive `max-width` has been added to prevent overflow on small screens.
- `selection_mode`, the existing actions, and the menu logic remain unchanged.
- The asset continues to be loaded through `web.assets_backend`.

I have performed functional tests locally. I am attaching a screenshot of the result.

image

victoralmau (approved):

PR #619, opened 2026-08-19
[18.0][IMP] rma_inter_company: Refund with/without compatibility

victoralmau:
Refund with/without compatibility

Use Case A (invoice exists in Company A)
- A refund (with an invoice) is processed for RMA A
- The refund for RMA B should be processed without an invoice

Use Case B (invoice exists in Company B)
- A refund (without an invoice) is processed for RMA A
- The refund for RMA B should be processed with an invoice

@Tecnativa TT64237


carlos-lopez-tecnativa (approved):
LGTM

pedrobaeza (approved):
/ocabot merge patch

PR #620, opened 2026-08-20
[16.0][IMP] rma: add pending reception status to dashboard 

sbejaoui:
image

### DIFF
--- a/rma/i18n/fr.po
+++ b/rma/i18n/fr.po
@@ -765,6 +765,11 @@ msgid "Count Rma Draft"
 msgstr "Nbre RMA en brouillon"
 
 #. module: rma
+#: model:ir.model.fields,field_description:rma.field_rma_operation__count_rma_pending_reception
+msgid "Count Rma Pending Reception"
+msgstr "Nbre RMA en attente de réception"
+
+#. module: rma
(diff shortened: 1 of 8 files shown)


pedrobaeza (commented):
Isn't `waiting_return` state the one for that? Adding these computed stored field to the existing model will trigger a massive update that should be pre-done by SQL.

sbejaoui:
the ambiguity in this case comes from the fact that the linear workflow we currently have 
(draft --> waiting reception --> received --> waiting replacement --> done) is not sufficient for all the business cases we need to support

for example, we have cases where the replacement product is sent as soon as the rma is confirmed, while the returned product is received later, this allows us to provide a quicker response to the customer

in that scenario, the waiting reception state overlaps with the replacement process, for now, this is acceptable because we consider the replacement state to be more relevant from a business perspective

ideally, the workflow should be more dynamic, the order of the steps could follow the configured operations, or we could have independent workflows for each type of action, reception, delivery, and repair, so the user can see the progress of each action at the same time

this is something I have had in the backlog for a while, and it could be a good topic for the next code sprint

in the meantime, my idea is to add a field dedicated to tracking the reception state independently from the main rma state, to be used it in the dashboard to expose rmas that are still waiting for reception

I agree that a migration script is needed here to initialize the computed field for existing records

Issue #615, opened 2026-08-11
Split not delivered RMA changes to Returned


mariancuadranetkia:
## Module
rma

## Describe the bug
When in a received but not delivered RMA, we use Split action, original RMA is changed to 'Delivered' RMA status automatically.

## To Reproduce
**Affected versions**: 17.0, 18.0

1. Install rma
2. Create and Receive RMA with quantity >=2 
3. Deliver 1 unit --> but NOT finish outgoing picking
4. Split RMA to extract other quantities
5. A new RMA is created in 'Received' with quantities extracted 
6. Original RMA status has changed to 'Delivered'

**Expected behavior**
Original status RMA should be the same before we use 'Split' action.

PR #621, opened 2026-08-27
[18.0][IMP] rma: Add delivered_qty_done + remaining_qty_to_done - #615

bielvidalnetkia:
…emaining_qty in extract_quantity #615


pedrobaeza (changes requested):
Having to modify some test result means that the patch is not correct. You should check another way to resolve the problem depicted.

bielvidalnetkia:
What would be the correct way to solve the problem, given that we don't have the 'delivered_qty_done' variable that was available in version 15? Do you think it would be a good idea to bring that variable back?

pedrobaeza:
I haven't studied it, but start adding a test that reproduce the problem, and then with that test, and not modifying the rest of the tests, you should do a fix that satisfies all.

pedrobaeza:
Please analyze through blame if this is something added later in previous version, and if so, you should cherry-pick the commits to keep attribution and explanations.

pedrobaeza:
You should also follow the proper commit message convention: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

bielvidalnetkia:
To start with, i've recovered the `delivered_qty_done` and `remaining_qty_to_done` fields; a new computed field was implemented for `delivered_qty_done` to calculate the units that have been processed. In the function `extract_quantity` function checks the second variable to verify the RMA status. Now, if the RMA outbound shipment has not been validated, the status is set to `waiting_return`.

bielvidalnetkia:
@pedrobaeza Can you review the pull request? Thanks you for your help.

pedrobaeza:
There are still pending comments, like the commit message convention (and there should be only one commit), and there's no test checking the flow that leads to your problem AFAIK (the steps in #615). Have you checked if you can cherry-pick something from the other branch?

@victoralmau FYI

bielvidalnetkia:
@pedrobaeza I do a new function to test the issue #615, change all the changes in one commit and change the commit title. I'ts correct now? Thanks you.

victoralmau:
Upon closer inspection, I see that this existed in v16 but was removed during the migration to v17 (https://github.com/OCA/rma/commit/9cc5bca1532edc9c8de4fd2e0d0a6cb4e0bb8161). I don't understand why. 

I'm going to restore it in v17 first, and then we can carry it over to v18.

victoralmau:
PR v17 https://github.com/OCA/rma/pull/623

bielvidalnetkia:
@victoralmau Thanks you for the PR, I need to include your commit in v18. How can I do it? With a cherry-pick?

victoralmau:
You can use this PR if you want; delete the existing commit and cherry-pick the commit from https://github.com/OCA/rma/pull/623

Also, change the PR title to make it similar to https://github.com/OCA/rma/pull/623 (with the prefix 18.0)

pedrobaeza:
@bielvidalnetkia delete it, not reverting it. If you want, we can do the PR for 18.

victoralmau:
https://github.com/OCA/rma/pull/624

PR #623, opened 2026-09-03
[17.0][IMP] rma: Add delivered_qty_done + remaining_qty_to_done

### LINKED ISSUE #615: Split not delivered RMA changes to Returned

victoralmau:
Add `delivered_qty_done` + `remaining_qty_to_done`

The fields were unnecessarily removed in https://github.com/OCA/rma/commit/9cc5bca1532edc9c8de4fd2e0d0a6cb4e0bb8161 but already existed in v16

The `_compute_delivered_qty_done()` method was added because the `_compute_delivered_qty()` method could not be used with compute_sudo (inconsistent across fields) and the same method could not be used for fields with store=True and store=False

Fixes #615 

@Tecnativa

### DIFF
--- a/rma/models/rma.py
+++ b/rma/models/rma.py
@@ -234,6 +234,11 @@ class Rma(models.Model):
         compute="_compute_delivered_qty",
         store=True,
     )
+    delivered_qty_done = fields.Float(
+        digits="Product Unit of Measure",
+        compute="_compute_delivered_qty_done",
+        store=True,
+    )
     can_be_returned = fields.Boolean(
         compute="_compute_can_be_returned",
     )
@@ -248,6 +253,11 @@ class Rma(models.Model):
     )
     remaining_qty = fields.Float(
         string="Remaining delivered qty",
+        digits="Product Unit of Measure",
+        compute="_compute_remaining_qty",
+    )
+    remaining_qty_to_done = fields.Float(
+        string="Remaining delivered qty to done",
         digits="Product Unit of Measure",
         compute="_compute_remaining_qty",
     )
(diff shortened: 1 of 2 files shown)


rma/models/rma.py
```diff
@@ -385,7 +390,33 @@ def _compute_delivered_qty(self):
...
+        "delivery_move_ids.scrapped",
+        "delivery_move_ids.quantity",
+        "delivery_move_ids.product_uom",
+    )
+    def _compute_delivered_qty_done(self):
+        """Compute 'delivered_qty_done' field.
+
+        delivered_qty_done: represents the quantity delivered and done.
+        For each 'done' move in delivery_move_ids the quantity done is
+        taken. This field is used to control when the RMA cam be set
```
```suggestion
        taken. This field is used to control when the RMA can be set
```

pedrobaeza (commented):
Is it possible to add a test showing the failure depicted in #615 ?

bielvidalnetkia (approved):
LGTM Functional

pedrobaeza (approved):

PR #627, opened 2026-09-08
[18.0][ADD] rma_product_variant_pricelist_base_price

rousseldenis:


rma_product_variant_pricelist_base_price/models/rma.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Rma(models.Model):
+    _inherit = "rma"
+
+    def _prepare_refund_line_vals(self):
+        """
+        If no sale order line associated to the RMA, use the variant price
+        computed on pricelist.
+        """
+        result = super()._prepare_refund_line_vals()
```
FrancoMaxime, line 14: I like to add the ensure_one() during inheritance, in the case of the super method are modified to work with multiple records, we get an error specifying that this one needs to be migrateed
```suggestion
        self.ensure_one()
        result = super()._prepare_refund_line_vals()
```

FrancoMaxime (commented):
LGTM: code review

a minor comment

rma_product_variant_pricelist_base_price/models/rma.py
```diff
@@ -0,0 +1,23 @@
...
+        if not self.env.company.sudo().price_display_variant_pricelist_id:
+            return result
+        line = self.sudo().sale_line_id
+        if not line and self.move_id.rma_id:
+            # We use the RMA from which it was created
+            line = self.move_id.rma_id.sudo().sale_line_id
```
sbejaoui, line 20: let see what we do with before

https://github.com/OCA/rma/pull/591/changes#r4070839172

sbejaoui (changes requested):

PR #630, opened 2026-09-11
[18.0][FIX] rma + rma_sale: Defining the correct partner fields in the RMA

victoralmau:
Changes done:
- [x] `rma`: Defining the correct partner fields in the RMA
- [x] `rma_sale`: Define the appropriate partner fields when creating an RMA
- [x] `rma_inter_company`: Improve test in the process of creating an RMA from another one

@Tecnativa TT64455


pedrobaeza (approved):
In this case, being a transversal problem, you can put only one commit, not one per module, but as you prefer.

carlos-lopez-tecnativa (approved):

PR #633, opened 2026-09-16
[19.0][IMP] rma: Allow users without RMA permissions to confirm stock moves not related to RMA

victoralmau:
FWP from 18.0: https://github.com/OCA/rma/pull/614

Allow users without RMA permissions to confirm stock moves not related to RMA

Use Case Example:
- Install `rma` + `stock_picking_invoice_link`
- Run the `stock_picking_invoice_link` tests

@Tecnativa


pedrobaeza (approved):

pedrobaeza (changes requested):
`_prepare_merge_moves_distinct_fields` is a `@api.model` method, so it's not assure that `self` contains the recordset.

pedrobaeza (approved):
Changed to the other approach.

Let's fast-track it as fw-port for fixing the issue.

/ocabot merge patch

